PR #28, opened 2015-07-06
Design Review: Remove static globals prototype

jeffdiclemente:
Static globals have been removed from CppMicroServices in this pull request.

NOTE: This is a pull request for the purposes of a design review.
Please refer to the email I sent regarding caveats and specific areas of importance for this code.


core/include/usFramework.h
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * Enable or disable auto-loading support.
+     *
+     * \param enable If \c true, enable auto-loading support, disable it otherwise.
+     *
+     * \remarks Calling this method will have no effect if support for
+     * auto-loading has not been configured into the CppMicroServices library of it
+     * it has been disabled by defining the US_DISABLE_AUTOLOADING envrionment variable.
+     */
+    void SetAutoLoadingEnabled(bool enable);
```
saschazelzer: Can be removed. Auto-loading will be factored out.

core/include/usFramework.h
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * Get the absolute path for persistent data. The returned path
+     * might be empty. If the path is non-empty, it is safe to assume
+     * that the path exists and is writable.
+     *
+     * \return The absolute path to the persistent storage path.
+     */
+    std::string GetStoragePath();
```
saschazelzer: We can remove this as well and rely on framework properties and the ModuleContext::GetDataFile() method.

core/include/usModuleImport.h
```diff
@@ -51,15 +51,7 @@ US_END_NAMESPACE
...
-  extern "C" void _us_import_module_initializer_ ## _module_name();        \
-  struct StaticModuleInitializer_ ## _module_name                          \
-  {                                                                        \
-    StaticModuleInitializer_ ## _module_name()                             \
-    {                                                                      \
-      _us_import_module_initializer_ ## _module_name();                    \
-    }                                                                      \
-  };                                                                       \
-  static StaticModuleInitializer_ ## _module_name _InitializeModule_ ## _module_name;
+  extern "C" void _us_import_module_initializer_ ## _module_name();
```
saschazelzer: We need to check the effect of these changes on static builds (probably broken). I also need to think about static modules in general.

core/include/usModuleInfo.h
```diff
@@ -48,6 +49,7 @@ struct US_Core_EXPORT ModuleInfo
   std::string name;
   std::string location;
   std::string autoLoadDir;
+  std::map<std::string, std::string> properties;
```
saschazelzer: I believe the properties member is not needed here and the module can access the framework launch properties via the core context. Actually, the complete class can probably be remove since it was a helper construct during static initialization and registration of a module. But now we have explicit install calls.

core/include/usModuleRegistry.h
```diff
@@ -27,9 +27,11 @@
 #include <string>
```
saschazelzer: I believe the file can now be made private (usModuleRegistry_p.h) because we should be able to access all of the information via the new Framework class, right?

core/include/usModuleSettings.h
```diff
@@ -31,6 +31,8 @@
```
saschazelzer: We should use framework properties instead (can be done at some point later in time). Don't worry about the autoload stuff.

core/src/util/usLog_p.h
```diff
@@ -73,19 +73,19 @@ struct LogMsg {
 US_END_NAMESPACE
 
 #if defined(US_ENABLE_DEBUG_OUTPUT)
-  #define US_DEBUG (US_PREPEND_NAMESPACE(ModuleSettings)::GetLogLevel() > US_PREPEND_NAMESPACE(DebugMsg) ? US_PREPEND_NAMESPACE(LogMsg)() : US_PREPEND_NAMESPACE(LogMsg)(US_PREPEND_NAMESPACE(DebugMsg), __FILE__, __LINE__, __FUNCTION__))
+  #define US_DEBUG US_PREPEND_NAMESPACE(LogMsg)(US_PREPEND_NAMESPACE(DebugMsg), __FILE__, __LINE__, __FUNCTION__)
```
saschazelzer: I understand why this change was made, but it disables the log level handling. Maybe we can use the GetModuleContex() function and framework properties instead in order to restore the log level mechanism?

jeffdiclemente: Using GetModuleContext() and framework properties sounds good. I'll explore that option.

core/include/usModuleRegistry.h
jeffdiclemente: Correct.

core/include/usModuleImport.h
jeffdiclemente: I built a static CppMicroServices build on Windows 64-bit and ran the unit tests. Aside from one failure in usModuleTest.cpp (which was resolved by removing the ifdef on line 116) and the already broken usLogTest unit tests, all other unit tests passed.

I will make sure that I test static CppMicroServices builds as I continue to make changes.

core/include/usModuleInfo.h
jeffdiclemente: I agree. The core context is an appropriate place given that the Framework launching properties need to be accessible through the BundleContext's "getProperty()" function (per OSGi spec, section 4.5.3 - Environment Properties).

It does look like the removal of ModuleInfo will simplify the code. I would like to make this change in a different feature branch since it is a significant change. Making ModuleRegistry private to the framework can then be rolled into this separate feature branch as well.

core/include/usModuleImport.h
saschazelzer: That sounds great, I will check this as well with gcc and clang on a Linux box.

core/include/usModuleInfo.h
saschazelzer: Sounds good to me.

core/include/usModuleImport.h
jeffdiclemente: I fixed the gcc build on Linux with [d337902] (https://github.com/jeffdiclemente/CppMicroServices/commit/d337902e1a427125efaad24bb59866fb6ee0af8c), however there are still unit test failures which will be resolved.

core/include/usModuleRegistry.h
jeffdiclemente: On second thought, making usModuleRegistry.h private should be done at the same time and doesn't need to wait until ModuleInfo is refactored.

core/include/usFramework.h
jeffdiclemente: In the interim, we should keep auto-install/load, until it has been factored out into a service.

jeffdiclemente:
All build configurations are now passing (via travis CI and in my own development environment).

I'd like to keep auto-install/load, until it has been moved out into its own service, since it will still be needed in the interim.

If you are alright with the general approach, I'm going to move onto fixing all the "TODO"s in the code and refining the solution for the official pull request.

saschazelzer:
Nice work, especially with the GetModuleContext function. I also agree with keeping the auto-install code for now. Please go ahead with the official pull request, looking forward to finalize this!

saschazelzer:
An improved version of this PR has been merged with #31 .

PR #30, opened 2015-08-24
Remove static globals

jeffdiclemente:
This addresses https://github.com/jeffdiclemente/CppMicroServices/issues/2, which is concerned with removal of static globals and other related code changes.

This pull request obsoletes https://github.com/saschazelzer/CppMicroServices/pull/28, which was primarily focused on the review of the high level design.

Important Notes:
- Auto-loading and ModuleSettings are in a state of transition. I am leaving it unfinished for now. Currently, bundles are auto-installed (i.e.auto-installed bundles have to be explicitly started/stopped), instead of auto-loaded. Requirements and/or guidance on how to transition this would be great.
- All new code uses OSGi nomenclature where applicable in anticipation of submitting code for https://github.com/jeffdiclemente/CppMicroServices/issues/1.


core/doc/snippets/uServices-modulecontext/main.cpp
```diff
@@ -16,9 +16,7 @@ void RetrieveModuleContext()
 //! [GetModuleContext]
 
 //! [InitializeModule]
```
karthikreddy09: You may want to remove the comments denoting the block since the code block does not exist any more. Initialization happens elsewhere ...

PR #31, opened 2015-08-25
Remove static globals

jeffdiclemente:
This addresses https://github.com/jeffdiclemente/CppMicroServices/issues/2, which is concerned with removal of static globals and other related code changes.

This pull request obsoletes https://github.com/saschazelzer/CppMicroServices/pull/28, which was primarily focused on the review of the high level design.

Important Notes:
- Auto-loading and ModuleSettings are in a state of transition. I am leaving it unfinished for now. Currently, bundles are auto-installed (i.e.auto-installed bundles have to be explicitly started/stopped), instead of auto-loaded. Requirements and/or guidance on how to transition this would be great.
- All new code uses OSGi nomenclature where applicable in anticipation of submitting code for https://github.com/jeffdiclemente/CppMicroServices/issues/1.


jeffdiclemente:
I'm not sure why the AppVeyor build fails. It has consistently done so with the following output:

> Build started
git clone -q https://github.com/saschazelzer/CppMicroServices.git C:\projects\cppmicroservices
git fetch -q origin +refs/pull/31/merge:
git checkout -qf FETCH_HEAD
Specify a project or solution file. The directory does not contain a project or solution file.

Is this an issue on the AppVeyor side or an issue on my end?

core/doc/snippets/uServices-modulecontext/main.cpp
```diff
@@ -16,9 +16,7 @@ void RetrieveModuleContext()
 //! [GetModuleContext]
 
 //! [InitializeModule]
```
WayneBailey: Remove comment

core/doc/snippets/uServices-modulecontext/main.cpp
```diff
@@ -16,9 +16,7 @@ void RetrieveModuleContext()
...
-#include <usModuleInitialization.h>
 
-US_INITIALIZE_MODULE
 //! [InitializeModule]
```
WayneBailey: Remove comment

core/doc/snippets/uServices-resources/main.cpp
```diff
@@ -38,11 +37,11 @@ void parseComponentDefinition(std::istream&)
 {
 }
 
-void extenderPattern()
+void extenderPattern(ModuleContext* moduleCtx)
 {
   //! [2]
   // Get all loaded modules
```
WayneBailey: This comment isn’t relevant anymore. Nor are any mentioning “loaded modules.”

core/doc/snippets/uServices-staticmodules/main.cpp
```diff
@@ -34,6 +41,4 @@ int main(int /*argc*/, char* /*argv*/[])
 }
 
 //! [InitializeExecutable]
```
WayneBailey: Remove comment

core/doc/snippets/uServices-staticmodules/main.cpp
```diff
@@ -34,6 +41,4 @@ int main(int /*argc*/, char* /*argv*/[])
...
-#include <usModuleInitialization.h>
-US_INITIALIZE_MODULE
 //! [InitializeExecutable]
```
WayneBailey: Remove comment

core/examples/driver/main.cpp
```diff
@@ -75,12 +75,20 @@ int main(int /*argc*/, char** /*argv*/)
 {
   char cmd[256];
 
-  std::vector<std::string> availableModules = GetExampleModules();
+  FrameworkFactory factory;
+  Framework* framework = factory.newFramework(std::map<std::string, std::string>());
```
WayneBailey: A comment regarding the comment would be helpful.

core/examples/driver/main.cpp
```diff
@@ -144,80 +141,43 @@ int main(int /*argc*/, char** /*argv*/)
...
-
-            std::vector<Module*> modules = ModuleRegistry::GetModules();
-            for (std::vector<Module*>::const_iterator moduleIter = modules.begin();
-                 moduleIter != modules.end(); ++moduleIter)
-            {
-              availableModules.erase(std::remove(availableModules.begin(), availableModules.end(), (*moduleIter)->GetName()),
-                                     availableModules.end());
-            }
-
+            /* It is possible to install a bundle based on its name, if its
```
WayneBailey: This limitation is not described in the comments for InstallBundle. Why does the bundle name need to be the same as the file that contains it (assuming this is what this comment implies)?

core/examples/driver/main.cpp
```diff
@@ -75,12 +75,20 @@ int main(int /*argc*/, char** /*argv*/)
...
-  /* module path -> lib handle */
-  std::map<std::string, SharedLibrary> libraryHandles;
+  std::vector<std::string> availableModules = GetExampleModules();
 
-  SharedLibrary sharedLib(LIB_PATH, "");
+  /* install all available bundles for this example */
+  for (std::vector<std::string>::const_iterator iter = availableModules.begin();
+      iter != availableModules.end(); ++iter)
+  {
+      frameworkContext->InstallBundle(LIB_PATH + PATH_SEPARATOR + LIB_PREFIX + (*iter) + LIB_EXT + "/" + (*iter));
```
WayneBailey: Where is this location syntax (i.e. <file-path>/<bundle-name>) defined?

core/examples/driver/main.cpp
```diff
@@ -228,39 +188,31 @@ int main(int /*argc*/, char** /*argv*/)
 
       long int id = -1;
       ss >> id;
-
+      // TODO: the "system bundle" is 0 in the OSGi spec. Change this once CppMicroServices is
+      //    inline with the spec.
       if (id == 1)
       {
         std::cout << "Info: Unloading not possible" << std::endl;
```
WayneBailey: This message can be more specific. Mention system bundle.

core/examples/driver/main.cpp
```diff
@@ -288,7 +240,7 @@ int main(int /*argc*/, char** /*argv*/)
       {
         std::cout << std::right << std::setw(2) << (*moduleIter)->GetModuleId() << std::left << " | ";
         std::cout << std::setw(20) << (*moduleIter)->GetName() << " | ";
-        std::cout << std::setw(9) << ((*moduleIter)->IsLoaded() ? "LOADED" : "UNLOADED");
+        std::cout << std::setw(9) << ((*moduleIter)->IsLoaded() ? "ACTIVE" : "RESOLVED");
```
WayneBailey: Is there a "resolved" state given there's no fully supported life cycle yet?

core/include/usFramework.h
```diff
@@ -0,0 +1,155 @@
...
+ *   must behave as follows:
+ *   - Start - Does nothing because the system bundle is already started.
+ *   - Stop - Stops all installed bundles.
+ *   - Uninstall - The Framework must throw a std::runtime_error exception indicating that the 
+ *     system bundle cannot be uninstalled.
+ *
+ * Framework instances are created using a FrameworkFactory. The methods of this class can be
+ * used to manage and control the created framework instance.
+ *
+ * @remarks This class is thread-safe.
```
WayneBailey: Always thread safe or depending on build option? If not build option, why?

core/include/usFramework.h
```diff
@@ -0,0 +1,155 @@
...
+     * @throw std::runtime_error If this Framework could not be initialized.
+     */
+    void init(void);
+
+    /**
+     * Start this Framework.
+     *
+     * <p>
+     * The following steps are taken to start this Framework:
+     * -# If this Framework is not in the STARTING state, initialize this Framework.
```
WayneBailey: So if I call this Start method, then I don't need to call the Init method above?
If so, then why the Init method?

core/src/util/usFrameworkPrivate.cpp
```diff
@@ -0,0 +1,61 @@
+/*=============================================================================
```
WayneBailey: It would be helpful to describe why FrameworkPrivate is necessary.

core/src/module/usModuleEvent.cpp
```diff
@@ -99,10 +99,12 @@ std::ostream& operator<<(std::ostream& os, ModuleEvent::Type eventType)
 {
   switch (eventType)
   {
-  case ModuleEvent::LOADED:    return os << "LOADED";
-  case ModuleEvent::UNLOADED:  return os << "UNLOADED";
-  case ModuleEvent::LOADING:   return os << "LOADING";
-  case ModuleEvent::UNLOADING: return os << "UNLOADING";
+  case ModuleEvent::LOADED:         return os << "LOADED";
```
WayneBailey: Why are events other than installed and uninstalled needed?

core/src/util/usFrameworkFactory.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+FrameworkFactory::FrameworkFactory(void) 
+{
+}
+
+FrameworkFactory::~FrameworkFactory(void) 
+{
+}
+
+Framework* FrameworkFactory::newFramework(std::map<std::string, std::string> configuration)
```
WayneBailey: Where are valid configuration options defined? 
Where ever such configurations are referenced there should be a pointer to where these options are defined.

core/src/util/usLog_p.h
```diff
@@ -69,25 +69,82 @@ struct LogMsg {
...
+public:
+    static Logger& instance()
+    { 
+      /* 
+      IMPORTANT: This is only thread safe if compiling with a C++11 compiler
+        which implements C++11 standard section 6.7.4:
+       "If control enters the declaration concurrently while the variable is being initialized, 
+        the concurrent execution shall wait for completion of the initialization."
+
+        However, even if a C++11 supported compiler isn't used, thread safe initialization
```
WayneBailey: If thread safety isn't an issue, why mention it?
Is there a compelling reason why someone other than the framework would like to use it and thus make thread safety an issue? If so what's that reason?

core/src/util/usSharedLibrary.cpp
```diff
@@ -111,6 +110,14 @@ SharedLibrary& SharedLibrary::operator =(const SharedLibrary& other)
   return *this;
 }
 
+bool SharedLibrary::IsSharedLibrary()
+{ // Testing for file extension isn't the most robust way to test
+  // for file type. Furthermore, the existence of this function
+  // is out of place; why test for a shared library if this class
+  // is supposed to represent a shared library?
```
WayneBailey: Good question. So what are you going to do about it?

core/test/modules/libA/resources/manifest.json
```diff
@@ -0,0 +1,3 @@
+{
+  "module.name" : "TestModuleA"
```
WayneBailey: No module.version needed?

core/doc/snippets/uServices-resources/main.cpp
jeffdiclemente: Yes, nomenclature changes will correct such occurrences.

FYI; "loaded" would equate to "started". "Install" and "Uninstall" are new states introduced by these changes.

core/test/modules/libA/resources/manifest.json
jeffdiclemente: Yes, module.version required for this test bundle.

core/src/module/usModuleEvent.cpp
jeffdiclemente: Clients can still start and stop bundles. LOADED/LOADING equates to STARTED/STARTING and UNLOADING/UNLOADED equates to STOPPED/STOPPING now. These are all valid bundle states given the current functionality of CppMicroServices.

Nomenclature changes will make this clearer.

core/examples/driver/main.cpp
jeffdiclemente: It depends on what's trying to be accomplished by the status information displayed by this code.

If we want to show users bundle state with a fully supported lifecycle, "resolved" and "active" would be acceptable. Otherwise, "resolved" would be "stopped" and "active" would be "started" given our current level of lifecycle support.

Ultimately, IsLoaded() would probably be replaced by something like GetState(), which is a standard OSGi defined function, when there is a fully supported lifecycle.

core/src/util/usLog_p.h
jeffdiclemente: I mention it because thread-safety will become an issue if the LogMsg interface is made public, as shown in [branch 25-module-lifecycle-enhancements](https://github.com/saschazelzer/CppMicroServices/tree/25-module-lifecycle-enhancements) in [this commit](https://github.com/saschazelzer/CppMicroServices/commit/b3f275f44641166a9618a99c6178b1bb722c5090)

core/examples/driver/main.cpp
saschazelzer: I suggest to keep it like this for now and agree with Jeff that we need to introduce a GetState() method later instead.

saschazelzer: I suggest to keep it like this for now and agree with Jeff that we need to introduce a GetState() method later instead.

saschazelzer: I suggest to keep it like this for now and agree with Jeff that we need to introduce a GetState() method later instead.

core/include/usFramework.h
saschazelzer: Such remarks are all over the place and they are true if the related build option is enabled. Otherwise they are certainly not. We could just add a general remark to the documentation of the build option.

core/src/util/usFrameworkFactory.cpp
saschazelzer: I agree, we should add `static const std::string` members to the `Framework `class defining the config options known to the framework.

core/src/util/usLog_p.h
saschazelzer: I think we can avoid potential problems by making sure we call `Logger::instance` at least once during static initialization of the CppMicroServices class. The loader will hold a "global loader mutex" which avoids any race conditions.

core/src/util/usSharedLibrary.cpp
saschazelzer: We can do a more robust check check later using the object format parser in the 25-module-lifecycle-enhancements branch (I will create a new issue)

saschazelzer:
Concerning the AppVeyor issue, I am not sure what's going on. Lets ignore this for now, we can fix the build later.

core/doc/doxygen/MicroServices_ModuleProperties.md
```diff
@@ -7,15 +7,22 @@ The following properties are always set by the C++ Micro Services library and ca
 the module author:
 
  * `module.id` - The unique id of the module (type `long`)
- * `module.name` - The human readable name of the module (type `std::string`)
```
saschazelzer: I believe I understand why you made this change. Still, I would like to see the module.name property added by the framework, such that for basic bundles no custom manifest.json file is required.

We should be able to use the `US_MODULE_INITIALIZATION` macro to inject a string field into the custom section we were talking about in the General Discussion document. The existing object format parsers in branch 25-module-lifecycle-enhancements can then extract that information. Because this requires some more coding, I suggest to keep your changes for the sake of merging them rather sooner than later. I will create a separate issue to investigate this after the merge.

core/examples/driver/main.cpp
saschazelzer: There should be an overload in the FrameworkFactory class of the `newFramework` method taking no arguments. See my comments for this class.

core/examples/driver/main.cpp
```diff
@@ -75,12 +75,20 @@ int main(int /*argc*/, char** /*argv*/)
 {
   char cmd[256];
 
-  std::vector<std::string> availableModules = GetExampleModules();
+  FrameworkFactory factory;
+  Framework* framework = factory.newFramework(std::map<std::string, std::string>());
+  framework->init();
+  framework->Start();
```
saschazelzer: As Wayne asked somewhere else, is there a reason for separate `init` and `Start` calls? Can we get rid of the `init`?

core/examples/driver/main.cpp
```diff
@@ -75,12 +75,20 @@ int main(int /*argc*/, char** /*argv*/)
 {
```
saschazelzer: Trying to run the `usCoreExamplesDriver` executable on my Fedora 22 box crashes with the following output:

```
[sascha@bigeye-nb bin]$ ./usCoreExamplesDriver 
ServiceTracker<S,TTT>::Open: 1
terminate called after throwing an instance of 'std::runtime_error'
  what():  module.name is not defined in the bundle manifest.
Aborted (core dumped)
```
Does it run for you?

core/include/usFramework.h
```diff
@@ -0,0 +1,155 @@
...
+{
+public:
+    virtual ~Framework(void);
+
+    /**
+     * Initialize this Framework.
+     *
+     * @throw std::runtime_error If this Framework could not be initialized.
+     */
+    void init(void);
```
saschazelzer: The general policy for method names is camel case. Please change to `Init(void)`.

core/include/usFramework.h
```diff
@@ -0,0 +1,155 @@
...
+    // Allow the framework to be constructed with configuration properties
+    // provided by a FrameworkFactory object.
+    Framework(std::map<std::string, std::string>& configuration);
+    Framework(void);
+
+    // This class is not copy-able
+    Framework(const Framework&);
+    Framework operator=(const Framework&);
+
+    FrameworkPrivate* f;
```
saschazelzer: Naming conventions for private impl pointers is `d`. Please use `FrameworkPrivate* d`.

core/include/usFrameworkFactory.h
```diff
@@ -0,0 +1,63 @@
...
+{
+public:
+    FrameworkFactory(void);
+    virtual ~FrameworkFactory(void);
+
+    /**
+     * Create a new Framework instance.
+     * 
+     * @param configuration The framework properties to configure the new framework instance. If framework properties
+     * are not provided by the configuration argument, the created framework instance must use some reasonable
```
saschazelzer: Please avoid "must" clauses. These are spec / interface wording but we do not expect people to write their own `FrameworkFactory` implementations.

core/include/usFrameworkFactory.h
```diff
@@ -0,0 +1,63 @@
...
+     * default configuration. The created framework instance must copy any information needed from the specified 
+     * configuration argument since the configuration argument can be changed after the framework instance has been created.
```
saschazelzer: Remove statement about copying config arguments, we do that anyway.

core/include/usFrameworkFactory.h
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return A new, configured Framework instance.
+     */
+    Framework* newFramework(std::map<std::string, std::string> configuration);
```
saschazelzer: Camel case method naming `NewFramework`.

Please add an overload taking no arguments.

core/include/usFrameworkFactory.h
```diff
@@ -0,0 +1,63 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+=============================================================================*/
+
+#ifndef USFRAMEWORKFACTORY_H
+#define USFRAMEWORKFACTORY_H
+
+#include "usGlobalConfig.h"
+#include "usFramework.h"
```
saschazelzer: Remove the include statement and forward declare the `Framework` class.

core/include/usFrameworkFactory.h
saschazelzer: We probably should use unique / shared pointer for the framework and module instances. Previously, the Module pointers lived as long as the program, but with the new life cycle they should be managed. I will create a new issue for that.

core/include/usGetModuleContext.h
```diff
@@ -29,10 +29,12 @@
 #endif
 
 #include <usGlobalConfig.h>
```
saschazelzer: This is not needed anymore.

core/include/usGetModuleContext.h
```diff
@@ -29,10 +29,12 @@
...
-#include <usModuleRegistry.h>
 #include <usModule.h>
```
saschazelzer: This is not needed anymore.

core/include/usModule.h
```diff
@@ -1,368 +1,430 @@
...
+  /**
+   * Returns this module's current state.
+   *
+   * <p>
+   * A module can be in only one state at any time.
+   *
+   * @return <code>true</code> if the module is <code>LOADED</code>
+   *         <code>false</code> if it is in any other state.
+   */
+  virtual bool IsLoaded() const;
```
saschazelzer: I believe this and all the following methods do not need to be virtual. Please investigate.

core/include/usModule.h
```diff
@@ -1,368 +1,430 @@
...
+
+  ModulePrivate* d;
+
+  Module();
+
+  void Init(CoreModuleContext* coreCtx, ModuleInfo* info);
+  void Uninit();
+
+protected:
+  // purposely not implemented
```
saschazelzer: Why are the disabled copy constructor and assignment operator `protected` instead of `private`?

core/include/usModuleContext.h
```diff
@@ -1,852 +1,870 @@
...
+   *
+   * <p>
+   * This method will always return <code>0</code> when the service
+   * associated with this <code>reference</code> has been unregistered.
+   *
+   * <p>
+   * The following steps are taken to get the service object:
+   * <ol>
+   * <li>If the service has been unregistered, <code>0</code> is returned.
+   * <li>The co
```
saschazelzer: I think we will not do URL parsing and downloading in the near future. We should probably say "Typically an absolute file system path" or similar.

core/include/usModuleInfo.h
```diff
@@ -27,6 +27,7 @@
 #include <usCoreConfig.h>
 
 #include <string>
+#include <map>
```
saschazelzer: I think this is not needed.

core/include/usModuleSettings.h
```diff
@@ -43,11 +45,15 @@ US_BEGIN_NAMESPACE
  * - \e US_AUTOLOAD_PATHS A ':' (Unix) or ';' (Windows) separated list of paths
  *   from which modules should be auto-loaded.
  *
+ * \deprecated Use FrameworkFactory::newFramework(std::map<std::string, std::string> configuration) to configure the framework.
```
saschazelzer: Since this is deprecated now, we would need standardized framework properties for threading, logging, and storage configurations. These chould probably go into the `Framework` class and their values should start with `org.cppmicroservices`. Do not use `org.osgi` because of trademarks issues.

core/src/CMakeLists.txt
```diff
@@ -57,9 +60,9 @@ set(_srcs
 
 set(_private_headers
   util/usAtomicInt_p.h
+  util/usFrameworkPrivate_p.h
```
saschazelzer: Please rename to `usFrameworkPrivate.h`. The `_p` is only used if there is no `Private` suffix in the name already.

core/src/module/usModule.cpp
```diff
@@ -119,15 +121,37 @@ bool Module::IsLoaded() const
 
 void Module::Start()
 {
-
+  MutexLockingStrategy::Lock(this->d);
```
saschazelzer: We need to review the locking strategy here. Framework locks must not be held when calling client code, e.g. the module activator and module changes listeners below.

You can "opt out" and defer this by adding some info about this to the existing threading issues.

core/src/module/usModuleContext.cpp
```diff
@@ -175,8 +177,14 @@ std::string ModuleContext::GetDataFile(const std::string &filename) const
 #else
     static const char separator = '/';
 #endif
-
-  std::string baseStoragePath = ModuleSettings::GetStoragePath();
+  
+  std::string baseStoragePath;
+  std::map<std::string, std::string>::iterator prop = d->module->coreCtx->frameworkProperties.find("org.osgi.framework.storage");
```
saschazelzer: Use something like Framework::STORAGE_LOCATION (look up the OSGi standard for a naming suggestion) instead of the raw string literal.

core/src/module/usModulePrivate.h
```diff
@@ -26,13 +26,14 @@
...
-#include "usModuleRegistry.h"
+#include "usModuleRegistry_p.h"
 #include "usModuleVersion.h"
 #include "usModuleInfo.h"
 #include "usModuleManifest_p.h"
 #include "usModuleResourceContainer_p.h"
+#include "usSharedLibrary.h"
 
-#include "usAtomicInt_p.h"
+#include "usWaitCondition_p.h"
```
saschazelzer: I think this is not needed (if the inheritance below is remove).

core/src/module/usModulePrivate.h
```diff
@@ -43,7 +44,7 @@ struct ModuleActivator;
 /**
  * \ingroup MicroServices
  */
-class ModulePrivate {
+class ModulePrivate : public MultiThreaded<> {
```
saschazelzer: I this needed?

core/src/util/usFrameworkPrivate_p.h
```diff
@@ -0,0 +1,55 @@
...
+US_BEGIN_NAMESPACE
+
+class CoreModuleContext;
+class Mutex;
+
+class FrameworkPrivate
+{
+public:
+    FrameworkPrivate(void);
+    FrameworkPrivate(std::map<std::string, std::string>& configuration);
```
saschazelzer: Use `const &` as parameter type.

core/src/util/usFrameworkPrivate_p.h
```diff
@@ -0,0 +1,55 @@
...
+    virtual ~FrameworkPrivate();
+
+    /**
+     * Make the initialization of the Framework thread-safe.
+     */
+    Mutex* initLock;
```
saschazelzer: Can we use automatic storage duration instead of a pointer here?

core/src/util/usFrameworkPrivate_p.h
```diff
@@ -0,0 +1,55 @@
...
+
+    CoreModuleContext* coreModuleContext;
```
saschazelzer: Same as for the Mutex above.

core/src/util/usLog_p.h
saschazelzer: On a second thought, having a static Logger instance is in conflict with the ability to create multiple `Framework` instances. So this needs a redesign anyway, possibly relying on a `ModuleContext` instance. I suggest to keep the code as is and add yet another issue to tackle this.

core/examples/driver/main.cpp
jeffdiclemente: I saw this issue after posting the pull request.
I have fixed this issue and will submit the fix to this pull request.

core/doc/snippets/uServices-modulecontext/main.cpp
jeffdiclemente: The comment should stay since I mistakenly removed

    #include <usModuleInitialization.h>
    US_INITIALIZE_MODULE

which I believe should remain in since it illustrates the connection between declaring a bundle and getting its bundle context.

core/src/util/usSharedLibrary.cpp
jeffdiclemente: @saschazelzer, I agree. In the meantime, I am moving this function out of usSharedLibrary.cpp (because it doesn't make much logical sense) and into usUtils.cpp.

core/examples/driver/main.cpp
jeffdiclemente: The `init()` and `Start()` functions are defined in OSGi and from what I can tell, its to facilitate the Start Level Service Specification within OSGi. 

Since CppMicroServices doesn't support the Start Level Service specification, I don't see a compelling reason to keep `init()`. I will remove it.

core/include/usModuleContext.h
jeffdiclemente: I agree. Based on your and Wayne's feedback, I've revised the entire comment block to

````
  /**
   * Installs a bundle from the specified location.
   *
   * The following steps are required to install a bundle:
   * -# If a bundle containing the same location identifier is already installed, the Bundle object for that 
   *    bundle is returned.
   * -# The bundle's associated resources are allocated. The associated resources minimally consist of a
   *    unique identifier and a persistent storage area if the platform has file system support. If this step
   *    fails, a std::runtime_error is thrown.
   * -# A bundle event of type <code>BundleEvent::INSTALLED</code> is fired.
   * -# The Bundle object for the newly or previously installed bundle is returned.
   *
   * @remarks A location identifier is defined as an absolute path to a shared library or executable file
   * followed by a slash (/) and the bundle's name.
   * 
   * For example: 
   * -# <code>InstallBundle("/path/to/bundle/foo.so/foo");</code>
   * -# <code>InstallBundle("/path/to/bundle/foo.so/my_static_bundle");</code>
   *
   * @param location The location identifier of the bundle to install.
   * @return The Bundle object of the installed bundle.
   * @throws std::runtime_error If the installation failed.
   */
````

core/include/usModuleSettings.h
jeffdiclemente: That sounds good. I'll create an issue to standardize the framework properties for theading, logging and storage along with an initial set of default framework property values. If you agree, I'll make this changes as a separate PR.

For storage configuration, can we not use the OSGi org.osgi.framework.storage property? This is a standard OSGi property used to define a Framework's persistent storage location. Its already used in this PR (usModuleContext.cpp - line 182). 

Could you elaborate on the trademark issues?

core/src/util/usFrameworkPrivate_p.h
jeffdiclemente: yes, I will do that.

core/include/usModule.h
jeffdiclemente: The ````Framework```` class inherits from ````Module```` since the ````Framework```` "is a" Bundle (i.e. its the system bundle). Clients can get a ````Module```` instance from the Framework which represents the system bundle and that instance needs to behave differently under certain cases (e.g. when ````Uninstall()````, ````GetLocation()````, ````Start()```` and ````Stop()```` are called). Considering this and the OSGi's specification for the Framework class, I turned the ````Module```` class into a base class so that ````Framework```` could derive from it. That's why they are all virtual.

core/include/usModule.h
jeffdiclemente: There is no reason. This may have happened during development and I forgot to move them back under ````private````. Good catch.

core/include/usFramework.h
jeffdiclemente: I agree with documenting this in one place. I've made note of this in BuildInstructions.md

core/include/usModule.h
saschazelzer: I was probably not clear about this. I understand the inheritance relationship, but the Framework class only overrides a small subset (Start, Stop, Uninstall, ?) so my suggestion was to remove `virtual` from all methods which do not need to be override by the Framework class.

core/include/usModuleContext.h
saschazelzer: Could you explain why the bundle name at the end is needed?

core/include/usModuleSettings.h
saschazelzer: Because the project is not officially affiliated with the OSGi consortium, I would like to avoid any troubles regarding the re-use of the term "OSGi". Maybe it is not an issue for string literals - I didn't consult a lawyer. But I know one project which had to change its name because the OSGi guys knocked on its doors. Granted, it was the project name itself but still...

core/include/usModuleContext.h
jeffdiclemente: It allows static bundles to be installed explicitly into the Framework. This assumes we are adhering to OSGi's spec for the ```InstallBundle```` API.

core/src/util/usSharedLibrary.cpp
saschazelzer: Okay.

core/include/usModuleSettings.h
jeffdiclemente: Ok, understood. I will steer clear of "org.osgi".

I will change "org.osgi.framework.storage" to, minimally, replace "org.osgi" with "org.cppmicroservices".

core/include/usModule.h
jeffdiclemente: Understood. No problem, I'll convert the other functions back to non-virtual.

core/src/module/usModule.cpp
jeffdiclemente: This multi-threading issue will have to be fixed for 3.0 sooner or later. I'd prefer sooner, so I will fix this one now within this PR.

core/include/usModuleSettings.h
jeffdiclemente: @saschazelzer I've created [an issue](https://github.com/jeffdiclemente/CppMicroServices/issues/5) to track the implementation of framework properties. Please review and let me know what you think. Thanks.

core/src/module/usModule.cpp
jeffdiclemente: After looking at the problem, I don't see a good solution, that doesn't hold a lock while calling into client code, which doesn't require a significant change.

I removed the use of ````MutexLockingStrategy```` and replaced it with ````MutexLock```` and ````Mutex```` since its simpler.

I will "opt out" and add information to [the multithreading issue](https://github.com/saschazelzer/CppMicroServices/issues/32) so we can re-visit it.

core/src/module/usModuleContext.cpp
jeffdiclemente: This will be fixed as part of [this issue](https://github.com/jeffdiclemente/CppMicroServices/issues/5)

jeffdiclemente:
@saschazelzer, please let me know if there is anything else which requires immediate attention before it can be merged into the development branch. Thanks.

saschazelzer:
Strange, I didn't get any notification about your new changes but maybe I just missed it. I will have a final look and merge it asap. Thanks!

saschazelzer:
All right, thanks a lot for this massive contribution. I will add the discussed todos as new issues.

Issue #43, opened 2015-09-24
Re-organize public header files


saschazelzer:
In anticipation of new modules in the source code, the public header files should be re-organized.

 - Core headers will be located under `include/us`
 - Other module headers will be located under `include/<module>`

All header files will be stripped of the "us" prefix and the `us` namespace will not be optional / changeable any more.

Header files also must not depend on configured information.

karthikreddy09:
@jeffdiclemente 
I like this idea ... headers in specific folders avoids conflict with similar named headers in global scope.

Would it make sense to use include/CppMicroServices instead of include/us ? I think it helps with readability in client code. 

<code>#include "CppMicroServices/Bundle.hpp" </code>
vs
<code>#include "us/Bundle.hpp" </code>

saschazelzer:
You are right, it is definitely more expressive, but also a lot more to type. I would like to get opinions from more people about this.

lygstate:
I like #include "us/Bundle.hpp"

PR #44, opened 2015-09-25
1 change to osgi naming

jeffdiclemente:
This is not an official PR. It is not meant to be merged. It is meant as a review to make sure we are all on the same page with regards to moving CppMicroServices to using OSGi nomenclature.

Contained in this PR is what I have in mind to address ["Switch to using OSGi nomenclature"](https://github.com/jeffdiclemente/CppMicroServices/issues/1).

@saschazelzer, we discussed switching to OSGi nomenclature to disambiguate the overloaded terms "module", "load" and "unload" used in C++. Based on our discussions in google docs, it appears we are in agreement, but I want to make sure we are agreeing on the same thing. 
Please let me know what you think.
Thanks.


saschazelzer:
I agree with the proposed naming changes. How easy would it be to update the changes to take new commits in the development branch into account?

There are some (few) branches which I would like to merge for the 3.0 release. I was thinking to do the "maintenance" work as the last step. There is also issue #43 with a potentially larger impact.

jeffdiclemente:
Given the current state of the development branch, It should take somewhere around 2 hours to:
* make the code changes
* verify that the changes are correct (i.e. that a change from "module" to "bundle" wasn't erroneously made)
* run all unit tests for Linux, Mac and Windows.

I don't believe that making these changes after additional commits will increase the time it takes significantly.

Of course, ensuring that any new development branch changes already use "bundle, "start", "stop", etc... will save time.

PR #59, opened 2015-11-18
58 fix us declare service interface

karthikreddy09:
Fixed a compile error and added test point for the optional macro US_DECLARE_SERVICE_INTERFACE


jeffdiclemente:
looks good. as we discussed offline, a short comment in the unit test to call out that the test point is testing the explicit declaration of the service may be useful to people first looking at the code.

other than that, ship it!

PR #60, opened 2015-11-18
Use shared_ptr<Bundle>

### LINKED ISSUE #38: Use managed pointers for services, modules and the framework
With the newly introduced life-cycle of the framework itself and its associated modules, a `Module*` could outlive its association with the framework.

A `Framework*` instance could be shared between different parties.

Service instances currently need to be managed by the registering party. Using shared pointer semantics could ease memory management and keep track of used "handles" from modules within other modules and make `ungetService` calls superfluous.

jeffdiclemente:
@CppMicroServices/developers 
I cherry-picked [Sascha's shared_ptr<Bundle> changes](https://github.com/CppMicroServices/CppMicroServices/tree/32-mutex-locking-review) as well as making additional conversions to use ```shared_ptr<Bundle>```.

Convert all instances of Bundle raw pointers to ```shared_ptr<Bundle>```.

Built and tested on Linux, Mac and Windows.

Closes #38

### DIFF
--- a/core/src/bundle/usBundleEvent.cpp
+++ b/core/src/bundle/usBundleEvent.cpp
@@ -32,7 +32,7 @@ public:
 
   BundleEventData& operator=(const BundleEventData&) = delete;
 
-  BundleEventData(BundleEvent::Type type, Bundle* bundle)
+  BundleEventData(BundleEvent::Type type, std::shared_ptr<Bundle> bundle)
     : type(type), bundle(bundle)
   {
 
@@ -45,12 +45,12 @@ public:
   }
 
   const BundleEvent::Type type;
-  Bundle* const bundle;
+  const std::shared_ptr<Bundle> bundle;
 
 };
 
 BundleEvent::BundleEvent()
-  : d(0)
+  : d(nullptr)
 {
 
 }
@@ -65,7 +65,7 @@ bool BundleEvent::IsNull() const
   return !d;
 }
 
-BundleEvent::BundleEvent(Type type, Bundle* bundle)
+BundleEvent::BundleEvent(Type type, std::shared_ptr<Bundle> bundle)
   : d(new BundleEventData(type, bundle))
 {
 
@@ -83,7 +83,7 @@ BundleEvent& BundleEvent::operator=(const BundleEvent& other)
   return *this;
 }
 
-Bundle* BundleEvent::GetBundle() const
+std::shared_ptr<Bundle> BundleEvent::GetBundle() const
 {
   return d->bundle;
 }
@@ -112,7 +112,7 @@ std::ostream& operator<<(std::ostream& os, const BundleEvent& event)
 {
   if (event.IsNull()) return os << "NONE";
 
-  Bundle* m = event.GetBundle();
+  auto m = event.GetBundle();
   os << event.GetType() << " #" << m->GetBundleId() << " (" << m->GetLocation() << ")";
   return os;
 }
--- a/core/src/bundle/usBundleEvent.cpp
+++ b/core/src/bundle/usBundleEvent.cpp
@@ -32,7 +32,7 @@ public:
 
   BundleEventData& operator=(const BundleEventData&) = delete;
 
-  BundleEventData(BundleEvent::Type type, const std::shared_ptr<Bundle>& bundle)
+  BundleEventData(BundleEvent::Type type, std::shared_ptr<Bundle> bundle)
     : type(type), bundle(bundle)
   {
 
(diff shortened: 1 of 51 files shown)


core/src/util/usFramework.cpp
```diff
@@ -66,8 +66,8 @@ void Framework::Initialize(void)
   void* frameworkInit = NULL;
   std::memcpy(&frameworkInit, &initFncPtr, sizeof(void*));
   bundleInfo->location = BundleUtils::GetLibraryPath(frameworkInit);
-
-  d->coreBundleContext.bundleRegistry.RegisterSystemBundle(this, bundleInfo);
+  
+  d->coreBundleContext.bundleRegistry.RegisterSystemBundle(std::static_pointer_cast<Framework>(this->shared_from_this()), bundleInfo);
```
karthikreddy09: why do a this->shared_from_this() ? why not simply shared_from_this() ?

core/src/bundle/usBundleEvent.cpp
```diff
@@ -65,7 +65,7 @@ bool BundleEvent::IsNull() const
   return !d;
 }
 
-BundleEvent::BundleEvent(Type type, Bundle* bundle)
+BundleEvent::BundleEvent(Type type, std::shared_ptr<Bundle> bundle)
```
karthikreddy09: why is this not const reference ? All its doing is passing the param onto the constructor of BundleEventData which should make a copy of the shared_ptr if it wants shared ownership.

core/src/bundle/usBundleEvent.cpp
```diff
@@ -32,7 +32,7 @@ class BundleEventData : public SharedData
 
   BundleEventData& operator=(const BundleEventData&) = delete;
 
-  BundleEventData(BundleEvent::Type type, Bundle* bundle)
+  BundleEventData(BundleEvent::Type type, std::shared_ptr<Bundle> bundle)
```
karthikreddy09: I think this one can be a const& the assignment will create a copy anyway.

core/src/util/usFramework.cpp
jeffdiclemente: Right, removing ```this->``` should be functionally the same. I'll remove ```this->``` from here and in usBundle.cpp

core/src/bundle/usBundleEvent.cpp
jeffdiclemente: good catch. Don't want to do an extra copy. I will pass it by const reference.

core/src/bundle/usBundleEvent.cpp
jeffdiclemente: good catch. Don't want to do an extra copy. I will pass it by const reference.

saschazelzer:
Did you run the memcheck tests on your changes? Changing The `q` member in `BundlePrivate` to a shared pointer looks like a classic smart pointer cycle to me.

In general I think the changes are okay but the benefit is not worth the trouble (cycles). Having the public API returning shared pointers and only the bundle registry keeping a shared pointer internally  should be sufficient without having to think about introducing cycles (I believe there are more already, e.g. in `ServiceRegistrationBasePrivate`).

jeffdiclemente:
Thanks Sascha. I have not run the memcheck tests. Do you have instructions on how to run them?

saschazelzer:
Ah, I forgot that this is not merged yet (32-mutex-locking-review). So running valgrind manually would be required, something like `valgrind usCoreTestDriver <test-name>`, maybe with some more valgrind options. Or Visual Studio memory profiling if you are on Windows.

saschazelzer:
However, I really think we should avoid this issue by simply not holding shared pointers to `Bundle` in class members except for the bundle registry.

jeffdiclemente:
Thanks Sascha. I'll take a deeper look into the cycles. If it can't be easily solved right now, we will avoid the issue by doing what you've suggested (i.e. class members, except bundle registry, do not hold a ```shared_ptr<Bundle>```).

jeffdiclemente:
There shouldn't be anymore cycles. I decided to use a weak_ptr since its slightly safer than reverting back to using a raw pointer.
Minimally, I'd like to merge this into the beta-release branch and fix any issue as we discover them as the changes don't affect the public APIs.

I also had to merge and resolve a conflict from the development branch.

jeffdiclemente:
@saschazelzer, are you alright with merging this PR into development? 
We haven't seen any issues regarding these changes during our upgrade to the 3.0 beta release.

saschazelzer:
Thanks, looks good to me. Ship it.

Issue #33, opened 2015-08-28
usUtils.cpp - Crash can occur if FormatMessage(...) fails


jeffdiclemente:
In us::GetLastErrorStr(), the buffer 'lpMsgBuf' returned from FormatMessage(...) is used without checking the return code of FormatMessage(...). 
A crash can occur if FormatMessage(...) fails (i.e. lpMsgBuf will be uninitialized).

This error was found using VS 2013 code analysis.

jeffdiclemente:
closing issue as fixed by PR #52

PR #63, opened 2015-12-01
55 resource compiler port to cpp

karthikreddy09:
Changed the resource compiler code to 
1. Use optionparser.h for parsing command line args
2. Support verbose option
3. Remove a separate append mode. Instead, provide append as an optional arg.


karthikreddy09:
@saschazelzer @jeffdiclemente Please review my changes to the resource compiler tool and related make files in the pull request for 55-resource-compiler-port-to-cpp

core/doc/doxygen/MicroServices_Resources.md
```diff
@@ -44,6 +44,10 @@ call for that shared bundle or executable.
 Here is a full example creating a bundle and embedding resource data:
 \include uServices-resources-cmake/CMakeLists_example.txt
 
+If you are not using CMake, you can run the resource compiler from the terminal.
```
jeffdiclemente: It would be helpful for first time users to see some examples for common usage of the resource compiler (e.g. merging resource zips, appending resource to a library, etc...).

tools/rc/usResourceCompiler.cpp
```diff
@@ -92,7 +90,7 @@ static char* get_error_str()
   LPVOID lpMsgBuf;
   DWORD dw = GetLastError();
 
-  FormatMessage(
+  FormatMessageA(
```
jeffdiclemente: This is a corner case; If ```FormatMessage``` fails, this function is left with a ```lpMsgBuf``` that is undefined and the pointer to potentially garbage memory is passed to ```va_start``` inside the ```exit_printf``` function. The end result to the user is going to be an unintelligible error message.

A similar problem was addressed as part of #33. I'd suggest doing something similar to fix this.

tools/rc/usResourceCompiler.cpp
```diff
@@ -137,6 +135,17 @@ static FILE* us_fopen(const char* filename, const char* mode)
   return file;
 }
 
+std::string us_tempfile()
+{
+  char szTempFileName[MAX_PATH];
+  if (GetTempFileNameA(".", "ZIP", 1, szTempFileName) == 0)
+  {
+    char* err_str = get_error_str();
```
jeffdiclemente: ```get_error_str``` will leak memory and while I understand that all instances of ```get_error_str``` result in the process terminating, I believe that a comment block above ```get_error_str``` explaining how to use the function would be useful for others modifying this code.

jeffdiclemente:
Are there unit tests for the resource compiler? I think there should be unit tests.

tools/rc/usResourceCompiler.cpp
```diff
@@ -159,18 +168,25 @@ static char* get_error_str()
   return strerror(errno);
 }
 
-static void free_error_str(char* buf)
+static char* us_strcpy(char* dest, size_t /*dest_size*/, const char* src)
```
jeffdiclemente: Can't ```us_strcpy``` be eliminated entirely and ```us_strncpy``` used instead?

tools/rc/usResourceCompiler.cpp
```diff
@@ -377,6 +373,58 @@ static int us_zip_writer_add_from_zip_reader(mz_zip_archive *pZip, mz_zip_archiv
   return us_zip_writer_add_dir_entries(pZip, archiveName, archived_dirs);
 }
 
+struct Custom_Arg : public option::Arg
+{
+  static void printError(const char* msg1, const option::Option& opt, const char* msg2)
```
jeffdiclemente: It would be nice to have this function take ```std::string&``` instead of ```const char*```.

tools/rc/usResourceCompiler.cpp
```diff
@@ -377,6 +373,58 @@ static int us_zip_writer_add_from_zip_reader(mz_zip_archive *pZip, mz_zip_archiv
...
+    if (option.arg != 0 && strtol(option.arg, &endptr, 10)){};
+    if (endptr != option.arg && *endptr == 0)
+      return option::ARG_OK;
+    
+    if (msg) printError("Option '", option, "' requires a numeric argument\n");
+    return option::ARG_ILLEGAL;
+  }
+};
+
+#define US_PROG_NAME "usResourceCompiler"
```
jeffdiclemente: I think a ```const std::string``` should be used instead. There is no advantage I can see to using a ```#define``` over a ```const std::string``` in this case.

tools/rc/usResourceCompiler.cpp
```diff
@@ -384,144 +432,95 @@ static int us_zip_writer_add_from_zip_reader(mz_zip_archive *pZip, mz_zip_archiv
...
-  // ---------------------------------------------------------------------------------
-  //      COMMAND LINE VALIDATION
-  // ---------------------------------------------------------------------------------
+  int errCode = US_OK, return_code = EXIT_SUCCESS;
+  std::string zipFile;
+  std::string bundleName;
+  std::string archiveName;
+
+  argc -= (argc > 0);
+  argv += (argc > 0); // skip program name argv[0] if present
```
jeffdiclemente: Isn't the program name always present as the first element in argv? When is it not present?

tools/rc/usResourceCompiler.cpp
```diff
@@ -384,144 +432,95 @@ static int us_zip_writer_add_from_zip_reader(mz_zip_archive *pZip, mz_zip_archiv
...
+  option::Stats stats(usage, argc, argv);
+  std::unique_ptr<option::Option[]> options(new option::Option[stats.options_max]);
```
jeffdiclemente: ```std::make_unique``` should be used if possible. It would save a memory operation and is exception safe.

tools/rc/usResourceCompiler.cpp
```diff
@@ -384,144 +432,95 @@ static int us_zip_writer_add_from_zip_reader(mz_zip_archive *pZip, mz_zip_archiv
...
+  std::unique_ptr<option::Option[]> buffer(new option::Option[stats.buffer_max]);
```
jeffdiclemente: ditto

tools/rc/usResourceCompiler.cpp
```diff
@@ -384,144 +432,95 @@ static int us_zip_writer_add_from_zip_reader(mz_zip_archive *pZip, mz_zip_archiv
...
-      if (ferror(appendStream))
-      {
-        exit_printf(&writeArchive, "Appending zipfile failed\n");
-      }
-    } while (numRead != 0);
-
-    fclose(zipfileStream);
-    fclose(appendStream);
-    exit(EXIT_SUCCESS);
+    std::cerr << "Atleast one of the options (--apend-binary | --out-file) is required." << std::endl;
```
jeffdiclemente: typo - "Atleast"

tools/rc/usResourceCompiler.cpp
```diff
@@ -532,107 +531,74 @@ int main(int argc, char** argv)
   // ---------------------------------------------------------------------------------
 
   // Create a new zip archive which will be copied to zipfile later
-  dbg_print("Creating zip archive\n");
-  remove(zipFile);
+  std::clog << "Creating zip archive " << outfile.c_str() << std::endl;
+  remove(outfile.c_str());
```
jeffdiclemente: What happens if an older ```outfile``` couldn't be removed from the file system?

tools/rc/usResourceCompiler.cpp
```diff
@@ -532,107 +531,74 @@ int main(int argc, char** argv)
...
+    
     // add the current file to the new archive
-    if ((errCode = us_zip_writer_add_file(&writeArchive, archiveName, fileName, NULL, 0, compressionLevel,
-                                         &archivedNames, &archivedDirs)))
+    if ((errCode = us_zip_writer_add_file(&writeArchive, resArchiveName.c_str(), opt->arg, NULL, 0, compressionLevel,
+                                          &archivedNames, &archivedDirs)))
     {
-      dbg_print("Adding %s failed\n", archiveName);
-      exit_printf(&writeArchive, us_error_msg[errCode], archiveName, fileName);
+      std::clog << "Adding " << resArchiveName.c_str() << " failed" << std::endl;
```
jeffdiclemente: The ```.c_str()``` should be unnecessary here as the stream should be able to handle a ```std::string```.

tools/rc/usResourceCompiler.cpp
```diff
@@ -532,107 +531,74 @@ int main(int argc, char** argv)
...
   {
-    const char* fileName = argv[argIndex];
-    const size_t fileNameLength = strlen(fileName);
-
-    // determine the argument type
-    if (strcmp(argv[argIndex], "-a") == 0)
-    {
-      mergeFlag = 0;
-      continue;
+    if (!bundleName.size()) {
```
jeffdiclemente: Couldn't this check for the bundle name be moved farther up? It seems redundant to check the bundle name each time a resource is added to the zip file.

tools/rc/usResourceCompiler.cpp
```diff
@@ -532,107 +531,74 @@ int main(int argc, char** argv)
...
-    archiveName[bundleNameLength] = '/';
-    US_STRCPY(archiveName + bundleNameLength + 1, (sizeof archiveName) - (bundleNameLength + 1), fileName);
-
-    printf("  adding: %s\n", archiveName);
-
+    std::string resArchiveName = opt->arg;
+    resArchiveName.insert(0, "/");
+    resArchiveName.insert(0, bundleName);
+    
+    std::clog << "  adding: " << resArchiveName.c_str() << std::endl;
```
jeffdiclemente: The .c_str() should be unnecessary here as the stream should be able to handle a std::string.

tools/rc/usResourceCompiler.cpp
```diff
@@ -532,107 +531,74 @@ int main(int argc, char** argv)
...
     mz_zip_archive currZipArchive;
-    const char* currArchiveFileName = NULL;
     mz_uint currZipIndex = 0;
     mz_uint numZipIndices = 0;
-
     memset(&currZipArchive, 0, sizeof(mz_zip_archive));
-    currArchiveFileName = argv[zipArgIndices[zipArgIndex]];
-    if (!mz_zip_reader_init_file(&currZipArchive, currArchiveFileName, 0))
+    std::string currArchiveFileName = opt->arg;
+    std::clog << "Merging zip file " << currArchiveFileName.c_str() << std::endl;
```
jeffdiclemente: The ```.c_str()``` should be unnecessary here as the stream should be able to handle a ```std::string```.

tools/rc/usResourceCompiler.cpp
```diff
@@ -643,22 +609,46 @@ int main(int argc, char** argv)
...
+  }
+  
 
-  return EXIT_SUCCESS;
+  if (appendbinaryopt && !outfileopt)
+  {
+    if(std::remove(outfile.c_str()))
+      std::cerr << "Error removing temporary zip archive "  << outfile.c_str() << std:: endl;
+    else
+      std::clog << "Removed temporary zip archive " << outfile.c_str() << std:: endl;
```
jeffdiclemente: The ```.c_str()``` should be unnecessary here as the stream should be able to handle a ```std::string```.

tools/rc/usResourceCompiler.cpp
```diff
@@ -643,22 +609,46 @@ int main(int argc, char** argv)
...
+    outFileStream << zipFileStream.rdbuf();
+    std::clog << "  Final file size : " << outFileStream.tellp() << std::endl;
+  }
+  
 
-  return EXIT_SUCCESS;
+  if (appendbinaryopt && !outfileopt)
+  {
+    if(std::remove(outfile.c_str()))
+      std::cerr << "Error removing temporary zip archive "  << outfile.c_str() << std:: endl;
```
jeffdiclemente: The ```.c_str()``` should be unnecessary here as the stream should be able to handle a ```std::string```.

tools/rc/usResourceCompiler.cpp
```diff
@@ -643,22 +609,46 @@ int main(int argc, char** argv)
...
+  // ---------------------------------------------------------------------------------
+  //      APPEND ZIP to BINARY if append-binary is specified
+  // ---------------------------------------------------------------------------------
+  
+  if (appendbinaryopt)
+  {
+    std::string bundleBinaryFile(appendbinaryopt->arg);
+    std::ofstream outFileStream(bundleBinaryFile.c_str(), std::ios::ate | std::ios::binary | std::ios::app);
+    std::ifstream zipFileStream(outfile.c_str(), std::ios_base::binary);
+    std::clog << "Appending file " << bundleBinaryFile.c_str() << " with contents of resources zip file at " << outfile.c_str() << std::endl;
```
jeffdiclemente: The ```.c_str()``` should be unnecessary here as the stream should be able to handle a ```std::string```.

core/doc/doxygen/MicroServices_Resources.md
karthikreddy09: Examples are shown in the resource compiler's help output. Do we still need examples here ?

tools/rc/usResourceCompiler.cpp
karthikreddy09: a static variable cannot be inline expanded like I want to in the usage[] at line#416

core/doc/doxygen/MicroServices_Resources.md
jeffdiclemente: I think it would be helpful. What if a developer wants to understand how to use the resource compiler but doesn't necessarily have access to run the resource compiler?

tools/rc/usResourceCompiler.cpp
```diff
@@ -643,22 +609,46 @@ int main(int argc, char** argv)
...
   }
+  
+  // ---------------------------------------------------------------------------------
+  //      APPEND ZIP to BINARY if append-binary is specified
+  // ---------------------------------------------------------------------------------
+  
+  if (appendbinaryopt)
+  {
+    std::string bundleBinaryFile(appendbinaryopt->arg);
+    std::ofstream outFileStream(bundleBinaryFile.c_str(), std::ios::ate | std::ios::binary | std::ios::app);
```
jeffdiclemente: you should be able to pass the ```std::string``` instead of the raw pointer.

tools/rc/usResourceCompiler.cpp
```diff
@@ -643,22 +609,46 @@ int main(int argc, char** argv)
...
+    std::ifstream zipFileStream(outfile.c_str(), std::ios_base::binary);
```
jeffdiclemente: ditto

jeffdiclemente:
To keep the code tidy, can any ```#define US_*``` macros (e.g. ```#define US_FOPEN us_fopen```) which aren't being used anymore be removed?

tools/rc/usResourceCompiler.cpp
jeffdiclemente: Ok. I can live with the ```#define``` for now.

tools/rc/usResourceCompiler.cpp
karthikreddy09: std::make_unique not available until C++14.

jeffdiclemente:
@karthikreddy09 , as we discussed offline, we need to address the issue that CppMicroServices expects a manifest.json in a specific location of the bundle's ZIP file and that this is not obvious to bundle developers.

tools/rc/usResourceCompiler.cpp
```diff
@@ -0,0 +1,634 @@
...
+  option::Option* outfileopt = options[OUTFILE];
+  if (outfileopt && outfileopt->count() > 1 )
+  {
+    std::cerr << "(--out-file | -o) appear multiple times in the arguments. Check usage." << std::endl;
+    return_code = EXIT_FAILURE;
+  }
+  
+  if (!appendbinaryopt && !outfileopt)
+  {
+    std::cerr << "At least one of the options (--apend-binary | --out-file) is required." << std::endl;
```
jeffdiclemente: typo - "--apend-binary"

tools/rc/usResourceCompiler.cpp
```diff
@@ -0,0 +1,634 @@
...
+#define US_PROG_NAME "usResourceCompiler"
+
+enum  OptionIndex { UNKNOWN, HELP, VERBOSE, BUNDLENAME, COMPRESSIONLEVEL, OUTFILE, RESFILE, MERGEZIP, APPENDBINARY };
+const option::Descriptor usage[] =
+{
+  {UNKNOWN,      0, "" , ""    , Custom_Arg::None, "\nUSAGE: " US_PROG_NAME " [options]\n\n"
+    "Options:" },
+  {HELP,         0, "h" , "help",Custom_Arg::None, " --help, -h  \tPrint usage and exit." },
+  {VERBOSE,         0, "v" , "verbose",Custom_Arg::None, " --verbose, -v  \tRun in verbose mode." },
+  {BUNDLENAME,    0, "b", "bundle-name", Custom_Arg::NonEmpty, " --bundle-name, -b \tThe bundle name as specified in the BUNDLE_NAME compile definition."},
```
jeffdiclemente: should be "US_BUNDLE_NAME" in the description of the option.

tools/rc/usResourceCompiler.cpp
```diff
@@ -0,0 +1,634 @@
...
+enum  OptionIndex { UNKNOWN, HELP, VERBOSE, BUNDLENAME, COMPRESSIONLEVEL, OUTFILE, RESFILE, MERGEZIP, APPENDBINARY };
+const option::Descriptor usage[] =
+{
+  {UNKNOWN,      0, "" , ""    , Custom_Arg::None, "\nUSAGE: " US_PROG_NAME " [options]\n\n"
+    "Options:" },
+  {HELP,         0, "h" , "help",Custom_Arg::None, " --help, -h  \tPrint usage and exit." },
+  {VERBOSE,         0, "v" , "verbose",Custom_Arg::None, " --verbose, -v  \tRun in verbose mode." },
+  {BUNDLENAME,    0, "b", "bundle-name", Custom_Arg::NonEmpty, " --bundle-name, -b \tThe bundle name as specified in the US_BUNDLE_NAME compile definition."},
+  {COMPRESSIONLEVEL,  0, "c", "compression-level", Custom_Arg::Numeric, " --compression-level, -c  \tCompression level used for zip . Value range is 0 to 9. Default value is 6." },
+  {OUTFILE, 0, "o", "out-file", Custom_Arg::NonEmpty, " --out-file, -o \tPath to output zip file. If the file exists it will be overwritten. If this option is not provided, a temporary zip fie will be created."},
```
saschazelzer: Typo: `fie` instead of `file`.

tools/rc/usResourceCompiler.cpp
```diff
@@ -0,0 +1,634 @@
...
+
+enum  OptionIndex { UNKNOWN, HELP, VERBOSE, BUNDLENAME, COMPRESSIONLEVEL, OUTFILE, RESFILE, MERGEZIP, APPENDBINARY };
+const option::Descriptor usage[] =
+{
+  {UNKNOWN,      0, "" , ""    , Custom_Arg::None, "\nUSAGE: " US_PROG_NAME " [options]\n\n"
+    "Options:" },
+  {HELP,         0, "h" , "help",Custom_Arg::None, " --help, -h  \tPrint usage and exit." },
+  {VERBOSE,         0, "v" , "verbose",Custom_Arg::None, " --verbose, -v  \tRun in verbose mode." },
+  {BUNDLENAME,    0, "b", "bundle-name", Custom_Arg::NonEmpty, " --bundle-name, -b \tThe bundle name as specified in the US_BUNDLE_NAME compile definition."},
+  {COMPRESSIONLEVEL,  0, "c", "compression-level", Custom_Arg::Numeric, " --compression-level, -c  \tCompression level used for zip . Value range is 0 to 9. Default value is 6." },
```
saschazelzer: In the `zip .` sub-string there seems to be a non-ascii character (try to copy / paste it into a text editor.

tools/rc/usResourceCompiler.cpp
```diff
@@ -0,0 +1,634 @@
...
+  
+  // ---------------------------------------------------------------------------------
+  //      ZIP ARCHIVE WRITING (temporary archive)
+  // ---------------------------------------------------------------------------------
+  
+  // Create a new zip archive which will be copied to zipfile later
+  std::clog << "Creating zip archive " << outfile << std::endl;
+  // clear the contents of a outfile if it exists
+  std::ofstream ofile(outfile, std::ofstream::trunc);
+  ofile.close();
```
saschazelzer: I am not sure how this is supposed to work in pure "append" mode (i.e. if I want to supply a bundle file path and an existing zip file that should be appended to the bundle shared library). The current code always truncates the output file (the bundle shared library in this case?)

saschazelzer:
I like the modifications. It is a somewhat weird mix of C and C++ now but we can gradually move to more C++ later. I have to remarks:

1. The old "append" mode really just appended a single zip file to an existing file. Right now, a temporary zip file will always be created (even if it will be identical to the file supplied on the cmd line) and then appended. Because this is used a lot in the CMake macros, I think it makes sense to introduce this as a special case.

2. I believe we can do better with naming the cmd line options, and I propose the following:

--bundle-name, -n
--out-file, -o
--res-add, -r
--zip-add, -z
--bundle-file, -b

jeffdiclemente:
@saschazelzer,

Regarding point 1,
* Is my understanding of the special case correct? Is it that there needs to be a way to append an already created zip file to a binary without re-creating the zip file?
* One of our goals was to combine the two distinct operations (i.e. creation of the resource zip file and appending the zip to the bundle) into one invocation of usResourceCompiler to make the general use of it dead simple to users.
  * One of the assumptions behind this change was that *--out-file, -o* wouldn't be used in the general case. In the general case, a temporary zip file would be created, appended and discarded. 
  * It sounds like this assumption might be wrong. Can you point me to use cases where two separate operations (i.e. create the resource zip, then later append it) are required?

<p>
Regarding point 2, I like the proposed changes in cmd line options, and to make sure we're on the same page, can you confirm my understanding of the changes?

*--bundle-name, -b* __changes to__ *--bundle-name, -n*
--out-file, -o (__no change__)
*--res-file, -r* __changes to__ *--res-add, -r*
*--merge-zip-file, -m* __changes to__ *--zip-add, -z*
*--append-binary, -a* __changes to__ *--bundle-file, -b*

In addition, based on our email conversation, I would propose to add:
* *--manifest, -m*

which would take an absolute/relative path to a manifest.json file and automatically add it to the zip. *--res-file, -r* would still work for backwards compatibility and if *--manifest, -m* is present, it will override the use of *--res-file, -r* __only__ for the manifest.json file.

saschazelzer:
I agree that making it simpler to use and combining zip creation and appending is a good thing to do.

Point 1: You correctly understood the special case. Of course it also works with the current changes, but for the CMake macros this special case is actually the usual case. the final zip file can be amended from different places in the build system using the usFuncationAddResources macro. Also, the zip file will be recreated automatically if the timestamps of the added resources change. A subsequent last modified timestamp change of the zip file itself triggers a custom buld step which appends it to the bundle's library. This custom build step now would always (unnecessarily) recreate the zip file in a temporary place and append it instead of just using the original one.

Point 2: Exactly. For consistency, what about `--manifest-add` instead of just `--manifest`?

saschazelzer:
I agree that making it simpler to use and combining zip creation and appending is a good thing to do.

Point 1: You correctly understood the special case. Of course it also works with the current changes, but for the CMake macros this special case is actually the usual case. the final zip file can be amended from different places in the build system using the usFuncationAddResources macro. Also, the zip file will be recreated automatically if the timestamps of the added resources change. A subsequent last modified timestamp change of the zip file itself triggers a custom buld step which appends it to the bundle's library. This custom build step now would always (unnecessarily) recreate the zip file in a temporary place and append it instead of just using the original one.

Point 2: Exactly. For consistency, what about `--manifest-add` instead of just `--manifest`?

jeffdiclemente:
Thanks. I believe I understand the CMake use case. I should be able to remove the unnecessary re-creation of the zip file and make both the CMake and command line use cases work in a simple and efficient manner.

Sounds good; ```--manifest-add``` is fine with me.

karthikreddy09:
I like the proposed option names and changes for the special case.

tools/rc/usResourceCompiler.cpp
```diff
@@ -0,0 +1,523 @@
...
+    throw std::runtime_error("A file already exists with this name");
+  }
+  
+  if (!mz_zip_writer_add_file(writeArchive.get(), archiveEntry.c_str(), resFileName.c_str(), NULL,
+                              0, compressionLevel))
+  {
+    throw std::runtime_error("Error writing file to archive");
+  }
+  // add a directory entry for the file if needed.
+  if(archiveEntry.find_last_of("/") != std::string::npos)
```
jeffdiclemente: this won't handle nested directories. nested directories need to be supported.

tools/rc/usResourceCompiler.cpp
```diff
@@ -0,0 +1,523 @@
...
+    {
+      std::cout << "\t" << parse.nonOption(i) << std::endl;
+    }
+    return_code = EXIT_FAILURE;
+  }
+  
+  option::Option* bundleFileOpt = options[BUNDLEFILE];
+  if (bundleFileOpt && bundleFileOpt->count() > 1 )
+  {
+    std::cerr << "(--append-binary | -a) appear multiple times in the arguments. Check usage." << std::endl;
```
jeffdiclemente: should be "bundle-file" and "-b"

tools/rc/usResourceCompiler.cpp
```diff
@@ -0,0 +1,523 @@
...
+  }
+  else
+  {
+    outFile = us_tempfile();
+    deleteTempFile = true;
+  }
+  
+  try
+  {
+    if (!options[RESADD] && !options[MANIFESTADD] && options[ZIPADD].count() == 1 && options[BUNDLEFILE])
```
jeffdiclemente: A short comment calling out the requirement that we only support appending one zip to a bundle would be helpful. 
The ```options[ZIPADD].count() == 1``` appears (from someone unfamiliar with the code and requirements) to be arbitrary. A "why" comment would be helpful.

tools/rc/usResourceCompiler.cpp
```diff
@@ -0,0 +1,523 @@
...
+          {
+            throw std::runtime_error("Failed to append file " + std::string(archiveName) + "from archive " + archiveFileName);
+          }
+        }
+        else
+        {
+          AddDirectory(std::string(archiveName));
+        }
+      }
+    }
```
jeffdiclemente: The previous C implementation kept the archived names sorted (```us_archived_names_sort```). Why was that done and does that need to be carried over into this implementation?

karthikreddy09: because the array was used in a binary search function. we use set datastructure to eliminate duplicate entries.

karthikreddy09:
@saschazelzer ... Let me know if you are OK with this pull request.

saschazelzer:
Very nice work, +1 for merging.

Issue #66, opened 2015-12-21
CppMicroServices headers should be self sufficient (i.e. standalone)


jeffdiclemente:
All headers should include what it needs and not rely on includes from other headers.

One of the main benefits is that is makes refactoring easier.

mrlegowatch:
I started to investigate this, but could not find a specific cause related to the use case that opened this issue. 

That use case is, David reported that at one point during conversion to 3.0, there was a template expansion issue originating from one CppMicroServices header that was resolved by adding another CppMicroServices header, which he did not believe his client code should be including, but now he can't remember the exact issue, or which header. He did say it was in plugin code, not in a plugin's client's code. Because this issue is regarding template expansion, it can't be found using the straightforward trick of #including only one installed header file at a time, twice, in a test source file.

I ran a variant of the chkhdr script found on StackOverflow (http://stackoverflow.com/questions/1892043/self-sufficient-header-files-in-c-c?lq=1) on the installed headers to find out if there were any latent issues, even though it wouldn't directly find the issue related to the use case.

This script found four header files with issues. Two of the header files are usServiceTrackerPrivate.h and usTrackedService_p.h, both which require ServiceTracker to be declared before they can be included. In CppMicroServices, ServiceTracker.h happens to be included first before either of these includes. Because these are both private headers, I don't believe these are related to David's issue. The other two are usServletConfig.h, which is missing a #include <string>, and usShellService.h, which is missing an include guard (if included twice, it produces errors).

A visual inspection of source files deviating from the rule of always including their own header file first (which is another trick that helps find standalone header issues) found usServiceListeners.cpp including usUtils_p.h before its own header, and usServiceRegistry.cpp including system headers before its own header. These should probably be cleaned up to increase the chances of exposing issues if refactoring takes place.

I can clean up what I found so far, but until we can find a way to flush out template expansion issues resulting from missing dependent headers, these minor cleanups won't resolve the use case that opened this issue.

PR #76, opened 2016-02-18
53 LDAP query bundle metadata

kevinleeMW:
Changes to allow LDAP query on bundle properties

Requirements

R1. Use LDAP queries against bundle/framework meta-data.
R2. Allow customization of LDAP search strategies against meta-data without impacting existing client code. 
R3. Make it easy and intuitive for C++ developers to query meta-data.
R4. LDAP query performance for full string match


jeffdiclemente:
What are the requirements for this feature? 
Knowing the requirements allows code reviewers to give more meaningful feedback.

core/src/util/usProperties.cpp
```diff
@@ -0,0 +1,130 @@
...
+    
+    keys.reserve(p.size());
+    values.reserve(p.size());
+    
+    for (std::map<std::string, Any>::const_iterator iter = p.begin();
+         iter != p.end(); ++iter)
+    {
+        if (Find(iter->first) > -1)
+        {
+            std::string msg = "BundleProperties object contains case variants of the key: ";
```
jeffdiclemente: This message is inconsistent with the one on line 67.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    int actualMatchCount = 0;
+    int expectedMatchCount = 1;
+
+    for (std::shared_ptr<Bundle> bundle : bundles)
+    {
+        
+        if (bundle->GetBundleId() != 0)
+        {
+            bundle->Start();
+            //std::cout << bundle->GetProperty("bundle.name").ToString() << std::endl;
```
jeffdiclemente: I'm assuming this no longer serves a purpose? If so, I'd suggest removing this commented code.

core/test/usTestDriverActivator.cpp
```diff
@@ -57,6 +57,7 @@ US_IMPORT_BUNDLE(CppMicroServices)
 US_IMPORT_BUNDLE(TestBundleA)
 US_IMPORT_BUNDLE(TestBundleA2)
 US_IMPORT_BUNDLE(TestBundleB)
+US_IMPORT_BUNDLE(TestBundleLQ_1)
```
jeffdiclemente: Shouldn't all the other "TestBundleLQ_" bundles be listed here as statically linked bundles as well?

core/src/util/usLDAPFilter.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+bool LDAPFilter::Match(const ServiceReferenceBase& reference) const
+{
+  return d->ldapExpr.Evaluate(reference.d->GetProperties(), true);
+}
+    
+bool LDAPFilter::Match(const Bundle& bundle) const
+{
+  return d->ldapExpr.Evaluate(Properties(bundle.d->bundleManifest.GetProperties()), true);
```
jeffdiclemente: Why not call ```bundle.d->bundleManifest.GetProperties()``` instead? 
Doing this also eliminates adding LDAPFilter as a friend class to ```Bundle```.

core/src/util/usLDAPExpr.cpp
```diff
@@ -0,0 +1,852 @@
+/*=============================================================================
```
jeffdiclemente: as we discussed offline, see if you can make git recognize this as a move of a file instead of a whole new file. If that is impossible, please summarize what areas have changed for code review. Thanks.

core/src/util/usLDAPExpr_p.h
```diff
@@ -0,0 +1,180 @@
+/*=============================================================================
```
jeffdiclemente: as we discussed offline, see if you can make git recognize this as a move of a file instead of a whole new file. If that is impossible, please summarize what areas have changed for code review. Thanks.

core/src/util/usLDAPFilter.cpp
```diff
@@ -0,0 +1,131 @@
+/*=============================================================================
```
jeffdiclemente: as we discussed offline, see if you can make git recognize this as a move of a file instead of a whole new file. If that is impossible, please summarize what areas have changed for code review. Thanks.

core/src/util/usProperties_p.h
```diff
@@ -0,0 +1,57 @@
...
+  int FindCaseSensitive(const std::string& key) const;
+
+  const std::vector<std::string>& Keys() const;
+
+private:
+
+  std::vector<std::string> keys;
+  std::vector<Any> values;
+
+  static Any emptyAny;
```
jeffdiclemente: ```static Any emptyAny``` should be ```const``` as it doesn't appear to change throughout its lifetime.

core/src/util/usProperties_p.h
```diff
@@ -0,0 +1,57 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+=============================================================================*/
+
+#ifndef USPROPERTIES_P_H
+#define USPROPERTIES_P_H
+
+#include "usServiceProperties.h"
```
jeffdiclemente: headers should be self sufficient (i.e. they include everything they need for their declaration). See #66 for reasons why.

You should add ```<vector>```, ```<map>```, ```<string>``` and ```Any``` to satisfy the declaration.

core/src/util/usProperties.cpp
```diff
@@ -0,0 +1,130 @@
...
+  {
+    if (Find(iter->first) > -1)
+    {
+      std::string msg = "ServiceProperties object contains case variants of the key: ";
+      msg += iter->first;
+      throw std::runtime_error(msg.c_str());
+    }
+    keys.push_back(iter->first);
+    values.push_back(iter->second);
+  }
```
jeffdiclemente: There looks to be common code shared between the two constructors. I'd suggest refactoring to hoist the common code into one place. This makes maintenance a little easier.

core/src/util/usProperties.cpp
```diff
@@ -0,0 +1,130 @@
...
+const Any& Properties::Value(const std::string& key) const
+{
+  int i = Find(key);
+  if (i < 0) return emptyAny;
+  return values[i];
+}
+
+const Any& Properties::Value(int index) const
+{
+  if (index < 0 || static_cast<std::size_t>(index) >= values.size())
```
jeffdiclemente: even for one line if/else statements, always using braces. This cuts down on inadvertently introducing failures.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include <usBundle.h>
+#include <usLDAPFilter.h>
+#include <usLog.h>
+
+#include "usTestUtils.h"
+#include "usTestingMacros.h"
+
+using namespace us;
+
+void TestLDAPQueryBundleMetaData(BundleContext* frameworkCtx)
```
jeffdiclemente: If usTestBundleLQ_2 through usTestBundleLQ_5 do not export services and only contain meta-data, you shouldn't need to create dummy .cpp files for them. You should be able to generate "meta-data only" bundles. This will cut down on the number of files to maintain.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    US_TEST_CONDITION(actualMatchCount == expectedMatchCount, test_description);
+}
+
+int usLDAPQueryTest(int /*argc*/, char* /*argv*/[])
+{
+    US_TEST_BEGIN("LDAPQueryTest");
+    
+    FrameworkFactory factory;
+    
+    std::shared_ptr<Framework> framework = factory.NewFramework(std::map<std::string, std::string>());
```
jeffdiclemente: this can be simplified by removing ```std::map<std::string, std::string>()``` since no configuration is required.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    std::vector<std::shared_ptr<Bundle>> bundles = frameworkCtx->GetBundles();
+    LDAPFilter ldap( "(bundle.testproperty=YES)" );
+    
+    int actualMatchCount = 0;
+    int expectedMatchCount = 0;
+    
+    std::string bundleName;
+    Any bundleTestProperty;
+    
+    for (std::shared_ptr<Bundle> bundle : bundles)
```
jeffdiclemente: I'd suggest using a ```const auto&``` for the range declaration to leverage type deduction and maintain const correctness. The ```bundle``` object will not be modified in the loop and ```LDAPFilter::Match(const Bundle& )``` expects a ```const&```.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+{
+    std::vector<std::shared_ptr<Bundle>> bundles = frameworkCtx->GetBundles();
+    LDAPFilter ldap( "(service.testproperty=YES)" );
+    
+    BundleContext* thisBundleCtx;
+    
+    int actualMatchCount = 0;
+    int expectedMatchCount = 1;
+
+    for (std::shared_ptr<Bundle> bundle : bundles)
```
jeffdiclemente: This loop appears to be unnecessary. Correct me if I'm wrong; the point of this test looks to be to test that query service properties work. 
If that's that case, its inefficient to iterate over all the bundles which don't export a service (unless there is a requirement I'm missing). The same result can be reached by using ```BundleContext::GetBundle(const std::string& name)```, passing in the name of the bundle exporting the service.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+        
+            if (sr && ldap.Match(sr))
+            {
+                actualMatchCount++;
+            }
+            bundle->Stop();
+        }
+    }
+    
+    std::string test_description = "Test LDAP query of service properties (expected # of matches: " + std::to_string(expectedMatchCount) + "  actual # of matches: " + std::to_string(actualMatchCount)+ ")";
```
jeffdiclemente: nit-pick: spacing of the ```+``` in ```std::to_string(actualMatchCount)+ ")";```

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    InstallTestBundle(frameworkCtx, "TestBundleLQ_4");
+    InstallTestBundle(frameworkCtx, "TestBundleLQ_5");
+    
+    TestLDAPQueryBundleMetaData(frameworkCtx);
+    TestLDAPQueryServiceProperties(frameworkCtx);
+    
+    framework->Stop();
+    
+    US_TEST_END()
+}
```
jeffdiclemente: nit-pick: no newline at end of the file.

core/include/usBundle.h
```diff
@@ -398,6 +409,7 @@ class US_Core_EXPORT Bundle : public std::enable_shared_from_this<Bundle>
   friend class BundleRegistry;
   friend class ServiceReferencePrivate;
   friend class Framework;
+  friend class LDAPFilter;
```
jeffdiclemente: can be removed if ```LDAPFilter::Match(...)``` uses ```Bundle::GetProperties()```

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    std::shared_ptr<Framework> framework = factory.NewFramework(std::map<std::string, std::string>());
+    framework->Start();
+    
+    BundleContext* frameworkCtx = framework->GetBundleContext();
+    
+    InstallTestBundle(frameworkCtx, "TestBundleLQ_1");
+    InstallTestBundle(frameworkCtx, "TestBundleLQ_2");
+    InstallTestBundle(frameworkCtx, "TestBundleLQ_3");
+    InstallTestBundle(frameworkCtx, "TestBundleLQ_4");
+    InstallTestBundle(frameworkCtx, "TestBundleLQ_5");
```
karthikreddy09: Why 5 bundles ? Consider adding the absolute minimum number of bundles required to satisfy the test cases

core/include/usBundle.h
```diff
@@ -244,6 +245,16 @@ class US_Core_EXPORT Bundle : public std::enable_shared_from_this<Bundle>
...
 
   /**
+   * Returns this bundle's Manifest properties as key/value pairs.
+   *
+   * @return A map containing this bundle's Manifest properties as 
+   *         key/value pairs.
+   *
+   * @sa \ref MicroServices_BundleProperties
+   */
+  std::map<std::string, Any> GetProperties() const;
```
karthikreddy09: Any reason why we use an ordered map here and an unordered map for service properties ?

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    
+    BundleContext* thisBundleCtx;
+    
+    int actualMatchCount = 0;
+    int expectedMatchCount = 1;
+
+    for (std::shared_ptr<Bundle> bundle : bundles)
+    {
+        
+        if (bundle->GetBundleId() != 0)
```
karthikreddy09: spacing

core/src/CMakeLists.txt
```diff
@@ -65,13 +61,19 @@ set(_srcs
...
 set(_private_headers
   util/usFrameworkPrivate.h
+  util/usLDAPExpr_p.h
+  util/usProperties_p.h
   util/usThreads_p.h
   util/usUtils_p.h
   util/usWaitCondition_p.h
 
   service/usServiceHooks_p.h
+  service/usServiceListenerEntry_p.h
```
karthikreddy09: Is this change related to LDAP issue ? if not i'd suggest creating another branch for this issue. I believe this falls under issue#67

core/src/util/usLDAPExpr.cpp
kevinleeMW: I searched online, and found that it seems not possible with github UI. 
Summary of changes:
- use usProperties class instead of usServicePropertiesImpl class.

core/src/util/usLDAPExpr_p.h
kevinleeMW: I searched online, and found that it seems not possible with github UI. 
Summary of changes:
- use usProperties class instead of usServicePropertiesImpl class.

core/src/util/usLDAPFilter.cpp
kevinleeMW: I searched online, and found that it seems not possible with github UI. 
Summary of changes:
- use usProperties class instead of usServicePropertiesImpl class.
- Added "Match(const Bundle& bundle)" method to enable LDAP query on bundle properties

core/src/util/usLDAPExpr.cpp
```diff
@@ -0,0 +1,852 @@
...
+{
+  return LDAPExpr(filter).Evaluate(pd, false);
+}
+
+bool LDAPExpr::Evaluate( const Properties& p, bool matchCase ) const
+{
+  if ((d->m_operator & SIMPLE) != 0)
+  {
+    // try case sensitive match first
+    int index = p.FindCaseSensitive(d->m_attrName);
```
karthikreddy09: This is wierd ... why do attempt a casesentive match at all if matchCase is false.

core/src/util/usLDAPFilter.cpp
karthikreddy09: +1. Bundle should have a GetProperties public method instead

core/include/usBundle.h
kevinleeMW: I will address this issue in the second iteration of this task (incl. optimization, etc.)

core/src/util/usLDAPExpr.cpp
saschazelzer: Some files like `usLDAPExpr.cpp` are still present in their original location (core/src/service). I guess this is the reason why the changes are not schown as a move. Could you remove the files, squash the commits and push-force the branch again?

core/src/util/usProperties.cpp
saschazelzer: This is my old code, where I used one-line if statements without braces. Feel free to to add them though.

core/include/usBundle.h
saschazelzer: We could probably rename the `ServiceProperties` typedef (a std::unordered_map) to `Dictionary` and use this consistently.

core/src/util/usLDAPExpr.cpp
saschazelzer: This was optimized for the general case were it is assumed that people use filters on entities that contain the attribute names as property keys with the same case and the filter compares property values most of the time. Case mismatches are assumed to be typos or "laziness".

core/src/CMakeLists.txt
kevinleeMW: These changes were made, because some of the files were moved; and made additional changes as I realized that some of the files were missing from this list.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+{
+  US_TEST_BEGIN("LDAPQueryTest");
+    
+  FrameworkFactory factory;
+  std::shared_ptr<Framework> framework = factory.NewFramework();
+  framework->Start();
+    
+  BundleContext* frameworkCtx = framework->GetBundleContext();
+    
+  InstallTestBundle(frameworkCtx, "TestBundleLQ");
```
jeffdiclemente: ```InstallTestBundle``` returns the ```std::shared_ptr<Bundle>``` it installed. This makes the call to ```frameworkCtx->GetBundle``` on the next line superfluous.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+  bundle->Stop();
+}
+
+int usLDAPQueryTest(int /*argc*/, char* /*argv*/[])
+{
+  US_TEST_BEGIN("LDAPQueryTest");
+    
+  FrameworkFactory factory;
+  std::shared_ptr<Framework> framework = factory.NewFramework();
```
jeffdiclemente: prefer using ```auto```

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+  framework->Start();
+    
+  BundleContext* frameworkCtx = framework->GetBundleContext();
+    
+  InstallTestBundle(frameworkCtx, "TestBundleLQ");
+  const auto& bundle = frameworkCtx->GetBundle("TestBundleLQ");
```
jeffdiclemente: why ```const auto&```?

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+int usLDAPQueryTest(int /*argc*/, char* /*argv*/[])
+{
+  US_TEST_BEGIN("LDAPQueryTest");
+    
+  FrameworkFactory factory;
+  std::shared_ptr<Framework> framework = factory.NewFramework();
+  framework->Start();
+    
+  BundleContext* frameworkCtx = framework->GetBundleContext();
```
jeffdiclemente: You could remove the variable and place ```framework->GetBundleContext()``` inside ```InstallTestBundle``` (assuming ```frameworkCtx->GetBundle``` is removed).

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+  // Exact string match of both key and value
+  US_TEST_CONDITION(ldapMatchCase.Match(sr), " Evaluating LDAP expr: " + ldapMatchCase.ToString());
+
+  // Testing case-insensitive key (should still pass)
+  US_TEST_CONDITION(ldapKeyMismatchCase.Match(sr), " Evaluating LDAP expr: " + ldapKeyMismatchCase.ToString());
+
+  // Testing case-insensitive value (should fail)
+  US_TEST_CONDITION(!ldapValueMismatchCase.Match(sr), " Evaluating LDAP expr: " + ldapValueMismatchCase.ToString());
+
+  bundle->Stop();
```
jeffdiclemente: What's the intended behavior of ```LDAPFilter::Match(ServiceReference)``` once the bundle has stopped (and the service reference is no longer valid)?
Does this behavior need to be tested?

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+#include "usTestUtils.h"
+#include "usTestingMacros.h"
+#include "usProperties_p.h"
+#include "usServiceReferenceBasePrivate.h"
+
+
+using namespace us;
+
+void TestLDAPFilterMatchBundle(const std::shared_ptr<Bundle> bundle)
```
jeffdiclemente: use ```const std::shared_ptr<Bundle>& bundle``` instead. That way there is no reference counting (which is unnecessary in this case and expensive) at function entry and exit.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+  US_TEST_CONDITION(ldapMatchCase.Match(*bundle), " Evaluating LDAP expr: " + ldapMatchCase.ToString());
+
+  // Testing case-insensitive key (should still pass)
+  US_TEST_CONDITION(ldapKeyMismatchCase.Match(*bundle), " Evaluating LDAP expr: " + ldapKeyMismatchCase.ToString());
+
+  // Testing case-insensitive value (should fail)
+  US_TEST_CONDITION(!ldapValueMismatchCase.Match(*bundle), " Evaluating LDAP expr: " + ldapValueMismatchCase.ToString());
+}
+
+void TestLDAPFilterMatchServiceReferenceBase(std::shared_ptr<Bundle> bundle)
```
jeffdiclemente: use ```const std::shared_ptr<Bundle>& bundle``` instead. That way there is no reference counting (which is unnecessary in this case and expensive) at function entry and exit.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+ limitations under the License.
+ 
+ =============================================================================*/
+
+#include <usFrameworkFactory.h>
+#include <usFramework.h>
+
+#include <usBundle.h>
+#include <usLDAPFilter.h>
+#include <usLog.h>
```
jeffdiclemente: is this needed? I don't see any invocations of the logging macros.

core/test/usLDAPQueryTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+#include "usTestUtils.h"
+#include "usTestingMacros.h"
+#include "usProperties_p.h"
+#include "usServiceReferenceBasePrivate.h"
+
```
jeffdiclemente: I don't believe you want to ```#include``` private headers here (```usProperties_p.h``` and ```usServiceReferenceBasePrivate.h```). The public headers should be sufficient.

core/test/usLDAPQueryTest.cpp
kevinleeMW: Quoting OSGi spec (Section 5.2.1):
"A ````ServiceReference```` object is valid only as long as the service is registered. However, its properties must remain available as long as the ````ServiceReference```` object exists."

From this, I think that ```LDAPFilter::Match(ServiceReference)```` should behave the same even after the bundle has stopped. 

Perhaps, an additional test condition can be added after the bundle stops to make sure that querying service properties via a ````ServiceReferenceBase```` object still works.

What do you guys think?

jeffdiclemente: I think it sounds pretty clear that service properties can be queried after the service they reference has been unregistered (e.g. on bundle stop).

What happens if you duplicate the four ```US_TEST_CONDITION``` macros after the ```bundle->Stop();``` and re-run the test?

kevinleeMW: The first of the four ````US_TEST_CONDITION```` macros will fail, because it checks to make sure that the ServiceReference object is valid, and it no longer is after the ````bundle->Stop();````. 

Including the second test macro (that does the querying and expects a match) after the ````bundle->Stop();```` seems sufficient, as it will be testing that the service properties are still available (for LDAP queries).

kevinleeMW:
I went through and addressed all of the feedback from Jeff, Abhinay and Sascha. Some of the suggestions will be tracked in Issue #85 as a follow-up task.

Is there any other flags from the pull request, @jeffdiclemente @saschazelzer @karthikreddy09 ?

jeffdiclemente:
I have no additional flags.
Thanks for creating #85 to further optimize.

saschazelzer:
Looks good to me, +1 for merging.

karthikreddy09:
Ship It !

PR #77, opened 2016-02-22
55 resource compiler port to cpp

karthikreddy09:
Removed c-style functions in usResourceCompiler.cpp and updated the option names.


saschazelzer:
Great work! Do you think it can be merged to development as well?

Issue #56, opened 2015-11-13
unit test failures on Mac OS X 10.10.4/Xcode 6.4


jeffdiclemente:
Two unit tests fail (usBundleManifestTest and usLDAPFilterTest ) due to what appears to be how Apple clang handles comparing types.

It appears that Apple clang compares the object addresses and not the string representaiton of the type.

saschazelzer:
This keeps coming back... :-(

Commit 611f732324912b77c7a5c971e5ec92433daf277f removed a workaround for this (also needed for older gcc versions). Look for `US_GCC_RTTI_WORKAROUND_NEEDED`. I guess I was too eager to remove this as I never felt really comfortable with this workaround.

Could you provide the versions of clang, ld and libc++ you are using?

Could you also search with `nm -om` (I think) through all object files for `weak private external typeinfo` entries that have a corresponding typeinfo entry in another object file, but with `weak external typeinfo`?

I found [1] a valuable blog post for this kind of issue.

[1] http://www.russellmcc.com/posts/2013-08-03-rtti.html

jeffdiclemente:
Thanks for the information, that's helpful.


as for the versions...

% xcrun clang --version
Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin14.4.0
Thread model: posix

% xcrun ld -v
@(#)PROGRAM:ld  PROJECT:ld64-242.2
configured to support archs: armv6 armv7 armv7s arm64 i386 x86_64 x86_64h armv6m armv7m armv7em
LTO support using: LLVM version 3.6.0svn

not sure what the exact version of libc++ is, but it'll be whichever ships with Xcode 4.6.0.

I'll take a look through all the object files and let you know what I find.

I've already fixed part of the problem by making the Any class have default visibility. std::map<std::string, Any> is still failing the type comparison though.

jeffdiclemente:
After working through this issue some more, here is what I know.

There is still a puzzle as to why the typeinfo symbols are handled the way they are (read on for details about what I mean).

I'm using Xcode 4.6.0 (clang 6.1.0).

First, I modified the ```Any``` class to have default visibility (i.e. added ```US_Core_EXPORT``` to the declaration) and moved the definitions of non-template functions into a separate source file. I also marked all template functions with ```US_Core_EXPORT```. 
[...]

PR #82, opened 2016-02-29
fix mac os x test failures

jeffdiclemente:
Fixes: #56

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -318,7 +318,15 @@ else()
     usFunctionCheckCompilerFlags(-Wno-mismatched-tags US_CXX_FLAGS)
   endif()
 
-  usFunctionCheckCompilerFlags("-fvisibility=hidden -fvisibility-inlines-hidden" _have_visibility)
+  # Issue #56: Workaround a clang compiler issue (https://llvm.org/bugs/show_bug.cgi?id=26749).
+  # The typeinfo symbols do not have reliable symbol visibility. Instead of exporting all
+  # symbols only export the typeinfo symbols by default. See
+  # https://developer.apple.com/library/mac/technotes/tn2185/_index.html#//apple_ref/doc/uid/DTS10004200-CH1-SUBSECTION2
+  if(US_COMPILER_CLANG OR US_COMPILER_APPLE_CLANG)
+    usFunctionCheckCompilerFlags("-fvisibility-ms-compat" _have_visibility)
+  else()
+    usFunctionCheckCompilerFlags("-fvisibility=hidden -fvisibility-inlines-hidden" _have_visibility)
+  endif()
   if(_have_visibility)
     set(US_HAVE_VISIBILITY_ATTRIBUTE 1)
     set(US_CXX_FLAGS "${US_CXX_FLAGS} ${_have_visibility}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -319,8 +319,10 @@ else()
   endif()
 
   # Issue #56: Workaround a clang compiler issue (https://llvm.org/bugs/show_bug.cgi?id=26749).
-  # The typeinfo symbols do not have reliable symbol visibility. Instead of exporting all
-  # symbols only export the typeinfo symbols by default. See
+  # The typeinfo symbols do not have reliable symbol visibility. Instead of exporting only
+  # symbols and associated typeinfo symbols which have explicit "default" visibility,
+  # export all typeinfo symbols by default and maintain hidden symbol visibility for all other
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -318,7 +318,15 @@ else()
     usFunctionCheckCompilerFlags(-Wno-mismatched-tags US_CXX_FLAGS)
   endif()
 
-  usFunctionCheckCompilerFlags("-fvisibility=hidden -fvisibility-inlines-hidden" _have_visibility)
+  # Issue #56: Workaround a clang compiler issue (https://llvm.org/bugs/show_bug.cgi?id=26749).
+  # The typeinfo symbols do not have reliable symbol visibility. Instead of exporting all
+  # symbols only export the typeinfo symbols by default. See
+  # https://developer.apple.com/library/mac/technotes/tn2185/_index.html#//apple_ref/doc/uid/DTS10004200-CH1-SUBSECTION2
+  if(US_COMPILER_CLANG OR US_COMPILER_APPLE_CLANG)
```
saschazelzer: Is it intended to use `-fvisiility-ms-compat` with Clang on non-Apple platforms as well? If no, can we just do `if(US_COMPILER_APPLE_CLANG)` instead? Otherwise, on what Linux platform and Clang version did you observe problems?

jeffdiclemente: good catch. ```-fvisibility-ms-compat``` is only meant to be used with clang on Mac OS X. ```-fvisibility-ms-compat``` must be used for both Apple's Clang (which is shipped with Xcode) and clang from [the LLVM website](http://llvm.org/releases/download.html).
I thought that ```US_COMPILER_CLANG OR US_COMPILER_APPLE_CLANG``` would cover both these cases, but did not consider clang on Linux.

Does ```US_COMPILER_CLANG``` represent clang only on Linux?
Is only using ```US_COMPILER_APPLE_CLANG``` sufficient?

saschazelzer: The `US_COMPILER_*` variables denote only the compiler itself, with no reference to the platform they run on. As far as I know, the only officially supported compiler on MacOS since a few years is Apples Clang build which is different from upstream and also has a different versioning scheme. The LLVM website does not offer binaries for MacOS as far as I can see.

Since we still support CMake < 3.0, the variable `US_COMPILER_CLANG` will be set instead of `US_COMPILER_APPLE_CLANG` when using CMake 2.8.12 (the latter will be set to `1` when using a newer CMake version). This means we would actually need to use

`if(APPLE AND (US_COMPILER_CLANG OR US_COMPILER_APPLE_CLANG))`

dr4g0nm4g3: Only a few LLVM versions [have pre-built Mac OS X binaries](http://llvm.org/releases/download.html#3.7.0).

Thanks, I will modify the CMake code to handle both clang compilers only on Mac OS X.

jeffdiclemente: Only a few LLVM versions [have pre-built Mac OS X binaries](http://llvm.org/releases/download.html#3.7.0).

Thanks, I will modify the CMake code to handle both clang compilers only on Mac OS X.

jeffdiclemente:
Unless anyone has any flags, I plan on merging this on Monday, March 7th.

CMakeLists.txt
```diff
@@ -318,7 +318,16 @@ else()
     usFunctionCheckCompilerFlags(-Wno-mismatched-tags US_CXX_FLAGS)
   endif()
 
-  usFunctionCheckCompilerFlags("-fvisibility=hidden -fvisibility-inlines-hidden" _have_visibility)
+  # Issue #56: Workaround a clang compiler issue (https://llvm.org/bugs/show_bug.cgi?id=26749).
+  # The typeinfo symbols do not have reliable symbol visibility. Instead of exporting all
```
saschazelzer: Maybe I misunderstand things here, but this reads like we were exporting all symbols before. Shouldn't this rather read like: Instead of exporting only symbols (and their associated typeinfo symbols) which have an explicit "default" visibility, export all typeinfo symbols by default but keep all other  symbols hidden (unless explicitly overridden).

saschazelzer: You are probably referring to the previous workaround, right? But without that knowledge, I think the comment is confusing when looking at the else branch below.

saschazelzer:
For the first time, this looks like a work-around without too much impact - great!

+1 for merging.

CMakeLists.txt
jeffdiclemente: You are right. the comment is ambiguous. I'll fix it up. Thanks!

PR #87, opened 2016-03-24
Fixed resource compiler to only call us_tempfile function when not in append mode.

karthikreddy09:
Avoids file drool on posix platforms


jeffdiclemente:
looks good. ship it!

PR #90, opened 2016-04-14
88 find activator inside executable bundles

karthikreddy09:
Remove the requirement that bundle inside an executable has to be named "main".

### DIFF
--- a/core/src/bundle/usBundleUtils.cpp
+++ b/core/src/bundle/usBundleUtils.cpp
@@ -32,23 +32,27 @@
 #define _GNU_SOURCE
 #endif
 #include <dlfcn.h>
+#include <sys/stat.h>
+#if defined(__APPLE__)
+#include <sys/param.h>
+#include <stdlib.h>
+#include <mach-o/dyld.h>
+#else
+#include <limits.h>
+#include <unistd.h>
+#endif
 #elif _WIN32
 #include <windows.h>
 #endif
 
 
-namespace us {
 
-namespace {
-#ifdef US_BUILD_SHARED_LIBS
-  const bool sharedLibMode = true;
-#else
-  const bool sharedLibMode = false;
-#endif
-}
+namespace us {
 
 namespace BundleUtils
 {
+
+bool IsCurrentExecutable(const std::string& filepath); // forward declaration.
 
 #ifdef __GNUC__
 
@@ -72,7 +76,8 @@ void* GetSymbol_impl(const BundleInfo& bundleInfo, const char* symbol)
   dlerror();
 
   void* selfHandle = nullptr;
-  if (!sharedLibMode || bundleInfo.name == "main")
+  
+  if (IsCurrentExecutable(bundleInfo.location))
   {
     // Get the handle of the executable
     selfHandle = dlopen(0, RTLD_LAZY);
@@ -102,7 +107,71 @@ void* GetSymbol_impl(const BundleInfo& bundleInfo, const char* symbol)
   }
   return nullptr;
 }
+  
 
+#if defined(__APPLE__)
+  
+std::string GetExecutablePath()
+{
+  unsigned int bufSize = MAXPATHLEN;
+  char pathBuf[bufSize];
+  if (_NSGetExecutablePath(pathBuf, &bufSize) != 0)
+  {
+    US_DEBUG << "GetExecutablePath() - _NSGetExecutablePath failed";
+    pathBuf[0] = '\0';
+  }
+  // the returned path may not be an absolute path
+  return std::string(pathBuf);
+}
+  
+#else
+
+std::string GetExecutablePath()
+{
+  char pathBuf[PATH_MAX];
+  ssize_t len = readlink("/proc/self/exe", pathBuf, sizeof(pathBuf));
+  if (len < 0)
+  {
+    US_DEBUG << "GetExecutablePath() " << GetLastErrorStr();
+    pathBuf[0] = '\0';
+  }
+  else
+  {
+    pathBuf[len] = '\0';
+  }
+  return std::string(pathBuf);
+}
+  
+#endif
+  
+struct FileInfo
+{
+  std::string path;
+  ino_t stat_ino;
+  dev_t stat_dev;
+  
+  FileInfo() : path(""), stat_ino(0), stat_dev(0) {}
+  
+  explicit FileInfo(const std::string& inPath) : path(inPath), stat_ino(0), stat_dev(0)
+  {
+    if(!path.empty())
+    {
+      struct stat pathStat;
+      if(stat(path.c_str(), &pathStat) == 0)
+      {
+        stat_ino = pathStat.st_ino;
+        stat_dev = pathStat.st_dev;
+      }
+    }
+  }
+} execInfo(GetExecutablePath());
+
+inline bool operator==(const FileInfo& lhs, const FileInfo& rhs)
+{
+  return (lhs.stat_dev == rhs.stat_dev &&
+          lhs.stat_ino == rhs.stat_ino);
+}
+  
 #elif _WIN32
 
 std::string GetLibraryPath_impl(void *symbol)
@@ -118,7 +187,7 @@ std::string GetLibraryPath_impl(void *symbol)
   }
 
   char bundlePath[512];
-  if (GetModuleFileName(handle, bundlePath, 512))
+  if (GetModuleFileNameA(handle, bundlePath, 512))
   {
     return bundlePath;
   }
@@ -127,24 +196,82 @@ std::string GetLibraryPath_impl(void *symbol)
   return "";
 }
 
+std::string GetExecutablePath()
+{
+  char pathBuf[1024];	// assuming this is a large enough buffer.
+  if (GetModuleFileNameA(nullptr, pathBuf, sizeof(pathBuf)) == 0 || GetLastError() == ERROR_INSUFFICIENT_BUFFER)
+  {
+    US_DEBUG << "GetModuleFileName failed" << GetLastErrorStr();
+    pathBuf[0] = '\0';
+  }
+  return std::string(pathBuf);
+}
+  
+struct FileInfo {
+  std::string path;
+  DWORD dwVolumeSerialNumber;
+  DWORD nFileIndexHigh;
+  DWORD nFileIndexLow;
+  
+  FileInfo() : path(""), dwVolumeSerialNumber(0), nFileIndexHigh(0), nFileIndexLow(0) {}
+  
+  explicit FileInfo(const std::string& inPath) : path(inPath), dwVolumeSerialNumber(0), nFileIndexHigh(0), nFileIndexLow(0)
+  {
+    if(!path.empty())
+    {
+      HANDLE hFile = CreateFileA(path.c_str(),
+                                 0,
+                                 0,
+                                 NULL,
+                                 OPEN_EXISTING,
+                                 FILE_ATTRIBUTE_NORMAL,
+                                 NULL);
+      if(hFile != INVALID_HANDLE_VALUE)
+      {
+        BY_HANDLE_FILE_INFORMATION fileinfo;
+        ZeroMemory(&fileinfo, sizeof(BY_HANDLE_FILE_INFORMATION));
+        if(GetFileInformationByHandle(hFile, &fileinfo))
+        {
+          dwVolumeSerialNumber = fileinfo.dwVolumeSerialNumber;
+          nFileIndexHigh = fileinfo.nFileIndexHigh;
+          nFileIndexLow = fileinfo.nFileIndexLow;
+        }
+        CloseHandle(hFile);
+      }
+      else
+      {
+        US_DEBUG << "CreateFile() - failed" << GetLastErrorStr();
+      }
+    }
+  }
+} execInfo(GetExecutablePath());
+
+inline bool operator==(const FileInfo& lhs, const FileInfo& rhs)
+{
+  return (lhs.dwVolumeSerialNumber == rhs.dwVolumeSerialNumber &&
+          lhs.nFileIndexHigh == rhs.nFileIndexHigh &&
+          lhs.nFileIndexLow == rhs.nFileIndexLow);
+}
+
 void* GetSymbol_impl(const BundleInfo& bundleInfo, const char* symbol)
 {
   HMODULE handle = nullptr;
-  if (!sharedLibMode || bundleInfo.name == "main")
+  
+  if (IsCurrentExecutable(bundleInfo.location))
   {
-    handle = GetModuleHandle(nullptr);
+    handle = GetModuleHandleA(nullptr);
   }
   else
   {
-    handle = GetModuleHandle(bundleInfo.location.c_str());
+    handle = GetModuleHandleA(bundleInfo.location.c_str());
   }
-
+  
   if (!handle)
   {
-    US_DEBUG << "GetSymbol_impl():GetBundleHandle() " << GetLastErrorStr();
+    US_DEBUG << "GetSymbol_impl():GetModuleHandle() " << GetLastErrorStr();
     return nullptr;
   }
-
+  
   void* addr = (void*)GetProcAddress(handle, symbol);
   if (!addr)
   {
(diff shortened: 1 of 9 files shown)


karthikreddy09:
@jeffdiclemente @saschazelzer @kevinleeMW Please review the changes to fix issue#88

core/test/bundles/execWithStaticBundle/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+  SOURCES usTestStaticBundle.cpp
+  RESOURCES manifest.json
+  RESOURCES_ROOT resources_static
+  SKIP_BUNDLE_LIST)
+
+set(BUILD_SHARED_LIBS ${US_BUILD_SHARED_LIBS})
+
+add_executable(usTestExecutableWithStaticBundle main.cpp)
+
+#MESSAGE( STATUS "CppMicroServices_LIBRARIES:         " ${Core_TARGET})
```
jeffdiclemente: remove comment if it has no use.

core/test/bundles/execWithStaticBundle/main.cpp
```diff
@@ -0,0 +1,51 @@
...
+  framework->Start();
+  BundleContext* frameworkCtx = framework->GetBundleContext();
+  auto bundle = frameworkCtx->InstallBundle(BIN_PATH + DIR_SEP + "usTestExecutableWithStaticBundle" + EXE_EXT + "/TestStaticBundle");
+  try 
+  {
+	  if (bundle == nullptr)
+		  throw std::runtime_error("Static bundle not found in the executable");
+	  if (bundle->GetName() != "TestStaticBundle")
+		  throw std::runtime_error("Check bundle name failed - " + bundle->GetName());
+    bundle->Start();
```
jeffdiclemente: formatting; indentations are off

core/test/bundles/execWithStaticBundle/main.cpp
```diff
@@ -0,0 +1,51 @@
...
+    auto service = bundle->GetBundleContext()->GetServiceReference<TestBundleService>();
+    if (!service)
+    {
+      throw std::runtime_error("Could not find service from the statically linked bundle");
+    }
+  }
+  catch (const std::exception& ex)
+  {
+	  std::cerr << ex.what() << std::endl;
+	  return -1;
```
jeffdiclemente: indentations

core/src/bundle/usBundleUtils.cpp
```diff
@@ -127,24 +196,82 @@ std::string GetLibraryPath_impl(void *symbol)
...
+  return std::string(pathBuf);
+}
+  
+struct FileInfo {
+  std::string path;
+  DWORD dwVolumeSerialNumber;
+  DWORD nFileIndexHigh;
+  DWORD nFileIndexLow;
+  
+  FileInfo() : path(""), dwVolumeSerialNumber(0), nFileIndexHigh(0), nFileIndexLow(0) {}
```
jeffdiclemente: consider removing the default ctor. From the usage of FileInfo, it doesn't appear that a FileInfo object with an empty path is allowed.

core/src/bundle/usBundleUtils.cpp
```diff
@@ -183,6 +310,11 @@ void* GetSymbol(const BundleInfo& bundle, const char* symbol)
 {
   return GetSymbol_impl(bundle, symbol);
 }
+  
+bool IsCurrentExecutable(const std::string& filepath)
+{
+  return execInfo.path == filepath || execInfo == FileInfo(filepath); // Avoid constructing the FileInfo object if paths match
```
jeffdiclemente: A comment about why `|| execInfo == FileInfo(filepath);` is here would be useful.

core/test/bundles/execWithStaticBundle/main.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Test point to ensure the statically linked bundle is installed properly.
+ */
+int main(int ac, char **av)
+{
+  auto framework = FrameworkFactory().NewFramework();
+  framework->Start();
+  BundleContext* frameworkCtx = framework->GetBundleContext();
+  auto bundle = frameworkCtx->InstallBundle(BIN_PATH + DIR_SEP + "usTestExecutableWithStaticBundle" + EXE_EXT + "/TestStaticBundle");
```
jeffdiclemente: shouldn't the ```/TestStaticBundle``` be removed? We don't support this anymore for ```InstallBundle```. In addition, isn't this test supposed to check that installing an EXE will install any statically linked bundles?

core/test/bundles/execWithStaticBundle/main.cpp
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }
+  catch (const std::exception& ex)
+  {
+	  std::cerr << ex.what() << std::endl;
+	  return -1;
+  }
+  framework->Stop();
+  (void)ac;
+  (void)av;
```
jeffdiclemente: this can be avoided by not naming the parameters in ```int main```.

core/test/bundles/execWithStaticBundle/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
+
+# This test is only applicable when build mode is shared libraries. 
+# The test point usExecutableBundleTest in usCoreDriver is sufficeint for static library builds.
+if(US_BUILD_SHARED_LIBS)
+
+set(BUILD_SHARED_LIBS 0)
+usFunctionCreateTestBundleWithResources(TestStaticBundle
```
jeffdiclemente: Suggestion; I think using "TestStaticallyLinkedBundle" is a more descriptive name to use than "TestStaticBundle". It uses the nomenclature that we want users to use (i.e. statically linked bundle).

core/test/usExecutableBundleTest.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * Test point to ensure the executable bundle is installed properly.
+ */
+int usExecutableBundleTest(int /*argc*/, char* /*argv*/[])
+{
+  US_TEST_BEGIN("ExecutableBundleTest");
+  
+  auto framework = FrameworkFactory().NewFramework();
+  framework->Start();
+  BundleContext* frameworkCtx = framework->GetBundleContext();
+  auto bundle = frameworkCtx->InstallBundle(BIN_PATH + DIR_SEP + "usCoreTestDriver" + EXE_EXT + "/main");
```
jeffdiclemente: adding ```/<bundle name>``` is no longer supported for ```InstallBundle```

core/test/bundles/execWithStaticBundle/main.cpp
karthikreddy09: We still need the bundle name to install. The change you are referring to is only in the EmbeddedBundles.

core/test/usExecutableBundleTest.cpp
karthikreddy09: Nop, we still need the bundle name in development branch,

core/test/bundles/execWithStaticBundle/CMakeLists.txt
karthikreddy09: @jeffdiclemente 
How about "TestBundleImportedByExec" ? this is similar to the name of the statically linked bundle in "TestBundleB".

jeffdiclemente: what about "TestBundleImportedByExe" ?

jeffdiclemente:
looks good. +1
:+1:

saschazelzer:
Thanks for working on this, I have two issues that I'd like to discuss:

1. There is a lot of platform specific code there, and as a consequence there is also the `FileInfo` struct business to cope with symlinks etc. I believe (hope) we can instead use the same API we already use for retrieving the bundle location to initialize a variable holding the executable location (which will have the same "form" as the later determined module locations):
 - On Windows you are already using GetModuleName() in both cases, so this is fine and I'd think we do not need to take care of NTFS reparse points or similar stuff. Just comparing the strings should do.
 - On Linux and MacOS, we already use dladdr for looking up shared object locations. To get the executable path we can just do:

   ```
void* h = dlopen(nullptr, RTLD_LAZY);
void* s = dlsym(h, "main");
std::string exec_path{};
Dl_info info{};
if (dladdr(s, &info)) exec_path = info.dli_fname;
```
If the executable was not compiled with `rdynamic`, the lookup would not work, but then also `us::GetBundleContext` would not work from within the executable. And because we use the same mechanism, simple string locations should again be enough. What do you think?

2. Could you explain why you change some Windows API functions to their explicit ANSI versions? If at all, I believe we should use the W versions, but then consistently everywhere. (as explained for example here: http://www.utf8everywhere.org)

**Comment by karthikreddy09** [comment]
1. I like the idea of using Dl_info for both Mac & Linux. However I am not very sure if we can discard the possibility of symlinks in the path. We are comparing a user provided string against a string(path) that the program detected. User may provide a path with symlink.

For example, suppose
The executable is located at "/mw/dev/sandbox/user/runme". 
"/sandbox" is a simlink to "/mw/dev/sandbox"
In the code, a call to InstallBundle("/sandbox/user/runme") should work just fine.
If we rely on simple string comparisons, this will fail because the input parameter to GetSymbol_Impl would have bundle.location as "/sandbox/user/runme" where as the program detects the location as "mw/dev/sandbox/user/runme".

 Does this example make sense ?

2. I prefer to use the explicit unicode (W) flavors of the APIs everywhere (which as you mention needs a larger refactor). I used the explicit ANSI versions because I felt it made sense readability wise to know the exact function that is called underneath. It would also mean even if UNICODE is turned ON, that particular code wont generate a compile error (I realized this is actually bad. Generates a runtime failure instead of a compile error. Which is the case today if std::string is UTF8 encoded). 

The more I think about it, It feels like using the generic names (MACROS) is the way to go. It would be consistent with the rest of the project and if we move to W flavors in the future the whole codebase needs to change anyway. I am OK to replace the A flavors with the generic ones.

**Comment by saschazelzer** [comment]
1. In principal, your example makes sense. First I was only looking at the `GetBundleContext()` function and there, no user supplied location is involved. It uses the `GetLibraryPath()` function to get the path for the calling library, which is then passed to `GetSymbol()`. We should be able to use the same technique in the `Bundle::Start` method, which is the only other user of `GetSymbol()`, i.e. getting the library location with a call to `dladdr` using e.g. the `_us_set_bundle_context_instance_<bundle-name>` symbol. This would enable us to rely on simple string comparisons, right?

2. Thanks for the details. Reverting to a consistent state for now would be good. We should move the code base later to the W versions with the appropriate code changes as needed.

**Comment by karthikreddy09** [comment]
@saschazelzer Sorry for the delayed response

1. In case of `GetBundleContext`, we pass a symbol adress to `GetLibraryPath` which uses `dladdr` to retrieve the path of the shared library. we then use the path to find another symbol

 In Bundle::Start(), we don't have a symbol address. we can construct the function name `_us_set_bundle_context_instance_<bundle-name>` as a string but we don't know its addess. So we cannot call `GetLibraryPath` and have to depend on the user provided path.

**Comment by saschazelzer** [comment]
I am trying to remember and understand what I wrote above. I think your analysis above is correct, but I still think we can pull this off without relying on user supplied paths.

Looking at the `Bundle::Start()` method we have the handle to the shared library of the bundle being started at our hand (from the `SharedLibrary` class. So if we would have an overload of `BundleUtils::GetSymbol()` which takes a library handle as an argument (instead of a library location), it could then just work with that (avoiding the executable checks). The other case where a string location is supplied to `BundleUtils::GetSymbol()` is from the `GetBundleContext()` method, but here the string is not user supplied and it could be safely compared with a stored string containing the executable path. Does it make sense?

**Comment by karthikreddy09** [comment]
This strategy works for shared libraries .... how about for executables that are bundles ? What if the user calls `bundlecontext->InstallBundle("/path/to/current_executable")` ? In this case we would not have a handle in `Bundle::Start()` because `IsSharedLibrary` fails for this path. 

We can workaround the issue by adding a special condition in the proposed `GetSymbol` function. When a `null` handle is passed to `GetSymbol`, it could assume the user meant the current executable. This sounds more like a hack though.

**Comment by saschazelzer** [comment]
You are right, supplying the executable path as an argument to `InstallBundle` would not work that way. I was also thinking that having to know the executable path isn't a good thing anyways. If, by convention, an empty string designates the path of the currently running executable, we could internally act accordingly. I am inclined to say that if a user insists and directly supplies the path to the executable, the install method should fail, assuming that it can only handle paths to shared libraries (and maybe zip files in the future).

Another option would be to automatically always install bundles from the executable during framework start. I don't see a use case for bundles linked into executables where the operator will not want them to be installed / started right away (if he would, he could have used a regular shared lib in the first place).

What do you think?

**Comment by saschazelzer** [comment]
I have thought some more about this and really think we should only allow users to directly install shared libraries (and maybe zip files as "fragments" later). In particular, all logical bundles embedded in the executable hosting a framework instance should be installed automatically on framework init. This means that in the general static linking scenario, explicit install commands are not required anymore.

**Comment by saschazelzer** [comment]
I have thought some more about this and really think we should only allow users to directly install shared libraries (and maybe zip files as "fragments" later). In particular, all logical bundles embedded in the executable hosting a framework instance should be installed automatically on framework init. This means that in the general static linking scenario, explicit install commands would not be required anymore.

**Comment by karthikreddy09** [comment]
Yes, It makes sense to only allow installing shared libs and automatically install all the logical bundles embedded in the executable. I think the auto-install should happen no matter how the framework is built (shared or static). E.g, if the user code is all statically linked but the framework is built as a shared lib and linked to the executable.

I think it is reasonable to also auto-start the embedded bundles. The order of install/start should not matter.

However, this new scheme results in conditional compile statements which is not ideal. something like the following ....

```
#if Sharedlibs
  Vector<Bundle> bundles = bc.install(location);
  for (auto bundle : bundles)
  {
    bundle.start();
  }
#endif // no need to install / start bundles for embedded logical bundles
```

I think an attempt to uninstall the embedded logical bundles should throw an exception.

saschazelzer:
Yes, that is also how I meant it: "auto-install" all embedded bundles of the executable, no matter of the build mode. For these bundles, I agree that an attempt to uninstall should throw.

I am less sure about the "auto-start" though. Bundle activation is potentially time consuming and could impact application start-up. And there might still be scenarios where you want to conditionally enable bundles ("features") in your statically build application.

karthikreddy09:
Hmm ... How do users handle the start ? 
They need to have some knowledge about the bundle to determine which bundles need to start among the list of installed bundles. In the case of shared libs, they could blindly iterate over a directory, install and start each of the plugins in the directory.

They could use LDAP filter to find the plugins they are interested in but is this OK ?

saschazelzer:
To me it is not really different than before. The application or operator is either going to start all bundles anyways (but then it is questionable if the usage of bundles makes sense at all) or already has a set of (initial) rules to provision the application. Then iterating over all bundles via `BundleContext::GetBundles()` and start the relevant entries makes sense to me. Actually, it is very application specific and should not depend on how the bundles where installed.

For example in the usShell application, the user is presented with a list of installed bundles and starts them by inspecting the list and using the bundle id to start specific bundles. A UI would have a point and click solution, declarative services would use proxy services and start the corresponding bundle when a particular service is used. All of these scenarios could (in the future) rely on the req/cap model to start dependencies automatically.

jeffdiclemente:
I agree with Sascha about "auto-start", I don't believe its needed at this time.
We don't know whether users will actually need "auto-start". We have no requirement for it.

karthikreddy09:
After some more thought, I am convinced that auto-start is not required for embedded bundles. 

There was some code in the [EmbeddedBundles ](https://github.com/CppMicroServices/CppMicroServices/tree/EmbeddedBundles) branch related to auto-install. I may be able to salvage some of that work.

saschazelzer:
I think it could be as simple as adding `bundleRegistry.Install(<exec-path>, systemBundle.get())` before the `bundleRegistry.Load()` line in usCoreBundleContext.cpp. In the `Load()` call, we could set a new boolean flag in the bundleRegistry which it uses later on to see if it should throw execptions in case clients want to install something "unusual", e.g. the executable again.

karthikreddy09:
@ sascha, sorry for the delayed response. 

Yes, this should work ... I don't think we need a separate flag though. We could key off of the system bundle's state. If the state is STATE_STARTING, we allow installing the executable else throw an exception. By the time the user calls the install API, system bundle's state would be STATE_ACTIVE

What do you think ?

saschazelzer:
Problem is, the framework is in the `STATE_STARTING` state after `Framework::Init()` method was called. A user could then get the framework's context and install bundles.

The flag could just be an additional boolean parameter to the `BundleRegisty::Install()` method to control the checking of the contraints.

karthikreddy09:
Good catch !! I have made the changes as per your suggestions and have auto-install working on the Mac platform. Will post my changes once I get a chance to test other configurations.

karthikreddy09:
Closing this pull request .... Refer to PR#109 for latest implementation.

PR #92, opened 2016-04-14
75 remove autoinstall

karthikreddy09:
Removed the CMAKE option for auto install and all the related APIs. Removed references to auto-install in documentation.


cmake/usCTestScript_custom.cmake
```diff
@@ -23,7 +23,7 @@ set(US_TEST_STATIC 1)
 set(US_SOURCE_DIR "${CMAKE_CURRENT_LIST_DIR}/../")
 
 set(US_BUILD_CONFIGURATION )
-foreach(i RANGE 7)
+foreach(i RANGE 3)
```
jeffdiclemente: is it possible to not hard code the max range? Could it be derived somehow? I could see how adding/removing build configurations frequently could lead to this value not being updated.

saschazelzer: For me, it would still be okay like this, if there is no easy solution to it. I hope that the number of configurations does not increase any more.

saschazelzer:
Looks good to me, ready for merge.

PR #96, opened 2016-05-11
Framework Listener and logging support

### LINKED ISSUE #40: Add FrameworkListener and improve Framework logging
The logging facility uses a singleton for managing log levels. Initialization is not thread-safe on compilers not supporting magic statics and multiple framework instances cannot be handled.

jeffdiclemente:
Implements FrameworkEvent, Framework Listeners and an internal Framework logger.

Closes #40 


jeffdiclemente:
@CppMicroServices/developers 
Please review and provide feedback. 
Built and tested on linux, mac and windows.

**Note:** this code is still incomplete from a behavioral standpoint and the framework listener tests will fail. I'm waiting for @saschazelzer 's changes to the bundle life cycle (i.e. the introduction of ```Framework::init()```) to complete this work.

jeffdiclemente:
I've noticed many indentation issues. These will be fixed before merging.

core/include/usFrameworkEvent.h
```diff
@@ -0,0 +1,217 @@
...
+  enum class Type : unsigned int {
+
+	/**
+	 * The Framework is about to be started.
+	 * <p>
+	 * The Framework's
+	 * \link BundleActivator::Start(BundleContext*) BundleActivator Start\endlink method
+	 * is about to be called.
+	 */
+    STARTING,
```
saschazelzer: I might be confused, but I cannot find the STARTING type in the OSGi Core 6.0 specs.

core/include/usFrameworkEvent.h
```diff
@@ -0,0 +1,217 @@
...
+    STOPPED = 64,
+
+    /**
+     * The Framework is about to be stopped.
+     * <p>
+     * The Framework's
+     * \link BundleActivator::Stop(BundleContext*) BundleActivator Stop\endlink method
+     * is about to be called.
+     */
+    STOPPING,
```
saschazelzer: Same as for STARTING.

core/include/usFrameworkEvent.h
```diff
@@ -0,0 +1,217 @@
...
+
+  /**
+   * Creates a Framework event of the specified type.
+   *
+   * @param type The event type.
+   * @param bundle The bundle associated with the event. This bundle is also the source of the event.
+   * @param message The message associated with the event.
+   * @param exception The exception associated with this event. Should be nullptr if there is no exception.
+   */
+  FrameworkEvent(Type type, const std::shared_ptr<Bundle>& bundle, const std::string& message, const std::exception_ptr exception = nullptr);
```
saschazelzer: The constructor signature is also slightly different then in the specs. Are there cases when you want to transport an error message without a corresponding exception?

core/include/usBundleContext.h
```diff
@@ -691,7 +736,18 @@ class US_Core_EXPORT BundleContext
...
+  template<class S, class T> friend class ServiceTracker;
+  template<class S, class TTT> friend class ServiceTrackerPrivate;
+  template<class S, class TTT> friend class TrackedService;
+
   BundleContext(BundlePrivate* bundle);
+
+  // Not for use by clients of the Framework.
+  // Provides access to the Framework's log sink to allow templated code
+  // to log diagnostic information. 
+  std::shared_ptr<LogSink> GetLogSink();
```
saschazelzer: I believe we can move this to the private implementation of BundleContext when we merge it with the lifecycle branch.

core/src/bundle/usBundleContext.cpp
```diff
@@ -169,9 +176,12 @@ struct ServiceHolder
...
       bc->UngetService(sref);
     }
-    catch (const std::exception& exp)
+    catch (const std::exception& )
     {
-      US_INFO << "UngetService threw an exception - " << exp.what();
+	  // don't throw exceptions from the destructor. For an explanation, see:
+	  // https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md
+	  // Following this rule means that a FrameworkEvent isn't an option here 
+	  // since it contains an exception object which clients could throw.
```
saschazelzer: This wasn't throwing before, was it? What about keeping a log message here, similar to the previous US_INFO call?

saschazelzer:
I like the changes and will do a more thorough review after the merge with the lifecycle branch.

After the merge, we need to properly send the framework events via the bundle thread mechanism, to avoid potential locking issues.

core/include/usFrameworkEvent.h
jeffdiclemente: You're right, it doesn't exist in the OSGi 6 spec. this was a carry over from using the BundleEvent implementation for FrameworkEvent.

If there are no objections; I'll remove ```STARTING```.

core/include/usFrameworkEvent.h
jeffdiclemente: You're right, it doesn't exist in the OSGi 6 spec. this was a carry over from using the BundleEvent implementation for FrameworkEvent.

If there are no objections; I'll remove ```STOPPING```.

core/include/usFrameworkEvent.h
jeffdiclemente: I felt that given that the OSGi spec allows an exception (i.e. ```Throwable```) object to be ```null``` and that there exists the ```INFO``` event type, which is described as informational, an exception is not always appropriate and another way to convey a message is necessary. My solution was to add the ```message``` parameter for cases where ```std::exception_ptr``` is ```nullptr```.

This is of course up for debate.

I didn't think it made sense, from a C++ developer standpoint, to provide a ```std::exception_ptr``` solely to provide an informational message.

core/src/bundle/usBundleContext.cpp
jeffdiclemente: You're right, this wasn't throwing before.

As long as the logger remains accessible through the ```BundleContext``` instance, I can log the exception similar to before. I will do that once the final lifecycle changes are merged into this branch.

jeffdiclemente:
@CppMicroServices/collaborators please review.

core/doc/doxygen/MicroServices_Debugging.dox
```diff
@@ -1,70 +1,2 @@
 
-/**
```
saschazelzer: Since this file is empty now, we should delete it.

core/include/usBundleContext.h
saschazelzer: What about moving the `GetLogSink()` to BundleContexPrivate? I think we can then also get rid of the new friend declarations and use `GetPrivate()` instead.

core/include/usFrameworkEvent.h
```diff
@@ -41,13 +40,14 @@ class FrameworkEventData;
...
- * when a general event occurs within the CppMicroServices environment.
+ * <code>FrameworkEvent</code> objects are delivered to listeners connected
+ * via BundleContext::AddFrameworkListener() when an event occurs within the
+ * Framework which a user would be interested in. A <code>Type</code> code 
+ * is used to identify the event type for future extendability.
  *
- * @see FrameworkListener
+ * @see BundleContext#AddFrameworkListener
  */
 class US_Core_EXPORT FrameworkEvent
```
saschazelzer: I suggest to modify this class such that it works similar to `BundleEvent` and `Service Event`.  This would mean using a `std::shared_ptr` instead of the `SharedDataPointer` template (this will allow us to rely on the compiler generated destructor). Also remove the copy constructor and copy assignment operator.

core/include/usFrameworkEvent.h
```diff
@@ -56,7 +56,10 @@ class US_Core_EXPORT FrameworkEvent
 
 public:
 
-  enum Type {
+  /**
+   * A type code used to identify the event type for future extendability.
+   */
+  enum class Type : unsigned int {
```
saschazelzer: The other event classes use an unscoped enum with a `uint32_t` type. We should be consistent.

core/include/usFrameworkEvent.h
```diff
@@ -95,31 +99,28 @@ class US_Core_EXPORT FrameworkEvent
...
-     *
+     * The Framework has been stopped.
      * <p>
-     * This event is fired when the Framework has been stopped because of a stop
-     * operation on the system bundle. The source of this event is the System
-     * Bundle.
+     * The Framework's
+     * \link BundleActivator::Stop(BundleContext*) BundleActivator Stop\endlink method
+     * has been executed.
      */
```
saschazelzer: This looks like a merge error. The original documentation is according to the specs and should be retained.

core/include/usFrameworkEvent.h
```diff
@@ -95,31 +99,28 @@ class US_Core_EXPORT FrameworkEvent
...
-     *
+     * The Framework is about to be stopped.
      * <p>
-     * This event is fired when the Framework has been stopped because of an
-     * update operation on the system bundle. The Framework will be restarted
-     * after this event is fired. The source of this event is the System Bundle.
+     * The Framework's
+     * \link BundleActivator::Stop(BundleContext*) BundleActivator Stop\endlink method
+     * is about to be called.
      */
```
saschazelzer: Again, documentation merge error.

core/include/usFrameworkEvent.h
```diff
@@ -95,31 +99,28 @@ class US_Core_EXPORT FrameworkEvent
...
-     * The Framework did not stop before the wait timeout expired.
-     *
-     * <p>
-     * This event is fired when the Framework did not stop before the wait
-     * timeout expired. The source of this event is the System Bundle.
+     * The Framework did not stop before the wait timeout expired. 
+	 * <p>
+	 * This event is fired when the Framework did not stop before the wait timeout expired. 
+	 * The source of this event is the System Bundle.
      */
```
saschazelzer: Indentation.

core/include/usFrameworkEvent.h
```diff
@@ -133,47 +134,50 @@ class US_Core_EXPORT FrameworkEvent
   ~FrameworkEvent();
 
   /**
-   * Can be used to check if this FrameworkEvent instance is valid,
-   * or if it has been constructed using the default constructor.
+   * Returns <code>false</code> if the FrameworkEvent is empty (i.e invalid) and
+   * <code>true</code> if the FrameworkEvent is not null and contains valid data.
+   */
+  bool operator!() const;
```
saschazelzer: The other event classes have an explicit boolean conversion operator. Can we have the same here instead?

core/include/usFrameworkEvent.h
```diff
@@ -133,47 +134,50 @@ class US_Core_EXPORT FrameworkEvent
...
-   *         <code>false</code> otherwise.
+   * @param type The event type.
+   * @param bundle The bundle associated with the event. This bundle is also the source of the event.
+   * @param message The message associated with the event.
+   * @param exception The exception associated with this event. Should be nullptr if there is no exception.
    */
-  bool IsNull() const;
+  FrameworkEvent(Type type, const Bundle& bundle, const std::string& message, const std::exception_ptr exception = nullptr);
 
   FrameworkEvent(const FrameworkEvent& other);
```
saschazelzer: Should be removed to enable default move.

core/include/usFrameworkEvent.h
```diff
@@ -133,47 +134,50 @@ class US_Core_EXPORT FrameworkEvent
...
 
   FrameworkEvent& operator=(const FrameworkEvent& other);
```
saschazelzer: Should be removed to enable default move.

core/include/usFrameworkEvent.h
```diff
@@ -133,47 +134,50 @@ class US_Core_EXPORT FrameworkEvent
...
+   * @return The exception. May be <code>nullptr</code> if there is no related exception.
    */
-  Bundle GetBundle() const;
+  std::exception_ptr GetThrowable() const;
 
   /**
-   * Returns the type of framework event.
-   * <p>
-   * The type values are:
+   * Returns the type of lifecyle event. The type values are:
```
saschazelzer: Should be `framework` not `lifecyle` event.

core/include/usFrameworkEvent.h
```diff
@@ -195,13 +199,13 @@ class US_Core_EXPORT FrameworkEvent
  * @{
  */
 US_Core_EXPORT std::ostream& operator<<(std::ostream& os, FrameworkEvent::Type eventType);
-US_Core_EXPORT std::ostream& operator<<(std::ostream& os, const FrameworkEvent& event);
+US_Core_EXPORT std::ostream& operator<<(std::ostream& os, const std::exception_ptr ex);
```
saschazelzer: I don't think this overload is a good idea, we should probably use a helper function for retrieving the exception message instead. Problem is, that this overload is not tied to CppMicroServices specific argument types and hence will be selected whenever someone tries to stream an `exception_ptr` and the FrameworkEvent.h header has been included.

core/include/usFrameworkFactory.h
```diff
@@ -48,29 +49,20 @@ class US_Core_EXPORT FrameworkFactory
     FrameworkFactory(void);
     virtual ~FrameworkFactory(void);
 
+    FrameworkFactory(const FrameworkFactory&) = delete;
+    FrameworkFactory& operator=(const FrameworkFactory&) = delete;
+
```
saschazelzer: I think we should make this class much simpler and remove all the constructors and assignment operators. This class has no state and it would be perfectly okay to copy it around.

core/src/CMakeLists.txt
```diff
@@ -65,6 +64,7 @@ set(_srcs
 set(_private_headers
   util/usFrameworkPrivate.h
   util/usLDAPExpr_p.h
+  util/usLog.h
```
saschazelzer: I suggest to rename that file to `usLog_p.h` to be consistent with the current state.

core/src/bundle/usBundleAbstractTracked_p.h
```diff
@@ -281,6 +278,8 @@ class BundleAbstractTracked : public MultiThreaded<MutexLockingStrategy<>,WaitCo
    */
   std::atomic<int> trackingCount;
 
+  BundleContext* bc;
```
saschazelzer: Could be `BundleContext* const` instead. If the `GetLogSink()` method is moved to the bunde private classe, this would become `BundleContextPrivate * const`.

core/src/bundle/usBundleContext.cpp
```diff
@@ -71,6 +73,10 @@ BundleContext& BundleContext::operator=(std::nullptr_t)
   return *this;
 }
 
+std::shared_ptr<LogSink> BundleContext::GetLogSink()
```
saschazelzer: Member function should be `const`.

core/src/bundle/usBundleResource.cpp
```diff
@@ -281,20 +281,14 @@ void* BundleResource::GetData() const
...
   if (!IsValid()) return nullptr;
 
-  void* data = d->archive->GetResourceContainer()->GetData(d->stat.index);
-  if (data == nullptr)
-  {
-    US_WARN << "Error uncompressing resource data for " << this->GetResourcePath() << " from "
-            << d->archive->GetBundleLocation();
-  }
-  return data;
+  return d->archive->GetResourceContainer()->GetData(d->stat.index);
```
saschazelzer: Should we still issue a diagnostic log message here, or just drop it?

core/src/bundle/usBundleResource.cpp
```diff
@@ -281,20 +281,14 @@ void* BundleResource::GetData() const
...
 }
 
 }
 
-using namespace us;
-
-std::ostream& operator<<(std::ostream& os, const BundleResource& resource)
+std::ostream& operator<<(std::ostream& os, const us::BundleResource& resource)
```
saschazelzer: Since we are looking at this, could you move the operator inside the `us` namespace?

core/src/bundle/usCoreBundleContext.cpp
```diff
@@ -59,19 +59,19 @@ std::map<std::string, Any> InitProperties(std::map<std::string, Any> configurati
   return configuration;
 }
 
-CoreBundleContext::CoreBundleContext(const std::map<std::string, Any>& props)
+CoreBundleContext::CoreBundleContext(const std::map<std::string, Any>& props, std::ostream* logger)
   : id(globalId++)
+  , frameworkProperties(InitProperties(props))
+  , sink(std::make_shared<LogSink>(logger, any_cast<bool>(frameworkProperties.at(Constants::FRAMEWORK_LOG))))
```
saschazelzer: I am not sure if a `shared_ptr` is worth it. We are not really sharing ownership of the log sink. Maybe a `unique_ptr` would be a better fit (and internally handing out `LogSink*`)?

Also the `any_cast<bool>` could throw an exception, preventing the framework from being constructed. Is this intended? If the property is not of type `bool`, I think we should fall back to using the default instead.

core/src/bundle/usCoreBundleContext.cpp
```diff
@@ -83,7 +83,6 @@ CoreBundleContext::~CoreBundleContext()
 
 void CoreBundleContext::Init()
 {
-  US_INFO("initializing");
```
saschazelzer: Can we keep this and the messages below as diagnostic message? I found them quite useful during debugging.

core/src/service/usServiceHooks.cpp
```diff
@@ -52,15 +54,15 @@ std::shared_ptr<ServiceListenerHook> ServiceHooks::AddingService(const ServiceRe
   {
     lh->Added(coreCtx->listeners.GetListenerInfoCollection());
   }
-  catch (const std::exception& e)
+  catch (const std::exception& )
```
saschazelzer: Because we capture the exception pointer now, we can get rid of the `catch (const std::exception&)` block and only use the catch all block.

core/src/service/usServiceHooks.cpp
```diff
@@ -124,15 +126,15 @@ void ServiceHooks::FilterServiceReferences(BundleContextPrivate* context, const
         {
           fh->Find(MakeBundleContext(context->shared_from_this()), service, filter, filtered);
         }
-        catch (const std::exception& e)
+        catch (const std::exception& )
```
saschazelzer: Same here, duplicated exception handling.

core/src/service/usServiceHooks.cpp
```diff
@@ -174,15 +176,15 @@ void ServiceHooks::FilterServiceEventReceivers(const ServiceEvent& evt,
         {
           elh->Event(evt, filtered);
         }
-        catch(const std::exception& e)
+        catch(const std::exception& )
```
saschazelzer: Duplicated exception handling.

core/src/service/usServiceHooks.cpp
```diff
@@ -215,15 +217,15 @@ void ServiceHooks::HandleServiceListenerReg(const ServiceListenerEntry& sle)
       {
         lh->Added(set);
       }
-      catch (const std::exception& e)
+      catch (const std::exception& )
```
saschazelzer: Duplicated exception handling.

core/src/service/usServiceHooks.cpp
```diff
@@ -263,15 +265,21 @@ void ServiceHooks::HandleServiceListenerUnreg(const std::vector<ServiceListenerE
       {
         lh->Removed(lis);
       }
-      catch (const std::exception& e)
+      catch (const std::exception& )
```
saschazelzer: Duplicated exception handling.

core/src/service/usServiceListeners_p.h
```diff
@@ -60,6 +60,10 @@ class ServiceListeners : private MultiThreaded<>
 
 private:
 
+  std::mutex frameworkListenerMapMutex;
```
saschazelzer: Using threading related types directly will conflict with the threading option in CMake turned off. I suggest to use the same approach as `bundleListenerMap` to handle that transparently.

core/src/service/usServiceObjects.cpp
```diff
@@ -114,9 +114,12 @@ struct UngetHelper
         }
       }
     }
-    catch (const std::exception& ex)
+    catch (const std::exception& )
     {
-      US_INFO << "UngetService threw an exception - " << ex.what();
+      // don't throw exceptions from the destructor. For an explanation, see:
```
saschazelzer: What about a diagnostic message here?

core/src/util/usFramework.cpp
```diff
@@ -72,7 +73,7 @@ void Framework::Start()
       d->operation = BundlePrivate::OP_ACTIVATING;
       break;
     case STATE_ACTIVE:
-      return;
+      break;
```
saschazelzer: Is that change intentional? This will lead to double activations...

core/src/util/usFrameworkPrivate.cpp
```diff
@@ -168,6 +169,8 @@ void FrameworkPrivate::Shutdown(bool restart)
   case Bundle::STATE_STOPPING:
     // Shutdown already inprogress
     break;
+  default:
```
saschazelzer: I would prefer to not use a `default` label and rely on the compiler errors if we don't handle a case explicitly.

core/include/usBundleContext.h
jeffdiclemente: Yes, I forgot to do that. Thanks for the reminder.

core/include/usFrameworkEvent.h
jeffdiclemente: Good point. I agree. I'll move the exception output code into the ```FrameworkEvent```'s ```operator<<```. It can always be moved into a helper function later if need be.

core/include/usFrameworkEvent.h
jeffdiclemente: Yes.

core/src/util/usLog.h
```diff
@@ -0,0 +1,101 @@
...
+  void Log(const std::string& msg)
+  {
+    if (!_enable) return;
+    std::lock_guard<std::mutex> lock(_sink_lock);
+    *_sink << msg;
+  }
+
+private:
+  bool _enable;
+  std::mutex _sink_lock;
```
saschazelzer: I suggest to inherit the `LogSink` class from `MultiThreaded<>` and not use a `std::mutex` directly. Then this code will work properly in the single-threaded CMake configuration as well.

core/src/util/usLog.h
```diff
@@ -0,0 +1,101 @@
...
+  std::ostream* _sink;
```
saschazelzer: This could be const `std::ostream* const`

core/src/util/usUtils.cpp
```diff
@@ -388,46 +388,43 @@ std::string GetLastErrorStr()
...
-
-MsgHandler installMsgHandler(MsgHandler h)
-{
-  MsgHandler old = handler;
-  handler = h;
-  return old;
-}
-
-void message_output(MsgType msgType, const char *buf)
+void TerminateForDebug(const std::exception_ptr ex)
```
saschazelzer: This function is now never used. Would it make sense to call it in case of a framework error and a proprietary framework property being set? Should we call it in cases where we only issue a DIAG_LOG call and the error condition is "critical enough" ?

httpservice/src/usHttpServletRequest.cpp
```diff
@@ -258,7 +262,7 @@ long long HttpServletRequest::GetDateHeader(const std::string& name) const
                                            &second) == 6)) {
 
     month = -1;
-    for (std::size_t i = 0; i < num_months; ++i)
+    for (int i = 0; i < num_months; ++i)
```
saschazelzer: Thanks for fixing that. However, now gcc complains in the statement below which assigns `i` to `month`. I suggest to keep `std::size_t` in the for loop and use
```
month = static_cast<int>(i);
```
before breaking the loop. Thanks!

core/test/usFrameworkTest.cpp
```diff
@@ -113,30 +134,43 @@ namespace
...
+        US_TEST_CONDITION(listener.CheckListenerEvents(pEvts), "Check framework bundle event listener");
 
-        US_TEST_CONDITION(f.GetState() == Bundle::STATE_ACTIVE, "Check framework is active")
+        US_TEST_CONDITION(f.GetState() == Bundle::STATE_ACTIVE, "Check framework is active");
 
         f.GetBundleContext().AddBundleListener(&listener, &TestBundleListener::BundleChanged);
 
+        // To test that the correct FrameworkEvent is returned, we must guarantee that the Framework
+        // is in a STARTING, ACTIVE, or STOPPING state.
+        std::thread waitForStopThread([&f]() {
```
saschazelzer: For gcc, a `#include <thread>` statement is now required due to the usage of `std::thread`.

Also, we should "ifdef" that such that it is only executed in the multi-thread configuration.

core/test/usFrameworkTest.cpp
```diff
@@ -159,30 +193,122 @@ namespace
...
         auto bundle = testing::InstallLib(f.GetBundleContext(), "TestBundleA");
-        US_TEST_CONDITION_REQUIRED(bundle, "Non-null bundle")
+        US_TEST_CONDITION_REQUIRED(bundle, "Non-null bundle");
+
         bundle.Start();
+        US_TEST_CONDITION(bundle.GetState() == Bundle::STATE_ACTIVE, "Check that TestBundleA is in an Active state");
+
+        // To test that the correct FrameworkEvent is returned, we must guarantee that the Framework
+        // is in a STARTING, ACTIVE, or STOPPING state.
+        waitForStopThread = std::thread([&f]() {
```
saschazelzer: Same here (ifdef multi-threaded).

core/test/usFrameworkTest.cpp
```diff
@@ -271,9 +397,13 @@ int usFrameworkTest(int /*argc*/, char* /*argv*/[])
 
     TestDefaultConfig();
     TestCustomConfig();
+	TestCustomLogSink();
     TestProperties();
     TestLifeCycle();
     TestEvents();
+    TestConcurrentFrameworkStart();
+    TestConcurrentFrameworkStop();
+    TestConcurrentFrameworkWaitForStop();
```
saschazelzer: I think we should enclose these three new tests into a "ifdef multi-threaded".

core/src/bundle/usBundleResource.cpp
jeffdiclemente: yep. will do.

core/src/bundle/usBundleResource.cpp
jeffdiclemente: I'm in favor of dropping it. I didn't think it was worth the time to plumb the logger into the design of ```BundleResource```.

core/src/bundle/usCoreBundleContext.cpp
jeffdiclemente: sure, I'll add them back.

core/src/bundle/usBundleResource.cpp
saschazelzer: If plumbing is the only concern, I think we could easily do
```
GetPrivate(GetBundleContext())->GetLogSink()
```
without any plumbing, right?

core/src/service/usServiceObjects.cpp
jeffdiclemente: good catch. I missed this one. I've added it back.

core/src/util/usFrameworkPrivate.cpp
```diff
@@ -25,15 +25,14 @@ limitations under the License.
...
 #include "usFramework.h"
-#include "usLog.h"
+#include "usFrameworkEvent.h"
 
 
 namespace us {
 
 FrameworkPrivate::FrameworkPrivate(CoreBundleContext* fwCtx)
-  : BundlePrivate(fwCtx)
+    : BundlePrivate(fwCtx), stopEvent{ true, FrameworkEvent::Type::FRAMEWORK_ERROR, std::string(), nullptr }
```
saschazelzer: gcc 4.6 does not support brace initialization. You could use
```
stopEvent(FrameworkEventInternal{...})
```
instead.

jeffdiclemente: Already fixed. I saw that in travis ci after I submitted the PR. Thanks.

core/src/bundle/usBundleResource.cpp
jeffdiclemente: Good call. I overlooked the use of ```GetBundleContext()```. That should work. Thanks!

core/src/util/usFramework.cpp
jeffdiclemente: If I remember correctly, yes it was intentional. What wasn't intentional was the side-effect of the double activation.

The Framework does need to send the ```FRAMEWORK_STARTED``` event even if ```Framework::Start()``` is called multiple times. At least this is how I interpret the OSGi spec.

The ```break;``` allowed execution to continue and eventually send the ```FRAMEWORK_STARTED``` event.

I agree that double activation shouldn't happen. I'll fix this.

httpservice/src/usHttpServletRequest.cpp
jeffdiclemente: will do.

core/src/util/usLog.h
jeffdiclemente: good catch. fixed.

core/src/util/usUtils.cpp
jeffdiclemente: What if we create a new macro (```DIAG_LOG_TERMINATE(message)```?) that acts like an assert instead of trying to fit ```TerminateForDebug()``` into an existing mechanism (i.e. framework events or diagnostic logger)? 
The macro would be a no-op (off) by default and turned on if CppMicroServices is built with a new cmake build option (e.g. ```US_ENABLE_FRAMEWORK_ASSERTS```). 

I'm suggesting this since there are no levels in the diagnostic logger and framework error events don't seem like an appropriate place to terminate the process, even if it is only for debug builds.

core/src/service/usServiceListeners_p.h
jeffdiclemente: fixed.

core/include/usFrameworkFactory.h
jeffdiclemente: Sounds fine to me.

core/src/CMakeLists.txt
jeffdiclemente: fixed.

core/include/usFrameworkEvent.h
jeffdiclemente: fixed.

core/include/usFrameworkEvent.h
jeffdiclemente: fixed

core/include/usFrameworkEvent.h
jeffdiclemente: fixed

core/include/usFrameworkEvent.h
jeffdiclemente: fixed

core/include/usFrameworkEvent.h
jeffdiclemente: fixed

core/include/usFrameworkEvent.h
jeffdiclemente: fixed

core/include/usFrameworkEvent.h
jeffdiclemente: fixed

core/include/usFrameworkEvent.h
jeffdiclemente: I agree. fixed

core/src/service/usServiceHooks.cpp
jeffdiclemente: fixed

core/src/service/usServiceHooks.cpp
jeffdiclemente: fixed

core/src/service/usServiceHooks.cpp
jeffdiclemente: fixed

core/src/service/usServiceHooks.cpp
jeffdiclemente: fixed

core/src/service/usServiceHooks.cpp
jeffdiclemente: fixed

core/src/util/usFrameworkPrivate.cpp
jeffdiclemente: The ```default``` case was meant as a run-time check to catch any conditions when the bundle state was not correct.

I can remove the ```default``` case if you believe it unnecessary. let me know.

core/src/util/usLog.h
jeffdiclemente: fixed

core/test/usFrameworkTest.cpp
jeffdiclemente: fixed

core/test/usFrameworkTest.cpp
jeffdiclemente: fixed

core/test/usFrameworkTest.cpp
jeffdiclemente: fixed

core/src/util/usFramework.cpp
saschazelzer: I am not sure about the framework event. Looking at Knopflerfish and Felix, the former only sends the event once, but Felix sends it every time `start()` is called. Reading section 4.2.5 of the Core 6.0 specs:

> After the system bundle enters the ACTIVE state, a Framework STARTED event is broadcast.

it appears to me that the event should only be send once, *after* it enters the ACTIVE state. Calling `start()` again means we are already at that state. This is my interpretation. What passage were you refering to?

core/src/util/usFrameworkPrivate.cpp
saschazelzer: My point is, that we are already checking this at compile time. If an enumeration value would not be handled, the compiler throws an error.

core/src/util/usUtils.cpp
saschazelzer: I agree with you, there is no sensible place for terminating the process. I somewhat overlooked the "termination" part and was mostly thinking about the ability to trigger a debug break, if a particular framework property is set. This could be nice to have in certain scenarios. For assertions, I believe we should just use the existing ` assert` macro.

What about moving this discussion into a new issue and move forward with the PR?

core/src/util/usFramework.cpp
jeffdiclemente: Ok. I was looking at Felix as a reference for this.
Section 4.2.5 does help but I still think that the statement
> After the system bundle enters the ACTIVE state, a Framework STARTED event is broadcast.

is ambiguous. It doesn't address concurrency. Section 10.2.2.11 doesn't offer much help either. 

I'll revert back to the ```return```. There is no clear need or use case to send Framework STARTED events multiple times.

I wasn't referring to any passage in particular. I was referring to the absence of any statement to the effect of "only send the Framework STARTED event once" in the OSGi spec.

core/src/util/usFrameworkPrivate.cpp
jeffdiclemente: yep, I understand. I'll remove the ```default``` case.

core/src/util/usUtils.cpp
jeffdiclemente: > What about moving this discussion into a new issue and move forward with the PR?

That sounds good. I'll create a new issue and we can discuss what to do.

core/include/usBundleContext.h
jeffdiclemente: I started working on this and am realizing that this may be more trouble than its worth. 

I can't see a way to push ```GetLogSink()``` into ```BundleContextPrivate``` without exposing some other private header/function. Even if I was to export ```GetPrivate()``` from the library (like what is done with ```MakeBundleContext()``` in ```usGetBundleContext.h```) clients can then access the ```BundleContextPrivate``` object by doing ```GetPrivate(MakeBundleContext(...))```.

I'm not sure this is worth the effort given our timeline for 3.0 right now. 
Is there a better solution which doesn't expose private implementation classes?

core/src/bundle/usCoreBundleContext.cpp
jeffdiclemente: I share your skepticism about the use of ```shared_ptr```. Can we defer exploring changing it to ```unique_ptr``` and capture that work in a issue?

Yes, it is intentional to throw if the framework property is not a ```bool```. I believe that we should fail early and force the user to do the right thing (i.e. use a ```bool```) instead of trying to account for possible user error (e.g. a user using a ```string``` instead of ```bool```). Since this is a Framework defined property I think we should enforce its correct use.
If the user does get an exception, the exception message plus adequate documentation about the framework properties should be enough for users to correct their mistake. Silently failing and using a default value would cause user confusion IMHO.

jeffdiclemente:
@saschazelzer Thanks for all the great feedback. There are only a couple comments left which require your attention and it looks like they were hidden with my latest submit to this PR.

core/include/usBundleContext.h
saschazelzer: You are right, the template code makes this really hard if not 
impossible. I agree that the private method in BundleContext is a good 
compromise.

core/src/bundle/usCoreBundleContext.cpp
saschazelzer: I agree, let's create a separate issue for the shared_ptr log sink.

Your reasoning about the exception handling makes sense, let's stick 
with your solution.

core/test/usBundleEventTest.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+  // copy test
+  BundleEvent dup_valid_event(valid_event);
+  US_TEST_CONDITION_REQUIRED(dup_valid_event == valid_event, "Test BundleEvent copy ctor");
+
+  // copy assignment test
+  dup_valid_event = invalid_event;
+  US_TEST_CONDITION_REQUIRED((dup_valid_event == invalid_event), "Test BundleEvent copy assignment");
+
+  dup_valid_event = std::move(BundleEvent(BundleEvent::Type::BUNDLE_INSTALLED, f));
```
saschazelzer: Clang complains about the move:
> Moving a temporary object prevents copy elision
just removing the `std::move` will do.

core/test/usFrameworkEventTest.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  // copy test
+  FrameworkEvent dup_error_event(error_event);
+  US_TEST_CONDITION_REQUIRED(dup_error_event == error_event, "Test copy ctor");
+
+  // copy assignment test
+  dup_error_event = invalid_event;
+  US_TEST_CONDITION_REQUIRED(!(dup_error_event == error_event), "Test copy assignment");
+
+  dup_error_event = std::move(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_STARTED, f, ""));
```
saschazelzer: Another move preventing copy elision.

saschazelzer:
This is looking really good. I found two more small issues about a `std::move` but other than that, I'd say it is good to merge.

jeffdiclemente:
excellent! 
I will fix those ```std::move``` issues, open an issue for the ```std::unique_ptr<LogSink>``` task, do a final build on Windows, Mac and Linux for my own sanity and then merge.

Issue #85, opened 2016-03-14
Consolidate data types for for bundle/service properties to use std::unordered_map


kevinleeMW:
This issue entails a follow-up task of "Support LDAP queries against bundle meta-data" (Issue #53):

Consolidating data types of bundle properties (std::map) and service properties (std::unordered_map that turns into a "Properties" object)

PR #105, opened 2016-07-05
25 bundle lifecycle

saschazelzer:
This PR introduces bundle lifecycle support.

Outstanding todos:
 - Work on single-threaded build configuration
 - Fix possible helgrind concurrency issues and enable helgrind tests


core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
  *
- * Represents a CppMicroServices bundle.
+ * An installed bundle in the Framework.
  *
- * <p>
- * A <code>%Bundle</code> object is the access point to a CppMicroServices bundle.
- * Each CppMicroServices bundle has an associated <code>%Bundle</code> object.
+ * A <code>%Bundle</code> object is the access point to define the lifecycle of an
+ * installed bundle. Each bundle installed in the CppMicroServices environment has
+ * an an associated <code>%Bundle</code> object.
```
jeffdiclemente: typo; there's an extra "an"

core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
+     * attempt to resolve a bundle at any time.
+     *
+     * The value of {@code STATE_RESOLVED} is 0x00000004.
+     */
+    STATE_RESOLVED = 0x00000004,
+
+    /**
+     * The bundle is in the process of starting.
+     *
+     * A bundle is in the {@code STATE_STARTING} state when its {@link #Start(int)
```
jeffdiclemente: Shouldn't this be ```{@link #Start(uint32_t)```?

core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
+     * activation is triggered.
+     *
+     * The value of {@code STATE_STARTING} is 0x00000008.
+     */
+    STATE_STARTING = 0x00000008,
+
+    /**
+     * The bundle is in the process of stopping.
+     *
+     * A bundle is in the {@code STATE_STOPPING} state when its {@link #Stop(int)
```
jeffdiclemente: Shouldn't this be ```{@link #Stop(uint32_t)```?

core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
+     *
+     * A bundle is in the {@code STATE_ACTIVE} state when it has been successfully
+     * started and activated.
+     *
+     * The value of {@code STATE_ACTIVE} is 0x00000020.
+     */
+    STATE_ACTIVE = 0x00000020
+  };
+
+  enum StartOptions : uint32_t {
```
jeffdiclemente: I think it would be a good idea to mark these options as not supported. 
Clients will most likely think there is full support for these options if they are documented.

core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
+     * framework that the bundle must be activated using the bundle's declared
+     * activation policy.
+     *
+     * @see Constants#PLUGIN_ACTIVATIONPOLICY
+     * @see #Start(uint32_t)
+     */
+    START_ACTIVATION_POLICY = 0x00000002
+  };
+
+  enum StopOptions {
```
jeffdiclemente: I think it would be a good idea to mark these options as not supported. Clients will most likely think there is full support for these options if they are documented.

jeffdiclemente: ```enum StopOptions``` is missing the declaration of the underlying type (```uint32_t```)

core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
   /**
-   * The property key for looking up this bundle's id.
-   * The property value is of type \c long.
+   * Constructs an invalid %Bundle object
    *
-   * @see \ref MicroServices_BundleProperties
+   * @see IsValid()
    */
-  static const std::string PROP_ID;
+  Bundle();
```
karthikreddy09: A note on how the bundle object becomes valid would be helpful.

core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
   /**
    * Returns this bundle's {@link BundleContext}. The returned
    * <code>BundleContext</code> can be used by the caller to act on behalf
    * of this bundle.
    *
-   * <p>
-   * If this bundle is not in the <code>STARTED</code> state, then this
+   * If this bundle is not in the <code>STATE_STARTED</code> state, then this
    * bundle has no valid <code>BundleContext</code>. This method will
    * return <code>nullptr</code> if this bundle has no valid
```
karthikreddy09: This comment is not valid anymore. The returned object is always a BundleContext (valid / invalid).

core/include/usBundle.h
```diff
@@ -192,36 +347,33 @@ class US_Core_EXPORT Bundle : public std::enable_shared_from_this<Bundle>
...
-   * Returns the version of this bundle as specified by the
-   * US_INITIALIZE_BUNDLE CMake macro. If this bundle does not have a
+   * Returns the version of this bundle as specified in its
+   * <code>manifest.json</code> file. If this bundle does not have a
    * specified version then {@link BundleVersion::EmptyVersion} is returned.
    *
-   * <p>
    * This method continues to return this bundle's version while
-   * this bundle is in the <code>STOPPED</code> state.
+   * this bundle is in the <code>STATE_STOPPED</code> state.
```
karthikreddy09: What about when the bundle is in STATE_UNINSTALLED ?? 

Here's the OSGi javadoc comment for this method ....

"Returns the version of this bundle as specified by its Bundle-Version manifest header. If this bundle does not have a specified version then Version.emptyVersion is returned.
This method must continue to return this bundle's version while this bundle is in the UNINSTALLED state."

core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
-   * <p>
-   * If this bundle is not in the <code>STARTED</code> state, then this
+   * If this bundle is not in the <code>STATE_STARTED</code> state, then this
    * bundle has no valid <code>BundleContext</code>. This method will
    * return <code>nullptr</code> if this bundle has no valid
    * <code>BundleContext</code>.
    *
    * @return A <code>BundleContext</code> for this bundle or
-   *         <code>0</code> if this bundle has no valid
+   *         <code>nullptr</code> if this bundle has no valid
```
karthikreddy09: This comment is not valid anymore. The returned object is always a BundleContext (valid / invalid).

core/include/usBundle.h
```diff
@@ -296,7 +448,7 @@ class US_Core_EXPORT Bundle : public std::enable_shared_from_this<Bundle>
    *
    * @param path The path name of the resource.
    * @return A BundleResource object for the given \c path. If the \c path cannot
-   * be found in this bundle or the bundle's state is \c STOPPED, an invalid
+   * be found in this bundle or the bundle's state is \c STATE_STOPPED, an invalid
```
karthikreddy09: Shouldn't this also work when the bundle is in UNINSTALLED state ?

core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
-   * @see \ref MicroServices_BundleProperties
+   * Invalid \c Bundle objects are created by the default constructor or
+   * can be returned by certain framework methods if the bundle has been
+   * uninstalled.
+   *
+   * @return \c true if this %Bundle object is valid and can safely be used,
+   *         \c false otherwise.
    */
-  static const std::string PROP_VENDOR;
+  bool IsValid() const;
```
jeffdiclemente: I recall you mentioned this would be removed.

core/include/usListenerFunctors.h
```diff
@@ -34,6 +34,25 @@ namespace us {
   class BundleEvent;
 
   typedef std::function<void(const ServiceEvent&)> ServiceListener;
+
+  /**
+   * \ingroup MicroServices
+   *
+   * A \c BundleEvent listener. When a \c BundleEvent is fired, it is
```
karthikreddy09: Now that BundleListener is asynchronous, we need implementation for SynchronousBundleListener to support extender pattern.

jeffdiclemente: The SynchronousBundleListener is necessary to implement Declarative Services.
@karthikreddy09 can you log an issue so we can track this work? Thanks.

core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
+     *
+     * @see #Stop(uint32_t)
+     */
+    STOP_TRANSIENT = 0x00000001
+  };
+
+  // Bundle(const Bundle &) = default;
+  // Bundle(Bundle&&) = default;
+  // Bundle& operator=(const Bundle&) = default;
+  // Bundle& operator=(Bundle&&) = default;
```
jeffdiclemente: Is the intent that the compiler will implicitly generate the necessary copy/move constructors and assignment operators?

Which of these operations are required?

IMHO, we should support correct copy and move operations for Bundle objects.

core/include/usBundle.h
```diff
@@ -47,143 +48,297 @@ typedef ServiceReference<void> ServiceReferenceU;
...
+class US_Core_EXPORT Bundle
 {
 
 public:
 
-  Bundle(const Bundle &) = delete;
-  Bundle& operator=(const Bundle&) = delete;
+  typedef Clock::time_point TimeStamp;
+
+  enum State : uint32_t {
```
jeffdiclemente: I think I understand why the ```State``` enums start with ```STATE_```; to avoid conflicts with other same named enums (e.g. ```BundleEvent```). Is this assumption correct?

These name conflicts could be resolved by using scoped enumerations for all conflicting enums. Doing so will require that ```static_cast<>``` be used if anyone wants to treat the enum as its underlying data type (e.g. doing bit-wise operations). This is not an unknown problem though and there is [a decent explanation](http://blog.bitwigglers.org/using-enum-classes-as-type-safe-bitmasks/) of how to solve this. Implementing a solution this this could be done later. 
IMHO this is a small price to pay for code clarity and type safety. Also it would conform to the OSGi spec.

core/include/usBundleContext.h
```diff
@@ -91,10 +92,61 @@ class US_Core_EXPORT BundleContext
 
 public:
 
-  BundleContext(const BundleContext&) = delete;
-  BundleContext& operator=(const BundleContext&) = delete;
+  BundleContext();
```
jeffdiclemente: Since the default ctor constructs an invalid ```BundleContext``` a comment similar to the one for ```Bundle``` would be helpful to users.

core/include/usBundleContext.h
```diff
@@ -91,10 +92,61 @@ class US_Core_EXPORT BundleContext
...
 
-  ~BundleContext();
```
jeffdiclemente: I'd simply like to confirm that removing the user-defined copy ctor/assignment operator and destructor was intentional to enable implicitly generated copy/move ctor and assignment operators.
Was this the intent?

core/include/usBundleContext.h
```diff
@@ -91,10 +92,61 @@ class US_Core_EXPORT BundleContext
...
+  /**
+   * Compares this \c BundleContext with the specified bundle
+   * context for order.
+   *
+   * How valid %BundleContext objects are ordered is an implementation
+   * detail and must not be relied on. Invalid \c BundleContext objects
+   * will always compare greater then valid \c BundleContext objects.
+   *
+   * @param rhs The \c BundleContext object to compare this object with.
+   * @return \c
```
jeffdiclemente: missing return value information.

core/include/usBundleContext.h
```diff
@@ -114,15 +166,15 @@ class US_Core_EXPORT BundleContext
...
+  Bundle GetBundle(long id) const;
 
   /**
-   * Get the bundle with the specified bundle name.
+   * Get the bundles with the specified bundle location.
    *
-   * @param name The name of the bundle to get.
-   * @return The requested \c Bundle or \c nullptr.
+   * @param name The location of the bundles to get.
+   * @return The requested {\c Bundle}s or an empty list.
```
jeffdiclemente: Consider using "vector" instead of "list" since a std::vector is returned.

core/include/usBundleEvent.h
```diff
@@ -108,7 +108,32 @@ class US_Core_EXPORT BundleEvent
      * <p>
      * The bundle has been removed from the Framework.
      */
-    UNINSTALLED
+    UNINSTALLED,
```
jeffdiclemente: I think its worthwhile to assign these enums the same values as defined in the OSGi core spec.
I did this with ```FrameworkEvents```.

core/include/usFramework.h
```diff
@@ -28,74 +28,176 @@
...
+     * {@link #STATE_STARTING}, {@link #STATE_ACTIVE} or {@link #STATE_STOPPING} states.
+     *
+     * @throws std::runtime_error If this Framework could not be initialized.
+     */
+    void Init();
+
+    /**
+     * Wait until this Framework has completely stopped. The {@code Stop}
+     * method on a Framework performs an asynchronous stop of the Framework
+     * if it was build with threading support.
```
jeffdiclemente: typo: "build" should be "built"

core/include/usFramework.h
```diff
@@ -28,74 +28,176 @@
...
+     *
+     * <p>
+     * Calling this method is the same as calling {@link #Start()}. There are no
+     * start options for the Framework.
+     *
+     * @param options Ignored. There are no start options for the Framework.
+     * @throws std::runtime_error If this Framework could not be started.
+     * @see #Start()
+     */
+    void Start(uint32_t options);
```
jeffdiclemente: Simply for the sake of conciseness and maintenance, ```void Start(uint32_t options);``` could be modified to take a default value of ```0``` and ```void Start();``` could then be eliminated completely. This could also be applied to the base class ```Bundle```.

core/include/usFramework.h
```diff
@@ -28,74 +28,176 @@
...
+     * <p>
+     * Calling this method is the same as calling {@link #Stop()}. There are no
+     * stop options for the Framework.
+     *
+     * @param options Ignored. There are no stop options for the Framework.
+     * @throws std::runtime_error If stopping this Framework could not be
+     *         initiated.
+     * @see #Stop()
+     */
+    void Stop(uint32_t options);
```
jeffdiclemente: Simply for the sake of conciseness and maintenance, ```void Stop(uint32_t options);``` could be modified to take a default value of ```0``` and ```void Stop();``` could then be eliminated completely. This could also be applied to the base class ```Bundle```.

core/include/usFramework.h
```diff
@@ -28,74 +28,176 @@
...
+    /**
      * Stop this Framework.
      *
-     * The following steps are taken to stop this Framework:
-     * -# This Framework's state is set to STOPPING.
-     * -# All installed bundles must be stopped.
+     * <p>
+     * The method returns immediately to the caller after initiating the
+     * following steps to be taken on another thread. If the Framework was not
+     * build with threading enabled, the steps are executed in the main thread.
```
jeffdiclemente: typo: "build" should be "built"

core/include/usGetBundleContext.h
```diff
@@ -28,14 +28,15 @@
 #error Missing preprocessor define US_BUNDLE_NAME
 #endif
 
+#include <usBundleContext.h>
+#include <usBundleContextPrivate.h>
```
jeffdiclemente: This will affect issue #43 requiring that ```usBundleContextPrivate.h``` be a "public" header. I would like to avoid doing this. 
Is there anything that can be done? Perhaps moving the declaration of ```MakeBundleContext```?

core/include/usBundle.h
saschazelzer: Yes.

core/include/usBundle.h
saschazelzer: Yes.

core/include/usBundle.h
saschazelzer: Definitely.

core/include/usBundle.h
saschazelzer: Right.

core/include/usBundle.h
saschazelzer: I will add one.

jeffdiclemente:
I noticed a few lingering mentions to ```ServiceConstants::``` which should be converted to ```Constants::```, most recently in ```usServiceObjects.h```. There may be more, I didn't do a thorough search.

core/include/usBundle.h
saschazelzer: Good catch, there is no STATE_STOPPED anymore (this is a left-over from the old way of doing things).

core/include/usBundle.h
saschazelzer: Yes.

core/include/usBundle.h
saschazelzer: I removed it in the `BundleContext`, but not here yet.

core/include/usListenerFunctors.h
saschazelzer: The behavior when calling bundle listeners is as specified for synchronous listeners in the specs. We do not have async events yet (although they are delivered in a separate thread, they are still synchronized with the bundle operation).

core/src/bundle/usBundle.cpp
```diff
@@ -303,40 +313,41 @@ std::vector<ServiceReferenceU> Bundle::GetServicesInUse() const
...
-  return result;
+  return os;
 }
 
 std::ostream& operator<<(std::ostream& os, const Bundle& bundle)
 {
   os << "Bundle[" << "id=" << bundle.GetBundleId() <<
         ", loc=" << bundle.GetLocation() <<
-        ", name=" << bundle.GetName() << "]";
+        ", name=" << bundle.GetSymbolicName() << "]";
```
jeffdiclemente: Would it be useful to also include the state of the bundle?

core/src/bundle/usBundleArchive_p.h
```diff
@@ -0,0 +1,171 @@
...
+   * @param recurse
+   * @return
+   */
+  std::vector<BundleResource> FindResources(const std::string& path, const std::string& filePattern,
+                                            bool recurse) const;
+
+  /**
+   * Get last modified timestamp.
+   */
+  Clock::time_point GetLastModified() const;
```
jeffdiclemente: Nothing calls ```Clock::time_point GetLastModified() const;``` and ```void SetLastModified(const Clock::time_point& tp);```. What data will be associated with this last modified time?

core/src/bundle/usBundleContext.cpp
```diff
@@ -37,68 +37,111 @@
...
+}
+
+bool BundleContext::operator<(const BundleContext& rhs) const
+{
+  return *this ? (rhs ? (d < rhs.d) : true) : false;
+}
+
+BundleContext::operator bool() const
+{
+  return d != nullptr && d->IsValid();
```
jeffdiclemente: This could be simplified further to 

```
return d && d->IsValid();
```

core/include/usBundle.h
saschazelzer: Yes, that is the intent. I wanted to make this intent clear by explicitly stating that we want to have the default constructors/operators here, but VS2013 does not support defaulted move constructors/operators.

core/include/usBundle.h
saschazelzer: I know about the scoped enum bitmask "solutions" but the reason to make the identifiers less generic was to avoid potential name clashes with macros, e.g. introduced (implicitly) from `windows.h`. Scoped enums would not help in that case. I had this once with the `FrameworkEvent::ERROR` identifier, which had to be renamed to e.g. `FRAMEWORK_ERROR`. In that sense, the above would be consistent and easier to use.

core/include/usBundleContext.h
saschazelzer: Agreed.

core/include/usBundleContext.h
saschazelzer: Correct. Unfortunately, we can not default all of them in VS2013.

core/include/usBundleContext.h
saschazelzer: The intention was to use "list" as an english noun describing a collection of items and not to refer to `std::list`. I wanted to decouple the documentation text from the underlying container type.

core/include/usBundleEvent.h
saschazelzer: You are right, it will also be consistent with the rest of the code.

Now, these enum identifiers are not consistent in style with the one from `Bundle::State` and should get a prefix (e.g. `BUNDLE_`) if we are not going to use scoped enums everywhere.

core/include/usBundle.h
jeffdiclemente: If I'm reading the rules for[ implicitly generated move constructors](http://en.cppreference.com/w/cpp/language/move_constructor) correctly, the compiler will not generate a move constructor by default in this case because there is a user-defined destructor. I dumped the symbols from a CppMicroServices.dll built with VS 2015 and noticed there is no move constructor. If I re-build without a user defined destructor, I do see a compiler generated move constructor.

Since the intent is to have a default move constructor (which I agree is the right thing to do), either the destructor needs to be removed or the move/copy constructors need to be user defined (to support VS 2013).

core/include/usBundleContext.h
jeffdiclemente: After seeing more instances of the use of "list" in the comments, I agree with you. The return type is clear from function signature.

core/include/usBundleInitialization.h
```diff
@@ -44,25 +46,22 @@ class BundleContext;
...
-  {                                                                                                                            \
-    return US_CONCAT(_us_bundle_context_instance_, US_BUNDLE_NAME);                                                            \
-  }                                                                                                                            \
-                                                                                                                               \
-  extern "C" US_ABI_EXPORT void US_CONCAT(_us_set_bundle_context_instance_, US_BUNDLE_NAME) (us::BundleContext* ctx) \
-  {                                                                                                                                    \
-    US_CONCAT(_us_bundle_context_instance_, US_BUNDLE_NAME) = ctx;                                                                     \
+#define US_INITIALIZE_BUNDLE                                                                                         \
+  std::atomic<us::BundleContextPrivate*> US_CONCAT(_us_bundle_context_instance_, US_BUNDLE_NAME) {};                 \
```
karthikreddy09: What is the purpose of "{}" at the end of this statement ?

core/include/usBundle.h
jeffdiclemente: You're right. I had forgotten about the macro define in wingdi.h.
I had initially thought clients could deal with it on their own (e.g. by using ```#undef ERROR```) however that puts a burden on clients.

Your approach to make the identifiers less generic looks to be the best way to avoid this macro name collision problem.

core/include/usBundleResource.h
```diff
@@ -28,13 +28,14 @@
 
 #include <ostream>
 #include <vector>
+#include <memory>
 
 US_MSVC_PUSH_DISABLE_WARNING(4099) // type name first seen using 'struct' now seen using 'class'
```
karthikreddy09: This warning disabling push/pop can be removed by changing "friend class ::std::hash<BundleResource>" to "friend struct ::std::hash<BundleResource>".

std::hash is a struct .... http://en.cppreference.com/w/cpp/utility/hash

core/include/usChrono.h
```diff
@@ -20,12 +20,20 @@
 
 =============================================================================*/
 
-#include <usGlobalConfig.h>
 
-namespace us {
+#ifndef USCLOCK_H
```
karthikreddy09: This file move operation is odd. Is this accidental or Github showing it wrong ? This is a new file and should have no history.

core/include/usFramework.h
karthikreddy09: +1

core/include/usFrameworkEvent.h
```diff
@@ -0,0 +1,207 @@
...
+
+#ifndef USFRAMEWORKEVENT_H
+#define USFRAMEWORKEVENT_H
+
+#include "usCoreExport.h"
+#include "usSharedData.h"
+
+#include <memory>
+
+#ifdef _MSC_VER
```
karthikreddy09: Could use the macro US_MSVC_PUSH_DISABLE_WARNING

core/include/usBundle.h
saschazelzer: Thanks for double-checking, you are right. I didn't "see" the destructor. Since we nned to declare the constructor virtual and cannot default the move methods, I will just define all of them.

core/include/usBundleEvent.h
saschazelzer: Maybe rather a `EVENT_` prefix...

core/include/usGetBundleContext.h
saschazelzer: Strange, I remember that I tried to find a solution to avoid having to include that header but didn't succeed. Now I looked at it again and it is actually quite simple. Good that you mentioned it!

core/src/bundle/usBundle.cpp
saschazelzer: Good idea.

core/include/usBundleResource.h
saschazelzer: I do not remember the details anymore, I just know that there was a reason for it back then. Maybe it was me being stupid. I will follow your suggestion and assume it will compile without warnings on all Visual Studio versions for now.

core/include/usFramework.h
saschazelzer: When I wrote this, I was thinking for some time about the pros and cons of the two approaches:
 - Two different functions are IMHO cleaner in a sense that they differentiate clearer between the advanced case which takes options and the simple common case, which has a simple documentation block. I am also not a huge fan of default arguments in the API, but we have that in other places already.
 - One common function means, well, one function less and a common large documentation block.

If you think the current approach is rather confusing to users, then we change it.

core/include/usChrono.h
saschazelzer: It is a new file with little contents and most a large part of it is the standard license header. So the standard Git similarity index value used to detect moves/renames kicks in and detects something. There's not much we can do here I think (you can change the similarity index when using git diff locally).

core/src/bundle/usBundleArchive_p.h
saschazelzer: Good catch. This is an omission. I will add a small test to check for correct timestamps. The last modified time is updated when bundles are installed, updated (not applicable yet) or uninstalled.

core/include/usFrameworkEvent.h
saschazelzer: I will leave that for Jeff to change in his logging branch to minimize merge conflicts.

jeffdiclemente: sounds good. i'll address this in #96

core/include/usFramework.h
jeffdiclemente: I have no problem with leaving the API as is. 
Using default parameters was only suggested as a means to write less code.

core/include/usBundleResource.h
```diff
@@ -277,15 +274,21 @@ class US_Core_EXPORT BundleResource
...
+        );
+
+  BundleResource(
+        int index,
+        const std::shared_ptr<const BundleArchive>& archive
+        );
 
-  friend class Bundle;
-  friend class BundlePrivate;
+  friend class BundleArchive;
```
jeffdiclemente: Since ````US_MSVC_PUSH_DISABLE_WARNING(4099)``` was removed this needs to be ```friend struct BundleArchive;``` otherwise a compile error occurs with VS.

core/include/usServiceProperties.h
```diff
@@ -43,97 +41,6 @@ namespace us {
  */
 typedef std::unordered_map<std::string, Any> ServiceProperties;
```
kevinleeMW: Is there a reason to keep this "usServiceProperties" class? Unless there is a reason to keep this typedef, I think it makes sense to remove this class and just use std::unordered_map. I can do this when I work on issue #85 (consolidating data types for bundle/service properties to use std::unordered_map).

core/src/bundle/usBundleStorageMemory.cpp
```diff
@@ -0,0 +1,117 @@
...
+    {
+      res.emplace_back(v.second->GetBundleId());
+    }
+  }
+  return res;
+}
+
+void BundleStorageMemory::Close()
+{
+  std::map<long, std::shared_ptr<BundleArchive>> ar;
```
jeffdiclemente: I'm not sure I understand why a stack variable is used. ```archives.v``` will be a valid object in an unspecified state and calling ```clear()``` will work after ```std::move``` but it seems redundant.

Wouldn't the same thing be accomplished by:
```
auto l = archive.Lock(); US_UNUSED(l);
archives.v.clear();
```

What am i missing?

saschazelzer: Good catch, I remember that there was some weirdness with this code. I'll clean it up.

core/include/usServiceProperties.h
saschazelzer: In issue #85 using a `Properties` typedef is mentioned. So renaming the above to `Properties` and using it consistently everywhere is what I thought #85 is about, right?

jeffdiclemente:
@saschazelzer this looks good to merge into development. Thanks!

PR #109, opened 2016-07-26
88 auto install embedded bundles from executable

karthikreddy09:
Install all embedded bundles in the executable binary at framework startup.


core/src/bundle/usBundleUtils.cpp
```diff
@@ -31,151 +30,103 @@
...
-  US_DEBUG << "GetLibraryPath_impl():GetBundleFileName() " << GetLastErrorStr();
-  return "";
+  
+void* GetExecutableHandle()
+{
+  return dlopen(0, RTLD_LAZY);;
 }
 
-void* GetSymbol_impl(const std::string& bundleName, const std::string& libLocation, const char* symbol)
+std::string GetExecutablePath()
```
karthikreddy09: Using dlsym to find "main" symbol in the handle returned by dlopen(0, RTLD_LAZY) does not work.
see http://coliru.stacked-crooked.com/a/fd8dd59f2c6c8baa

core/src/bundle/usBundle.cpp
```diff
@@ -138,6 +138,11 @@ void Bundle::Stop(uint32_t options)
 
 void Bundle::Uninstall()
 {
+  if (!IsSharedLibrary(GetLocation()))
+  {
+    throw std::runtime_error("Bundles embedded in the executable are not uninstallable");
```
jeffdiclemente: The "not uninstallable" part sounds awkward to me. What do you think about something like

> Bundles embedded in an executable cannot be uninstalled.

?

core/src/bundle/usBundleUtils.cpp
```diff
@@ -31,151 +30,103 @@
...
-  HMODULE handle = nullptr;
-  if (!sharedLibMode || bundleName == "main")
+  std::string execPath;
+  uint32_t bufsize = MAXPATHLEN;
+  std::unique_ptr<char[]> buf(new char[bufsize]);
+#if _WIN32
+  if (GetModuleFileName(nullptr, buf.get(), bufsize) == 0 || GetLastError() == ERROR_INSUFFICIENT_BUFFER)
   {
-    handle = GetModuleHandle(nullptr);
+    US_DEBUG << "GetModuleFileName failed" << GetLastErrorStr();
```
jeffdiclemente: all forms of the logger macros ```US_```* are no longer available. Use ```DIAG_LOG```

core/src/bundle/usBundleUtils.cpp
saschazelzer: That is because the executable is not linked with `-rdynamic`, as mentioned in https://github.com/CppMicroServices/CppMicroServices/pull/90#issuecomment-210188430. See also http://coliru.stacked-crooked.com/a/e68dcb2883c24514. I am okay with the  current implementation though.

core/src/bundle/usBundleRegistry.cpp
```diff
@@ -59,10 +59,17 @@ void BundleRegistry::Clear()
...
+std::vector<Bundle> BundleRegistry::Install(const std::string& location,
+                                            BundlePrivate* caller,
+                                            bool allowExecutable)
 {
   CheckIllegalState();
-
+  
+  if (!allowExecutable && !IsSharedLibrary(location))
+  {
+    throw std::runtime_error("Installing bundle at "+ location + " not permitted");
```
saschazelzer: It would be nice to communicate the real reason why the install failed, e.g.
> Cannot install bundles from \<location\> because it is not a shared library

core/src/bundle/usCoreBundleContext.cpp
```diff
@@ -120,6 +121,9 @@ void CoreBundleContext::Init()
 
   serviceHooks.Open();
   //resolverHooks.Open();
+  
+  // auto-install all embedded bundles inside the executable
+  std::vector<Bundle> bundles = bundleRegistry.Install(BundleUtils::GetExecutablePath(), systemBundle.get(), true);
```
saschazelzer: The `bundles` variable is never used, it should be removed.

core/include/usGetBundleContext.h
```diff
@@ -29,10 +29,12 @@
...
 #include <usBundleContext.h>
-#include <usBundleUtils.h>
 
-#include <memory>
-#include <cstring>
+#define GETBUNDLECONTEXT_FCNPREFIX _us_get_bundle_context_instance
+#define GETBUNDLECONTEXTINSTANCEFCN(Prefix, BundleName) GETBUNDLECONTEXTINSTANCEFCN_(Prefix, BundleName)
+#define GETBUNDLECONTEXTINSTANCEFCN_(Prefix, BundleName) Prefix##_##BundleName
+
+extern "C" us::BundleContextPrivate* GETBUNDLECONTEXTINSTANCEFCN(GETBUNDLECONTEXT_FCNPREFIX, US_BUNDLE_NAME)();
```
saschazelzer: It is a good idea to clean that up and define some common macros. We can do some more cleanup and simplify the `GetBundleContext` method, see 89dbe3f0d10af9d80e8e8b2388a67e1274700292 for my suggestions. If you agree with these changes, you can cherry-pick that commit and apply it to the HEAD of your branch.

saschazelzer:
This looks good to me for merging, after the comments have been resolved.

core/include/usGetBundleContext.h
karthikreddy09: This is great ... I like the changes. Thanks !! :-)

core/src/bundle/usBundleUtils.cpp
karthikreddy09: Sorry, I should have read your comment carefully. Clearly, I had my blinders on.

saschazelzer:
Great work. Just to be clear: +1 for merging (the US_DEBUG conflicts could be resolved in the merge commit).

karthikreddy09:
Working on it … :-)
I rebased auto-install onto development and encountered test failures in static builds. Investigating the failures.

-Abhinay.

core/src/bundle/usBundleUtils.cpp
karthikreddy09: @jeffdiclemente 
GetLogSink() is a private method of BundleContext ... I have used std::cerr as the sink. Let me know if that is not the right thing to do.

jeffdiclemente: you can add these functions as ```friend```s in ```usBundleContext.h```. Then use ```GetBundleContext()```

saschazelzer: Please do not add more friend declarations. This is what `GetPrivate` is for when used in cpp files. Use `GetPrivate(GetBundleContext())->GetLogSink()`.

karthikreddy09: Awesome .. Thanks !!

karthikreddy09: @saschazelzer GetPrivate(GetBundleContext())->GetLogSink() does not work because BundleContextPrivate does not provide a GetLogSink method. It's in BundleContext class.

saschazelzer: Right, my mistake. You should be able to do something like `GetPrivate(GetBundleContext())->bundle->coreCtx->sink`. We can make the more elegant later.

core/src/bundle/usBundleUtils.cpp
```diff
@@ -67,6 +68,12 @@ void* dlsym(void *handle, const char *symbol)
 
 namespace us {
 
+// Private util function to return system bundle's log sink
+std::shared_ptr<LogSink> GetFrameworkLogSink()
+{
+  auto sink = GetBundleContext().GetLogSink();
```
jeffdiclemente: could be simplified to ``` return GetBundleContext().GetLogSink();```

jeffdiclemente:
This looks good to merge. 
Can you please include a change to ```usBundleEvent.cpp``` line 108  from
```
  case BundleEvent::BUNDLE_STOPPED:         return os << "BUNDLE_STOPPED"; 
```
to
````
  case BundleEvent::BUNDLE_STOPPED:         return os << "STOPPED"; 
````
This is a trivial issue which was missed in a previous PR. Thanks.

PR #111, opened 2016-07-26
106 - Add support for building for Android (std::to_string)

kevinleeMW:
Support building for Android, since "std::to_string" is not available in the latest Android SDKs


jeffdiclemente:
minor issue; indentation within ```namespace```. Let's keep with the code format standard and not indent inside the ```namespace```.

core/src/util/AndroidCompat.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "AndroidCompat.h"
+
+#include "usAny.h"
+
+namespace {
+template <typename T>
+std::string internal_to_string(T val)
+{
+  return us::Any(val).ToString();
```
saschazelzer: I think we should just use a `std::stringstream` here directly instead of going through `us::Any`. The any class uses a string stream as well, but it also does a dynamic memory allocation on construction.

core/src/util/AndroidCompat.h
```diff
@@ -0,0 +1,56 @@
...
+std::string to_string(unsigned long val);
+std::string to_string(long long val);
+std::string to_string(unsigned long long val);
+std::string to_string(float val);
+std::string to_string(double val);
+std::string to_string(long double val);
+}
+
+namespace std {
+using namespace detail_AndroidCompat;
```
saschazelzer: I am not really in favor of adding types to the `std` namespace. It is disallowed by the standard (with some exceptions) and it requires us consistently include the AndroidCompat header everywhere we use `std::to_string`. Not including it will work until you really build for Android.

I suggest to create a "to_string()" function in the `us` namespace and make the android compat code an implementation detail of that function. Then we use our function instead of `std::to_string`. Of course one could still use `std::to_string` by accident, but it feels a little safer to me this way.

kevinleeMW: Sascha,

I just wanted to clarify: does your suggestion entail having to include an AndroidCompat header (that contains `us::to_string()`) everywhere we need to use `to_string()`?

saschazelzer: We need to include "some" header containing the `us::ToString` function 
declaration (I suggest to use our camel case style, makes it also easier 
to search for the std variant). If we can re-use an existing one, like 
usUtil_p.h, that would be fine. My point is, that it would be nice to 
confine the android specifics to the function definition in the 
according cpp file.

kevinleeMW: Added "us::ToString()" ti usUtils_p.h / usUtils.cpp.
Added hard-coded conversions using std::ostringstream in test code.

core/src/util/usUtils.cpp
```diff
@@ -388,6 +388,37 @@ std::string GetLastErrorStr()
...
 }
 
+//-------------------------------------------------------------------
+// Android Compatibility functions
+//-------------------------------------------------------------------
+
+namespace {
+
+template <typename T>
+std::string internal_to_string(T val)
```
jeffdiclemente: I think you can simplify this and remove the overloaded ```ToString()``` declarations by making ```ToString``` a template function. Doing this would also allow you to use the template ```ToString``` function in the test code.

karthikreddy09: +1

core/src/util/usUtils.cpp
```diff
@@ -388,6 +388,37 @@ std::string GetLastErrorStr()
...
+{
+#if defined(__ANDROID__)
+#include <sstream>
```
karthikreddy09: #include in the middle of a file is odd ... I understand you want to minimize the number of conditional compile blocks and unnecessary includes. What do you think about moving this function to the top of the file and using a separate function body for the if and the else block

```
#if defined(__ANDROID__)
#include <sstream>
Std::string ToString(...)
{
...
}
#else
Std::string ToString(...) { return std::to_string(...); }
#endif
```

jeffdiclemente:
This looks good to merge. 
Give @saschazelzer a little time to respond with any comments.

saschazelzer:
Looks good to me, thanks for working on this.

In the long run, I think we should consider using a specialized `IdToString(long)` function, which efficiently converts integral numbers to strings without any streams and locales involved. The places in the framework really only need to convert ids to strings and the `GetMatchingServiceListeners()` function would probably benefit from such an optimized conversion routine.

core/src/util/usUtils_p.h
```diff
@@ -64,26 +68,19 @@ void MakePath(const std::string& path);
 
 US_Core_EXPORT void RemoveDirectoryRecursive(const std::string& path);
 
-}
+} // namespace fs
 
-}
```
karthikreddy09: nice cleanup !

karthikreddy09:
Looks good to merge !

PR #118, opened 2016-08-05
43 header reorg

kevinleeMW:
Summary:
- Change "core" to "framework" directory
- Add subdirectories for clear separation of headers from different bundles
- Remove "us" prefix from file names (in rare cases, replace with "cppmicroservices")
- Move from "us" to "cppmicroservices" namespace
- Install only public headers (including "private" headers in "detail" subdirectory)
- Clean up and re-organize "#include" directives

Note:
- I tried to rename the files using "git mv", but a couple of files didn't work quite right. Reviewers may not be able to view a diff side-by-side for those files.
- "Private" headers inside "detail" directory all include "detail" namespace, EXCEPT ServiceTracker.tpp. This was done so that the public header "ServiceTracker.h" does not also gain a "detail" namespace.


cmake/MacroCreateBundle.cmake
```diff
@@ -105,8 +111,6 @@ endif()
 
 set_target_properties(${${PROJECT_NAME}_TARGET} PROPERTIES
   SOVERSION ${${PROJECT_NAME}_VERSION}
-  PUBLIC_HEADER "${${PROJECT_NAME}_PUBLIC_HEADERS}"
-  PRIVATE_HEADER "${${PROJECT_NAME}_PRIVATE_HEADERS}"
```
saschazelzer: Is there a reason to not use the `PUBLIC_HEADER` property any more?

cmake/MacroCreateBundle.cmake
```diff
@@ -128,31 +132,32 @@ if(${PROJECT_NAME}_RESOURCES)
...
+          EXPORT ${PROJECT_NAME}Targets
           RUNTIME DESTINATION ${RUNTIME_INSTALL_DIR} ${US_SDK_INSTALL_COMPONENT}
           LIBRARY DESTINATION ${LIBRARY_INSTALL_DIR} ${US_SDK_INSTALL_COMPONENT}
-          ARCHIVE DESTINATION ${ARCHIVE_INSTALL_DIR} ${US_SDK_INSTALL_COMPONENT}
-          PUBLIC_HEADER DESTINATION ${HEADER_INSTALL_DIR} ${US_SDK_INSTALL_COMPONENT}
-          PRIVATE_HEADER DESTINATION ${HEADER_INSTALL_DIR} ${US_SDK_INSTALL_COMPONENT})
+          ARCHIVE DESTINATION ${ARCHIVE_INSTALL_DIR} ${US_SDK_INSTALL_COMPONENT})
+
+  install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/cppmicroservices DESTINATION ${HEADER_INSTALL_DIR})
+  install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/cppmicroservices DESTINATION ${HEADER_INSTALL_DIR})
```
saschazelzer: Same question as above, was there a problem with the `PUBLIC_HEADER` target property?

saschazelzer:
Thanks for working on this massive change!

I have a few remarks / questions:
- Since we are at it, could you also remove the `_p` suffix from the header files in the details folder?
- When you said "remove the us prefix from all files", I didn't realize that you really meant *all* files. We should add the prefix back in some cases in my opinion:
 * **Executable and library names** On Linux, these are typically installed into a common `lib` and `bin` folder (the current defaults need to be adapted) and I think the names need some prefix to better distinguish them from all the installed files.
 * **CMake file and function / macro names** Basically the same reasoning as above. Although the custom CMake files typically are installed into a project specific sub-directory, this directory can be added as a search path in external projects. Then a call to `include(FunctionEmbedResources)` misses some CppMicroServices specific information. Same goes for calling the CMake function later.
 * **Include guards** Header files now need to be prefixed when being included, but their include guards are not scoped in any way. A define like `BUNDLE_H` is more likely to clash with some other project's define then `US_BUNDLE_H`.
- The default installation sub-directories need to be updated (removing the `CppMicroServices` sub-directory in the Linux case), but that issue was there before your modifications. I could adapt them after your merge myself.

cmake/MacroCreateBundle.cmake
kevinleeMW: Sascha,
I couldn't find an easy way to preserve the header file directory hierarchy using the PUBLIC_HEADER target property (unless we specify different HEADER_INSTALL_DIRs).
I searched online for a solution for the needs, and thought "install(DIRECTORY ...)" would work the best.

Do you have any recommendations for installing public headers with directory hierarchy preserved?

kevinleeMW:
@saschazelzer ,

I agree with your feedback. One minor suggestion: how about doing header include guards with "cppmicroservices" prefix? (e.g. `CPPMICROSERVICES_BUNDLE_H`)

cmake/MacroCreateBundle.cmake
saschazelzer: Thanks for the explanation. No, I do not have much experience with that - your solutions looks good to me. There was just some Mac OS specifics to the PUBLIC_HEADER property when creating a Mac OS "framework", which was once done if I remember correctly, but not any more.

saschazelzer:
Sure, good point.

saschazelzer:
+1 for merging

framework/src/bundle/BundleResourceBuffer.cpp
```diff
@@ -35,7 +35,9 @@
 #define REMOVE_LAST_NEWLINE_IN_TEXT_MODE 1
 #endif
 
-namespace us {
+namespace cppmicroservices {
+  
+  namespace detail {
```
jeffdiclemente: indentation

jeffdiclemente:
Are we going to change the generated ```us_resource.cpp``` files to ```cppmicroservices_resource.cpp```?

framework/src/bundle/BundleResourceBuffer.cpp
kevinleeMW: Thanks for catching this!

kevinleeMW:
@jeffdiclemente 

Yes, I will change it to cppmicroservices_resources.cpp.
While I was working on this, I also noticed that the directories in which these resources are put into are called "us_${US_RESOURCE_TARGET}". This creates directories like:

framework/us_CppMicroServices
framework/test/us_usFrameworkTestDriver
framework/test/bundles/libA/us_TestBundleA

@saschazelzer @jeffdiclemente @karthikreddy09 
Does this "us_" prefix serve a purpose like other instances where we want to preserve them?
If not, I'll remove these prefixes as well.

saschazelzer:
I think you can remove these directory prefixes.

On August 9, 2016 10:27:13 PM GMT+02:00, kevinleeMW <notifications@github.com> wrote:
>@jeffdiclemente 
>
>Yes, I will change it to cppmicroservices_resources.cpp.
>While I was working on this, I also noticed that the directories in
>which these resources are put into are called
>"us_${US_RESOURCE_TARGET}". This creates directories like:
>
>framework/us_CppMicroServices
>framework/test/us_usFrameworkTestDriver
>framework/test/bundles/libA/us_TestBundleA
>
>@saschazelzer @jeffdiclemente @karthikreddy09 
>Does this "us_" prefix serve a purpose like other instances where we
>want to preserve them?
>If not, I'll remove these prefixes as well.
>
>---
>You are receiving this because you were mentioned.
>https://github.com/CppMicroServices/CppMicroServices/pull/118#issuecomment-238680091

-- 

jeffdiclemente:
+1 for merging after changing the ```us_resource.cpp``` files to ```cppmicroservices_resource.cpp``` and the  ```us_``` prefix on generated folder names mentioned above.

kevinleeMW:
Thanks for all your feedback and help, everyone!

PR #122, opened 2016-08-10
General cleanup

saschazelzer:
This PR contains a couple of commits cleaning up different things. No functional changes.

I want to make sure this is in everyone's interest. Please see the commit list for a summary.


CMakeLists.txt
```diff
@@ -200,11 +202,10 @@ if(WIN32 AND NOT CYGWIN)
   set(default_header_install_dir include/)
   set(default_auxiliary_install_dir share/)
 else()
-  string(TOLOWER ${PROJECT_NAME} PROJECT_NAME_LOWER)
   set(default_runtime_install_dir bin/)
-  set(default_library_install_dir lib/${PROJECT_NAME_LOWER})
-  set(default_archive_install_dir lib/${PROJECT_NAME_LOWER})
-  set(default_header_install_dir include)
+  set(default_library_install_dir lib/)
```
jeffdiclemente: This change breaks the ```framework\examples\makefile``` example. ```framework\examples\makefile\Makefile``` needs to be changed as well.

saschazelzer: Thanks, I still had  the old files lying around when running the tests.

On August 11, 2016 2:05:33 PM GMT+02:00, Jeff DiClemente <notifications@github.com> wrote:
>>    set(default_runtime_install_dir bin/)
>> -  set(default_library_install_dir lib/${PROJECT_NAME_LOWER})
>> -  set(default_archive_install_dir lib/${PROJECT_NAME_LOWER})
>> -  set(default_header_install_dir include)
>> +  set(default_library_install_dir lib/)
>
>This change breaks the ```framework\examples\makefile``` example.
>```framework\examples\makefile\Makefile``` needs to be changed as well.
>
>-- 
>You are receiving this because you were assigned.
>https://github.com/CppMicroServices/CppMicroServices/pull/122/files/a818fb88a0cbb6a80d255ab73e7ce37716c491aa..c1e7629b0a8b4fc4a235920c08e2a59cfe874f73#r74409731

-- 

framework/include/cppmicroservices/Constants.h
```diff
@@ -43,7 +43,7 @@ namespace Constants {
 
 /**
  * Location identifier of the OSGi <i>system bundle </i>, which is defined
- * to be &quot;System Bundle&quot;.
+ * to be &quot;System %Bundle&quot;.
  */
 US_Framework_EXPORT extern const std::string SYSTEM_BUNDLE_LOCATION; // = "System Bundle";
```
kevinleeMW: Should we also move from `US_` to `CPPMICROSERVICES_` prefix here?

framework/src/CMakeLists.txt
```diff
@@ -63,33 +63,33 @@ set(_srcs
...
+  bundle/BundleStorage.h
+  bundle/BundleStorageFile.h
+  bundle/BundleStorageMemory.h
+  bundle/BundleThread.h
+  bundle/BundleUtils.h
+  bundle/CoreBundleContext.h
+  bundle/Debug.h
+  bundle/Fragment.h
+  bundle/ModuleObjFile.h
+  bundle/Resolver.h
```
kevinleeMW: Thanks for catching this, Sascha. Also, I didn't realize that you wanted to remove the "_p" suffix from all headers. Awesome!

framework/src/bundle/BundleHooks.h
```diff
@@ -23,7 +23,7 @@
 #ifndef CPPMICROSERVICES_BUNDLEHOOKS_P_H
```
kevinleeMW: I made the header include guards changes to reflect the filenames, so I think removing the (and remove the `_P` suffix makes sense. However, I checked to see if there are headers that share the same name in different locations, and I found one public and one private version of "ServiceListenerHook.h". I don't know what the right move forward would be; perhaps we can just leave the current header include guards as is for now.

framework/test/TestUtils.h
```diff
@@ -79,7 +79,7 @@ class HighPrecisionTimer
 #endif
 };
 
-// Copied from usUtils_p.h/usUtils.cpp
+// Copied from usUtils.h/usUtils.cpp
```
kevinleeMW: Remove the "us" prefix from this comment

httpservice/CMakeLists.txt
```diff
@@ -25,12 +25,12 @@ else()
...
 
 set(_private_headers
-  src/HttpOutputStreamBuffer_p.h
-  src/HttpServlet_p.h
-  src/HttpServletRequest_p.h
-  src/HttpServletResponse_p.h
-  src/ServletConfig_p.h
-  src/ServletContainer_p.h
+  src/HttpOutputStreamBuffer.h
+  src/HttpServlet.h
```
kevinleeMW: An extension of my comment about header guards: there are also five headers from httpservice that have the same file name, one public and one private.

framework/src/bundle/ModuleMachOFile.cpp
```diff
@@ -19,7 +19,7 @@
 
 =============================================================================*/
 
-#include "ModuleObjFile_p.h"
+#include "ModuleObjFile.h"
```
karthikreddy09: I think we should replace "Module" with "Bundle" for these file names. The classes inside these files have "Bundle" as prefix.

framework/include/cppmicroservices/Constants.h
saschazelzer: I was thinking about this as well but thought I'd leave it like this for now. It is not something clients come in contact with directly.

framework/src/bundle/BundleHooks.h
saschazelzer: Good point, I will have a look.

framework/src/bundle/ModuleMachOFile.cpp
saschazelzer: Will do.

jeffdiclemente:
Thanks for taking the time to do a thorough cleanup of the code Sascha!
+1 for a merge.

saschazelzer:
Thanks for the reviews!

PR #125, opened 2016-08-11
Move virtual methods to BundlePrivate and fix the system bundle.

saschazelzer:


framework/include/cppmicroservices/Framework.h
```diff
@@ -140,7 +148,9 @@ class US_Framework_EXPORT Framework : public Bundle
      *
      * @throws std::runtime_error If this Framework could not be started.
      */
+#ifdef DOXYGEN_RUN
```
karthikreddy09: Why not use a single `#ifdef` around all these methods ?

karthikreddy09:
Thanks for making these changes.... Looks good for merge !

framework/include/cppmicroservices/Framework.h
saschazelzer: There is no particular reason... making the ifdef scope as small as possible felt right to me.

PR #127, opened 2016-08-22
fix building on FreeBSD (dl is linux only, bsd has it with libc)

mpkh:
seems like not relevant for development branch

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -340,7 +340,7 @@ endif()
 
 # link libraries for the CppMicroServices lib
 set(US_LINK_LIBRARIES )
-if(UNIX)
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   list(APPEND US_LINK_LIBRARIES dl)
 endif()
 


saschazelzer:
Thanks for the pull request. Normally, we only accept PRs targeting the development branch, see our [contribution guidelines](https://github.com/CppMicroServices/CppMicroServices/blob/development/CONTRIBUTING.md) for the details.

The issues still is relevant for the development branch. Instead of relying on `CMAKE_SYSTEM_NAME`, I suggest to use a `find_library` call for `dl` and link to it only if the resulting variable is valid.

mpkh:
Strange but on development branch I can't see manual `dl` linking there
https://github.com/CppMicroServices/CppMicroServices/blob/development/CMakeLists.txt

saschazelzer:
Yes, the project structure changed. Now it is here: https://github.com/CppMicroServices/CppMicroServices/blob/development/framework/CMakeLists.txt#L23

mpkh:
@saschazelzer okay I'll try to elaborate it

PR #132, opened 2016-08-28
131 auto install and bundle registry exceptions

### LINKED ISSUE #131: Auto install of executable bundles leads to exceptions that are hidden
The auto-install logic for executable bundles triggers an exception in the `BundleRegistry` class, when it tries to install all bundle archives which may have been persisted from a previous framework run.

Currently, we do not have persistence, but the code finds the archives from the previously auto-installed bundles and tries to install them again (thinking they cam from a previous run and need to be re-installed). This leads to an exception which was never seen because it is caught and handled as a framework event. However, the logic above is executed during framework initialization and no framework event listeners can possibly be installed at that time (there is no bundle context for the system bundle yet).

Proposed solution:
 1. Auto-install executable bundles *after* the bundle registry loaded
 1. Print possible exceptions to std::cerr instead of using framework events, until we support *init framework listeners*. It has the disadvantage of not being user re-directable right now, but is better compared to dropping them on the floor as we do currently.

saschazelzer:
This fixes #131 according to the proposed solutions.

### DIFF
--- a/framework/src/bundle/BundleRegistry.cpp
+++ b/framework/src/bundle/BundleRegistry.cpp
@@ -67,12 +67,12 @@ std::vector<Bundle> BundleRegistry::Install(const std::string& location,
                                             bool allowExecutable)
 {
   CheckIllegalState();
-  
+
   if (!allowExecutable && !IsSharedLibrary(location))
   {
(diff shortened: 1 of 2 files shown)


karthikreddy09:
+1 for merge

jeffdiclemente:
Looks good to me. +1 for merge.

PR #136, opened 2016-09-02
Test CI with OS X

karthikreddy09:
Add Mac OS X to travis CI builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,20 +21,6 @@ matrix:
               packages: valgrind
           compiler: gcc
           env: BUILD_CONFIGURATION=1
-        - os: linux
-          addons:
-            apt:
-              sources: ubuntu-toolchain-r-test
-              packages: valgrind
-          compiler: clang
-          env: BUILD_CONFIGURATION=0
-        - os: linux
-          addons:
-            apt:
-              sources: ubuntu-toolchain-r-test
-              packages: valgrind
-          compiler: clang
-          env: BUILD_CONFIGURATION=1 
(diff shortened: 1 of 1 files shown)


saschazelzer:
Awesome, having Mac OS supported by Travis CI will be great.

A few comments:
 * If you want to test CI config changes, just pushing the branch itself is enough and will trigger the CI hooks. You don't need a PR.
 * Some elements in the build matrix have `addons: &valgrind` and some don't. What is the purpose of the `&` ?
 * It is unfortunate that we need to expand the build matrix manually. When we re-enable the single-threaded build configuration, this will double the number of entries. But given my understanding of the Travis CI config file and the need for the `MACOSX_DEPLOYMENT_TARGET` environment variable, I couldn't think of a cleaner way :-/

Thanks for working on this!

saschazelzer:
Oh, and the Linux Clang configurations are not supported on Ubuntu Trusty. Clang just doesn't play well with `libstdc++` from gcc with C++11 enabled, so I suggest to just remove these entries from the build matrix.

karthikreddy09:
I noticed the push/branch CI nodes but never made the connection. 

(&) is used to create anchor nodes in YAML. The same nodes can be later referenced using a (*). The anchor and reference nodes can be used to remove redundancy in the file. I haven't yet figured how to do it right in our yml file

karthikreddy09:
@CppMicroServices/developers let me know if you are OK with merging this PR.

saschazelzer:
+1 for merging

This is really nice to have!

Issue #135, opened 2016-08-31
Fix up bundle property and manifest header handling


saschazelzer:
Due to historical reasons. the function names and their implementation  which relate to the handling of properties and bundle manifest headers are not in line with OSGi standards. This will be changed by deprecating the existing functions and adding new ones, satisfying the standard requirements.

### Bundle

Deprecate:

* `std::map<std::string, Any> GetProperties() const`
* `Any GetProperty(const std::string& key) const`
* `std::vector<std::string> GetPropertyKeys() const `

New:

* `AnyMap GetHeaders() const`

### BundleContext

New:

* `Any GetProperty(const std::string& key) const`
* `AnyMap GetProperties() const`

PR #138, opened 2016-09-07
135 property and header handling

saschazelzer:
This PR introduces (partial) OSGi compliance for retrieving bundle manifest headers and framework properties:
- Changes as outlined in #135
- Localization of header attributes is not supported
- A new `AnyMap` class to avoid testing an `Any` for different container types
- The framework properties `FRAMEWORK_VERSION` and `FRAMEWORK_VENDOR` are now always added. Also, the `FRAMEWORK_STORAGE` property is always set.

Note that my editor fixed up all the white-space issues in `BundleManifest.cpp` so the diff unfortunately has a lot of noise.


framework/test/BundleTest.cpp
```diff
@@ -287,6 +299,11 @@ void frame037a()
 
   US_TEST_CONDITION(buExec.GetBundleContext().GetDataFile("").substr(0, baseStoragePath.size()) == baseStoragePath, "Test for valid data path")
   US_TEST_CONDITION(buExec.GetBundleContext().GetDataFile("bla").substr(0, baseStoragePath.size()) == baseStoragePath, "Test for valid data file path")
+
+  US_TEST_CONDITION(buExec.GetBundleContext().GetProperty(Constants::FRAMEWORK_UUID).Empty() == false, "Test for non-empty framework uuid property")
+  auto props = buExec.GetBundleContext().GetProperties();
+  US_TEST_CONDITION_REQUIRED(props.empty() == false, "Test for non-empty bundle props")
+  US_TEST_CONDITION(props.count(Constants::FRAMEWORK_VERSION) > 0, "Test for existing framework version prop")
```
jeffdiclemente: Will we ever have more than one ```FRAMEWORK_VERSION``` property? If not, I'd suggest making this check for the existence of only one ```FRAMEWORK_VERSION``` property.

framework/include/cppmicroservices/Bundle.h
```diff
@@ -398,6 +425,9 @@ class US_Framework_EXPORT Bundle
    * @return The value of the requested property, or an empty string
    *         if the property is undefined.
    *
+   * @deprecated Since 3.0, use GetHeaders() or BundleContext::GetProperty
```
karthikreddy09: I would suggest adding a compiler warning ...

```
#ifdef __GNUC__
#define DEPRECATED __attribute__((deprecated))
#elif defined(_MSC_VER)
#define DEPRECATED __declspec(deprecated)
#else
#pragma message("WARNING: DEPRECATED has no implementation for this compiler")
#define DEPRECATED
#endif

//...

//don't use me any more
DEPRECATED void OldFunc(int a, float b);

//use me instead
void NewFunc(int a, double b);
```

framework/include/cppmicroservices/AnyMap.h
```diff
@@ -0,0 +1,322 @@
...
+   * map.GetDotted("three.b.1"); // returns Any(8)
+   * \endcode
+   *
+   * @param key The key hierachy to query.
+   * @return A reference to the key's value.
+   *
+   * @throws std::invalid_argument if the \c Any value for a given key is not of type \c AnyMap or \c std::vector<Any>.
+   * std::out_of_range if the key is not found or a numerical index would fall out of the range of an \c int type.
+   */
+  mapped_type& GetDotted(const key_type& key);
```
karthikreddy09: nitpick .... method naming is inconsistent in this class. some are lowercase and otehrs are PascalCase.

framework/test/BundleTest.cpp
saschazelzer: Okay, will do. I was only concerned about existence in the test. We do not support multi maps.

framework/include/cppmicroservices/Bundle.h
saschazelzer: Yes, I missed to mention this in the initial comment. I did not do this on purpose just yet. Are you ready to handle these kind of compiler warnings? Some other projects I worked with had troubles suppressing third party warnings, especially when compiling with `-Werror` (warnings as errors). I just want to make sure this is okay for you.

framework/include/cppmicroservices/AnyMap.h
saschazelzer: This is somewhat intentional, but I am unsure what would be best. The lower case methods mimic the STL functions for containers. The pascal case is our standard style. What do you prefer. Same style (which one)?

framework/include/cppmicroservices/Bundle.h
karthikreddy09: We are okay with compiler warnings. Our build system provides a mechanism to temporarily disable treating compiler warnings as errors. There are processes in place to ensure the compiler warnings are addressed eventually.

framework/include/cppmicroservices/AnyMap.h
karthikreddy09: Since AnyMap is a Framework defined class, I think using PascalCase uniformly will help with code readability.

saschazelzer: Just to clarify, STL containers and iterators are supposed to provide a few standard typedefs, which are in `underscore_style`. So to comply with that, we would still end up with a mixture of naming styles. So in that light, maybe everything in that class should have `underscore_type` style...

framework/include/cppmicroservices/Constants.h
```diff
@@ -56,80 +56,109 @@ US_Framework_EXPORT extern const std::string SYSTEM_BUNDLE_SYMBOLICNAME; // = "s
...
  * functionality.
  *
- * The header value may be retrieved via the \c Bundle::GetProperty method.
+ * The header value may be retrieved from the \c AnyMap object
+ * returned by the \c Bundle::GetHeaders() method.
  */
 US_Framework_EXPORT extern const std::string BUNDLE_DESCRIPTION; // = "bundle.description";
 
 /**
  * Manifest header identifying the bundle's menifest version.
```
jeffdiclemente: typo; "menifest"

saschazelzer:
I decided to split the `AnyMap` class into a pure STL container class `any_map` following STL naming conventions and a derived class `AnyMap` adding convenience functions. These convenience functions use our PascalCase naming. They are still mixed with the inherited under_score style functions, but I feel reluctant to create a STL base container using PascalCase.

saschazelzer:
@karthikreddy09 Are you still thinkgin about the "any map" issue? What is your conclusion?

karthikreddy09:
@saschazelzer Sorry, I have been procrastinating on this issue. 

I am having a hard time figuring out why a user would want to iterate the map returned by `Bundle::GetHeaders` and check for `typeid` of values. 

The primary use case would be to retrieve a specific value out of the headers. `GetUsingCompoundKey` will solve this use case. Another one I can think of is to print the map, the stream operator overload serves this purpose. 

Even if the user wants to iterate the map returned by Bundle::GetHeaders, Would it not be sufficient to document the fact that the only possible value types are : scalar type, `std::string`, `std::vector<us::Any>`, `AnyMap`(case insensitive due to manifest header requirement)

Could you provide an example usecase or a code snippet to help me understand why a user would want to iterate and check the `typeid` of values inside the returned `AnyMap`. 

You mentioned about `AnyMap` replacing all direct usages of STL associative containers in a Any type. Did you mean all maps with key_type=`std::string` ? or any key_type ?  A `us::Any` object with content_type=`std::map<int,us::Any>` cannot be stored in a AnyMap.

saschazelzer:
I agree that for our own specific use cases today, we would not need the more generic `AnyMap` class.

However, I am not comfortable with focusing on the origins of an `Any` instance (e.g. the `GetHeaders` function) and making assumptions based on that origin. The `Any` class is generic framework API and as such independent of any assumptions about its contained types.

1. Something like `AtCompoundKey` returns a `Any`. This could hold some map (e.g. the key from a manifest files references a JSON object). Prior knowledge (e.g. from documentation) could be used to extract the actual value using the documented map type. But I am concerned about the generic case without prior knowledge.

1. `AtCompoundKey` is part of the type-erasing `AnyMap` class and hence "just works". As a free function, it would need to branch out to different map types to be able to recurse into their contents in order to resolve the compound key.

1. It is more elegant IMHO to test for only one map type instead of many, in case the underlying type needs to be accessed (which is the case in our `ToString` and `ToJSON` functions where the compiler branches out for us, selecting the correct overload, or which would also be the case for a free `AtCompoundKey` function implementation withoug `AnyMap`).

1. An example of recursive iteration of an arbitrary `Any` is the webconsole bundle (the code is not part of the development branch yet). It takes an arbitrary `Any` object (typically from the `GetHeaders` function, but it should not need to assume that) and recursively iterates the Any contents to display it in structured way (e.g. a tree) on a web page. Granted, in that case using the `ToJSON` function will be easier because JSON is native to JavaScript.

The `Any` class can hold a `std::map<int,Any>` type (if a stringification function is provided), but I do not see how we could elegantly type-erase the key. The proposal here tries to homogenize the `map<string,Any>` case for different containers types and comparators (but still a fixed set).

karthikreddy09:
It sounds reasonable to simplify for the most common use case of std::map. 

+1 for merge.

PR #143, opened 2016-09-16
Fix Windows static destruction corner case

jeffdiclemente:
It's been observed that when CppMicroServices is used via JNI, during process shutdown, even if the Framework was explicitly shut down, a crash can still occur.  The fix is to remove unnecessary access to thread and synchronizxation objects unless required.  

This also includes a minor fix to resolve a GCC build warning.

### DIFF
--- a/framework/test/singletonframework/CMakeLists.txt
+++ b/framework/test/singletonframework/CMakeLists.txt
@@ -1,11 +1,17 @@
+# Create a non-bundle shared library for the sole
+# purpose of testing static destruction on Windows.
+#
 project(singletonframework)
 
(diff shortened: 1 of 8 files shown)


framework/test/singletonframework/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+project(singletonframework)
```
karthikreddy09: A short comment about the purpose of this library would be helpful

jeffdiclemente: I agree. I'll add one within the header or source file. I think that'll be more visible.

jeffdiclemente:
discarding pull request. This needs more work...

PR #147, opened 2016-09-26
Fix windows static destruction corner case

jeffdiclemente:
In certain cases such as when using CppMicroServices through a JNI bridge, Windows crashes still occur even if the Framework is explicitly stopped and has finished its stop before static destruction.

This fix attempts to handle this corner case by returning immediately if the Framework has already been stopped. Integration testing shows that this crash no longer occurs with this fix.

### DIFF
--- a/framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
+++ b/framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
@@ -51,7 +51,7 @@ void BundleAbstractTracked<S,TTT,R>::SetInitial(const std::vector<S>& initiallis
     for(typename std::list<S>::const_iterator item = initial.begin();
       item != initial.end(); ++item)
     {
-      DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::setInitial: " << (*item);
+      DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::setInitial: " << (*item) << "\n";
     }
   }
 }
@@ -80,7 +80,7 @@ void BundleAbstractTracked<S,TTT,R>::TrackInitial()
       if (tracked[item])
       {
         /* if we are already tracking this item */
-        DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::trackInitial[already tracked]: " << item;
+        DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::trackInitial[already tracked]: " << item << "\n";
         continue; /* skip this item */
       }
       if (std::find(adding.begin(), adding.end(), item) != adding.end())
@@ -88,12 +88,12 @@ void BundleAbstractTracked<S,TTT,R>::TrackInitial()
         /*
          * if this item is already in the process of being added.
          */
-        DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::trackInitial[already adding]: " << item;
+        DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::trackInitial[already adding]: " << item << "\n";
         continue; /* skip this item */
       }
       adding.push_back(item);
     }
-    DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::trackInitial: " << item;
+    DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::trackInitial: " << item << "\n";
     TrackAdding(item, R());
     /*
      * Begin tracking it. We call trackAdding
@@ -125,14 +125,14 @@ void BundleAbstractTracked<S,TTT,R>::Track(S item, R related)
       if (std::find(adding.begin(), adding.end(),item) != adding.end())
       {
         /* if this item is already in the process of being added. */
-        DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::track[already adding]: " << item;
+        DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::track[already adding]: " << item << "\n";
         return;
       }
       adding.push_back(item); /* mark this item is being added */
     }
     else
     { /* we are currently tracking this item */
-      DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::track[modified]: " << item;
+      DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::track[modified]: " << item << "\n";
       Modified(); /* increment modification count */
     }
   }
@@ -164,7 +164,7 @@ void BundleAbstractTracked<S,TTT,R>::Untrack(S item, R related)
     { /* if this item is already in the list
        * of initial references to process
        */
-      DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::untrack[removed from initial]: " << item;
+      DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::untrack[removed from initial]: " << item << "\n";
       return; /* we have removed it from the list and it will not be
                * processed
                */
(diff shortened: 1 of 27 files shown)


jeffdiclemente:
There is a sporadic build issue on Mac. I believe there is a race condition and haven't yet been able to nail down what exactly is happening as it only seems to reproduce in travis CI.

This issue occurs when one stops the framework and the framework instance then goes out of scope without ```WaitForStop``` being called.

saschazelzer:
This looks painful... it seems that there is an exception thrown, which is not caught by the CppMicroServices test macros (possibly a segfault) but by the ctest driver instead. Do you have some more clues already?

What you also could try is to call `bin/usFrameworkTestDriver FrameworkTest` directly in the `.travis.yml` file (without going through ctest) which might give more console outpout. It might also generate a core dump which we could analyze using lldb. If you want, I can help you with that in a separate branch.

jeffdiclemente:
This has been painful. I don't have any more clues yet.
Thanks for the suggestion to use ```.travis.yml```, I'll try that.

If you have time, yes, any help would be much appreciated.

saschazelzer:
Okay, I will try and modify the travis config in a new branch.

jeffdiclemente:
Trying to catch any exception hasn't been successful yet...

jeffdiclemente:
It appears what I initially thought was the issue may be true. There is a possible race in the ```ServiceHooks::Open``` and  ```ServiceHooks::Close``` methods around the ```listenerHookTracker``` object.

I still need to revert all the code changes used for debugging...

jeffdiclemente:
@CppMicroServices/developers please review.

I kept the changes to catch all exceptions in the ```US_TEST_END``` macro and to output the thread id for all diagnostic logger messages.

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -51,7 +51,7 @@ void BundleAbstractTracked<S,TTT,R>::SetInitial(const std::vector<S>& initiallis
     for(typename std::list<S>::const_iterator item = initial.begin();
       item != initial.end(); ++item)
     {
-      DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::setInitial: " << (*item);
+      DIAG_LOG(*bc->GetLogSink()) << "BundleAbstractTracked::setInitial: " << (*item) << "\n";
```
karthikreddy09: Can we incorporate the "\n" into the DIAG_LOG macro or the logsink. Its painful to have to specify it every time. And can we use std::endl instead of "\n" ?

framework/src/util/FrameworkFactory.cpp
```diff
@@ -40,6 +39,14 @@ struct CoreBundleContextHolder
   ~CoreBundleContextHolder()
   {
     auto const state = ctx->systemBundle->state.load();
+
+    // The framework may have already completed its stop. In this case,
+    // nothing needs to be done; the framework can be destroyed.
+    // Allowing a call to WaitForStop can cause a crash during static DLL
+    // destruction on Windows even if the framework was explicitly stopped
+    // prior to static destruction.
+    if (Bundle::STATE_INSTALLED == state || Bundle::STATE_RESOLVED == state) return;
```
karthikreddy09: why not use unary operators and avoid the return statement in the middle of the function.
`if ((Bundle::STATE_INSTALLED | Bundle::STATE_RESOLVED) & state) `

karthikreddy09: I think you can get rid of the other return statement because of this early exit. it can simply become an if else

jeffdiclemente: Sure, I'll re-work the logic.

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
jeffdiclemente: > Can we incorporate the "\n" into the DIAG_LOG macro or the logsink.

Yes, I'll do that.

>  can we use std::endl instead of "\n"

I'd prefer to use ```\n``` and not flush the stream after each log message.

jeffdiclemente:
The latest travis build shows the failure still randomly occurs. 
I tried running xcode8 with thread sanitizer on my Mac and it didn't detect any issues.

[I'm currently adding xcode8 to travis](https://github.com/CppMicroServices/CppMicroServices/tree/add-xcode8-to-travisci)

At this point, I"m not sure what else I can do and believe I should merge this and add xcode8 w/ a thread sanitizer build configuration to travis. This way if this failure happens again we have a chance of getting some meaningful output.

@CppMicroServices/developers thoughts?

framework/include/cppmicroservices/detail/Log.h
```diff
@@ -64,13 +65,13 @@ class LogSink : public MultiThreaded<>, public std::enable_shared_from_this<LogS
...
 
-  LogMsg(LogSink& sink, const char* file, int ln, const char* func)
+  LogMsg(LogSink& sink, const std::thread::id tid, const char* file, int ln, const char* func)
     : enabled(false), buffer(), _sink(sink)
   {
     enabled = _sink.Enabled();
     if (enabled)
     {
-      buffer << "In " << func << " at " << file << ":" << ln << " : ";
+      buffer << "[" << tid << "] " << "In " << func << " at " << file << ":" << ln << " : ";
```
saschazelzer: Can't we just call `std::this_thread::get_id()` in-place here, instead of passing it as an additional argument to the `LogMsg` constructor? This would avoid the default argument and `thread` include directive in the header.

framework/src/service/ServiceHooks.cpp
```diff
@@ -75,10 +74,8 @@ void ServiceHooks::RemovedService(const ServiceReference<ServiceListenerHook>& /
 
 void ServiceHooks::Open()
 {
-  {
-    auto l = this->Lock(); US_UNUSED(l);
-    listenerHookTracker.reset(new ServiceTracker<ServiceListenerHook>(GetBundleContext(), this));
-  }
+  auto l = this->Lock(); US_UNUSED(l);
+  listenerHookTracker.reset(new ServiceTracker<ServiceListenerHook>(GetBundleContext(), this));
```
saschazelzer: Changing the lock scope in this way is not a good idea because the `->Open()` call below assumes that it is called outside of any lock and will subsequently call into some user call back function.

But I believe that you are right and this code is not properly synchronized with the `Close()` function, because the `listenerHookTracker` unique ptr could have been reset  just before we try to dereference it and call `Open()`. What I think we should do, is to use a `std::shared_ptr<ServiceTracker<ServiceListenerHook>>` instead of a `unique_ptr` for the `listenerHookTracker` member.Then we can keep a local `shared_ptr` inside the `Open()` call and use it outside of the lock scope.

saschazelzer:
+1 for adding a XCode8 with tsan build.

framework/src/service/ServiceHooks.cpp
jeffdiclemente: Yes, you're right. I forgot about the call back functions.

I was looking at this code with your suggestion in mind and a question popped into my head; does ```listenerHookTracker``` need to be ```reset()``` on each ```ServiceHooks:::Open()``` and ```ServiceHooks::Close()``` call? What if ```listenerHookTracker``` was only created once and allowed to be destroyed during the ```ServiceHooks``` object clean-up? 
Is there state within ```listenerHookTracker``` which requires re-creating the object each time ```ServiceHooks::Open()``` is called?

saschazelzer: That is a good question - I am not sure yet. I was trying to see how we could end up in such a race condition in the first place but I need more time for this. Also, the lock scope originally enclosed the full function body but was then changed. Maybe it was due to the callback issue, but I do not remember. Will investigate this further...

jeffdiclemente: The only reason I could find to ```reset()``` the ```listenerHookTracker``` each time was as an alternative to finding a place to instantiate ```listenerHookTracker``` once. Instantiating ```listenerHookTracker``` in the ```ServiceHooks``` constructor would have been ill-advised since ```listenerHookTracker``` is instantiated with the ```ServiceHooks``` ```this``` pointer.

jeffdiclemente:
I'm going to put this PR on hold for now and move the ```Log``` changes, the singleton framework test and the possible race condition in ```ServiceHooks.cpp```  into separate issues/PRs. There is too much going on in this one PR.

PR #148, opened 2016-09-29
89 activator lookup

karthikreddy09:


framework/test/TestingMacros.h
```diff
@@ -136,4 +136,23 @@ namespace cppmicroservices {
   STATEMENT ;                                                                                 \
   US_TEST_FOR_EXCEPTION_END(EXCEPTIONCLASS)
 
+
+#define US_TEST_FOR_NO_EXCEPTION_END()                                         \
+cppmicroservices::TestManager::GetInstance().TestPassed();                     \
+}                                                                              \
+catch (...) {                                                                  \
```
jeffdiclemente: I like the macro to test for no exceptions. I think it would be even more helpful to developers if any ```std::exception``` type information was displayed in the output.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -582,73 +582,77 @@ void BundlePrivate::Start(uint32_t options)
...
   std::exception_ptr res;
-
-  coreCtx->listeners.BundleChanged(BundleEvent(BundleEvent::BUNDLE_STARTING, MakeBundle(this->shared_from_this())));
-
-  try
+  auto const thisBundle = MakeBundle(this->shared_from_this());
+  coreCtx->listeners.BundleChanged(BundleEvent(BundleEvent::BUNDLE_STARTING, thisBundle));
+  
+  Any bundleActivatorVal = thisBundle.GetProperty(Constants::BUNDLE_ACTIVATOR);
+  bool useActivator = bundleActivatorVal.Empty() ? false : any_cast<bool>(bundleActivatorVal);
```
saschazelzer: This will throw if `useActivator` is not of type `bool`. Probably not what we want.

third_party/civetweb/civetweb.c
```diff
@@ -129,16 +129,26 @@ mg_static_assert(sizeof(void *) >= sizeof(int), "data type size check");
 
 #ifdef __MACH__
 
-#define CLOCK_MONOTONIC (1)
-#define CLOCK_REALTIME (2)
+#include <AvailabilityMacros.h>
```
saschazelzer: Thanks a lot for doing this (unrelated) work. Since this is a change in a third-party lib, it would be great if you could create a patch file (using e.g. git diff) and add it to the `third_party/patches/` folder and add a note in the `third_party/README` file, similar to the others. Thanks!

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -588,7 +588,20 @@ std::exception_ptr BundlePrivate::Start0()
...
+  
+  if(!bundleActivatorVal.Empty())
+  {
+    try
+    {
+      useActivator = any_cast<bool>(bundleActivatorVal);
+    }
+    catch (const BadAnyCastException& ex)
+    {
+      DIAG_LOG(*coreCtx->sink) << ex.what();
```
jeffdiclemente: I think it would be more useful to clients if this is a ```FrameworkEvent``` of type ```FRAMEWORK_WARNING``` instead so that clients can resolve potentially incorrect manifests.

PR #152, opened 2016-10-07
Build dummy resource file with correct compiler flags on OS X

karthikreddy09:


CMakeLists.txt
```diff
@@ -338,6 +338,21 @@ else()
...
+  # Issue #151: Since we explicitly compile CMakeResourceDependencies.cpp, we cache the OSX specific compiler flags
+  if(APPLE)
+    set(US_OSX_CXX_FLAGS ) 
+    if(CMAKE_OSX_SYSROOT)
+      list(APPEND US_OSX_CXX_FLAGS -isysroot ${CMAKE_OSX_SYSROOT})
+    endif()
+    if(CMAKE_OSX_ARCHITECTURES)
+      list(APPEND US_OSX_CXX_FLAGS -arch ${CMAKE_OSX_ARCHITECTURES})
+    endif()
+    if(CMAKE_OSX_SYSROOT)
```
saschazelzer: This looks like a copy and paste error, should be `if(CMAKE_OSX_DEPLOYMENT_TARGET)`

cmake/usFunctionEmbedResources.cmake
```diff
@@ -138,7 +138,7 @@ function(usFunctionEmbedResources)
       # section name is "us_resources" because max length for section names in Mach-O format is 16 characters.
       add_custom_command(
         OUTPUT ${_source_output}
-        COMMAND ${CMAKE_CXX_COMPILER} -isysroot ${CMAKE_OSX_SYSROOT} -c ${US_CMAKE_RESOURCE_DEPENDENCIES_CPP} -o stub.o
+        COMMAND ${CMAKE_CXX_COMPILER} ${US_OSX_CXX_FLAGS} -c ${US_CMAKE_RESOURCE_DEPENDENCIES_CPP} -o stub.o
```
saschazelzer: That function is part of our "CMake API", which means it is exposed to clients. As such, the `CMAKE_OSX_CXX_FLAGS` variable will never be set for client projects. I suggest to just move the added CMake block from the top-level CMakeLists.txt file into the usFunctionEmbedResources.cmake file.

PR #153, opened 2016-10-17
Consolidate LDAP queries of bundle/service properties using AnyMap

kevinleeMW:
With the AnyMap class available to work with ordered and unordered maps, LDAP query users can now work with both types of maps with a single LDAP query API.


kevinleeMW:
Will be merging these changes tomorrow EOD if there aren't any additional needs for changes.

PR #167, opened 2016-11-17
Fix coverity issues

karthikreddy09:
Fixed a few issues reported by Coverity


saschazelzer:
@jeffdiclemente Thanks for pointing out the changelog, but in that case I think no addition is required. The changelog contains changes compared to 2.1 but the issues being fixed here have been introduced after 2.1.

jeffdiclemente:
@saschazelzer you're right.
forget about updating CHANGELOG.md

Issue #144, opened 2016-09-21
Include major version number in library names and install dirs


saschazelzer:
Append the CppMicroServices major version number to shared library names and default installation directories. This helps identifying the major version in use and enables easier parallel installations of different major versions.

New default paths:
- `/usr/local/include/cppmicroservices3/` (include dir, containing `cppmicroservices` sub-dir)
- `/usr/local/lib/libCppMicroServices.so.3.0.0` (Linux symlink, already done like that)
- `C:/install/bin/CppMicroServices3.dll` (Windows DLL with major version number)
- `/usr/local/share/cppmicroservices3/` (auxiliary directory)

PR #168, opened 2016-11-18
144 major version in filenames

saschazelzer:
This is a PR for issue #144 with the following changes:

### Parallel major version installs

 - Executable now have the major version number appended, e.g. `usResourceCompiler3` and `usShell3`
 - Default include install dir is now: `${PREFIX}/CppMicroServices3`. E.g. `/usr/local/include/cppmicroservices3/cppmicroservices/Bundle.h`
 - On Windows, libraries have the major version number appended, e.g. `CppMicroServices3.dll` (on other platforms, names already contained the full version and symbolic links are set up)

### Parallel release/debug installs

 - By default, the debug postfix `d` is used, e.g. for debug builds on Linux we have `libCppMicroServicesd.so.3.0.0`, on Windows it is `CppMicroServices3d.dll`

### MinGW support

 - Fixed warnings and errors
 - Added Appveyor MinGW-w64 builds
 
### Other

 - Only build Release configurations on Appveyor to reduce work load
 - Library prefix (e.g. `lib`) and extensions (`.dll` etc.) are handled uniformly now via CMake configure_file of `GlobalConfig.h.in`.


framework/test/SharedLibraryTest.cpp
```diff
@@ -35,40 +35,34 @@ int SharedLibraryTest(int /*argc*/, char* /*argv*/[])
   US_TEST_BEGIN("SharedLibraryTest");
 
 #ifdef US_PLATFORM_WINDOWS
-  const std::string LIB_SUFFIX = ".dll";
   const char PATH_SEPARATOR = '\\';
```
jeffdiclemente: ```PATH_SEPARATOR``` could also be removed if ```DIR_SEP``` is used from ```TestingConfig.h```.

CMakeLists.txt
```diff
@@ -361,8 +364,8 @@ endif()
 if(MINGW)
   # suppress warnings about auto imported symbols
   set(US_CXX_FLAGS "-Wl,--enable-auto-import ${US_CXX_FLAGS}")
-  # we need to define a Windows version
-  set(US_CXX_FLAGS "-D_WIN32_WINNT=0x0500 ${US_CXX_FLAGS}")
+  # we need to define a Windows version (Windows Vista)
+  set(US_CXX_FLAGS "-D_WIN32_WINNT=0x0600 -DNTDDI_VERSION=0x06000000 ${US_CXX_FLAGS}")
```
saschazelzer: I was using the lowest version that contained the necessary API to compile the project successfully (unfortunately, I do not remember which part required this). Officially, Windows 7 can still be our minimum supported version.

framework/test/SharedLibraryTest.cpp
saschazelzer: Good catch.

saschazelzer:
Is anyone having some doubts / comments / opinions about this one?

jeffdiclemente:
I'm currently reviewing. I will provide feedback/approval today

Issue #162, opened 2016-10-31
Add integration tests for ResourceCompiler


ksubramz:
The tests are needed so that any changes to ResourceCompiler don't cause a surprising change in behavior. The bare minimum requirements are 

- Test injecting a manifest inside a bundle
- Test injecting any file into the zip blob
- If there's already a zip blob, test the merging (--zip-add)
- Test --bundle-file option (whether an archive is appended to the bundle)

Some extra comments:
- We need some sort of zipinfo mechanism to query the contents inside the bundle
- We can use CTests to do the testing

jeffdiclemente:
marking as closed by #169

Issue #161, opened 2016-10-28
Improve ResourceCompiler code


ksubramz:
- If manifest.json doesn't exist when using the `manifest-add` option, usResourceCompiler errors out with the following message 

```
Error: Error writing file to archive
```
  
I guess it's better if we check the path before we call `ZipArchive::AddResourceFile` and report error early and clearly that we can't find the file

 - In `ZipArchive::AddResourcesFromArchive`, we have a logic depending on whether the current entry in the zip archive is a file or a directory. Seems better to use the `mz_zip_reader_is_file_a_directory()` function to simplify this.

(will expand this issue as I find more)

ksubramz:
Bullets 4, 5, 6, 7 and 8 have been addressed in PR #169

PR #169, opened 2016-11-18
Tests for resource compiler

ksubramz:
Primarily addresses #162 but also addresses a few bullet points of #161

### DIFF
--- a/framework/test/ResourceCompilerTest.cpp
+++ b/framework/test/ResourceCompilerTest.cpp
@@ -125,10 +125,10 @@ static inline void testExists(const std::vector<std::string>& entryNames,
 static void testManifestAdd(const std::string& rcbinpath, const std::string& tempdir)
 {
   std::ostringstream cmd;
-  cmd << rcbinpath;
+  cmd << "'" << rcbinpath;
   cmd << " --bundle-name " << "mybundle";
   cmd << " --out-file " << tempdir << testing::DIR_SEP << "Example.zip";
-  cmd << " --manifest-add " << tempdir << testing::DIR_SEP << "manifest.json";
+  cmd << " --manifest-add " << tempdir << testing::DIR_SEP << "manifest.json'";
   int ret = std::system(cmd.str().c_str());
   US_TEST_CONDITION_REQUIRED(ret == 0, "Cmdline invocation in testManifestAdd returns 0")
 
@@ -147,11 +147,11 @@ static void testManifestAdd(const std::string& rcbinpath, const std::string& tem
 static void testManifestResAdd(const std::string& rcbinpath, const std::string& tempdir)
 {
   std::ostringstream cmd;
-  cmd << rcbinpath;
+  cmd << "'" << rcbinpath;
   cmd << " --bundle-name mybundle ";
   cmd << " --out-file Example2.zip ";
   cmd << " --manifest-add manifest.json ";
-  cmd << " --res-add resource1/resource1.txt";
+  cmd << " --res-add resource1/resource1.txt'";
 
   auto cwdir = testing::GetCurrentWorkingDirectory();
   testing::ChangeDirectory(tempdir);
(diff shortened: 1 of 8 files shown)


framework/test/ResourceCompilerTest.cpp
```diff
@@ -0,0 +1,448 @@
...
+  cmd << " --res-add resource1/resource1.txt";
+
+  auto cwdir = testing::GetCurrentWorkingDirectory();
+  testing::ChangeDirectory(tempdir);
+  std::system(cmd.str().c_str());
+  testing::ChangeDirectory(cwdir);
+
+  ZipFile zip(tempdir + testing::DIR_SEP + "Example2.zip");
+  US_TEST_CONDITION(zip.size() == 4, "Check number of entries of zip.")
+  US_TEST_CONDITION(zip[0].name == "mybundle/manifest.json", "Check name of entry 0.")
```
karthikreddy09: The test assumes the zip file contents are retrieved in a particular order. If the order of insertion changes in the implementation, the test needs to change. It is preferable to check for existence irrespective of the order.

tools/rc/ResourceCompiler.cpp
```diff
@@ -411,36 +413,54 @@ int main(int argc, char** argv)
...
     return_code = EXIT_FAILURE;
   }
-  
-  if (argc == 0 || options[HELP] || return_code == EXIT_FAILURE)
+
+  if ((options[MANIFESTADD] || options[RESADD]) && !options[BUNDLENAME])
   {
-    option::printUsage(std::clog, usage);
-    return return_code;
+      std::cerr << "if either --manifest-add or --res-add option is provided, --bundle-name option must be provided." << std::endl;
```
jeffdiclemente: nit-pick on grammar: Start sentence with capital letter and remove redundant word "option".

> ```If either --manifest-add or --res-add option is provided, --bundle-name must be provided.```

tools/rc/ResourceCompiler.cpp
```diff
@@ -411,36 +413,54 @@ int main(int argc, char** argv)
...
+      return_code = EXIT_FAILURE;
```
jeffdiclemente: indentation...

tools/rc/ResourceCompiler.cpp
jeffdiclemente: indentation...

tools/rc/ResourceCompiler.cpp
```diff
@@ -411,36 +413,54 @@ int main(int argc, char** argv)
...
+      std::cerr << arg << " appears multiple times in the arguments. Check usage." << std::endl;
+      return_code = EXIT_FAILURE;
+    }
+  };
+  check_multiple_args(BUNDLEFILE, "(--bundle-file | -b)");
+  check_multiple_args(OUTFILE, "(--out-file | -o)");
+  check_multiple_args(BUNDLENAME, "(--bundle-name | -n)");
+  check_multiple_args(MANIFESTADD, "(--manifest-add | -m)");
+
+  if (argc == 0 || options[HELP])
```
jeffdiclemente: The usage instructions used to be displayed when there was an invalid set of options, this appears to have changed. Any particular reason?

tools/rc/ResourceCompiler.cpp
```diff
@@ -411,36 +413,54 @@ int main(int argc, char** argv)
     }
     return_code = EXIT_FAILURE;
   }
-  
-  option::Option* bundleFileOpt = options[BUNDLEFILE];
-  if (bundleFileOpt && bundleFileOpt->count() > 1 )
+
+  // multiple specifications of the following args are illegal.
+  auto check_multiple_args = [&options, &return_code](OptionIndex optionidx, const std::string& arg)
```
jeffdiclemente: I believe this lambda can be further simplified. ```const std::string& arg``` isn't redundant since the same text can be obtained from the ```option::Option``` object passed into the capture list (i.e. ```&options```). 
You could even take this a step further and pass in a vector of ```OptionIndex``` values instead of invoking the lambda ```N``` times.

With all these checks for invalid options combinations, it may make sense to re-factor this code to consolidate all this logic in one place.

tools/rc/ResourceCompiler.cpp
ksubramz: Abhinay and I had a discussion about this. The consensus was we could change the behavior to display the usage information only if "--help" is provided. I felt that traditionally, command-line programs display only the errors. Of course, if you feel the other way, the majority wins so I'll revert back to displaying usage information for errors  :)

ksubramz:
The checks are passing and the changes are up for another batch of review. Appreciated :)

karthikreddy09:
Thanks for making the changes @ksubramz

Issue #83, opened 2016-03-01
Removing Listeners does not work well


karthikreddy09:
we use std::function objects for listeners. 
1. The listener has to be a function pointer. Anything else used does not work. e.g: lamdas
2. When the same service listener is added with different filters, the remove is ambigious

jeffdiclemente:
This is related to #95 as in solving one can also solve the other.

jeffdiclemente:
fixed in PR #179

Issue #95, opened 2016-05-11
Cannot add more than one listener if its expressed as a lambda


jeffdiclemente:
I noticed this while implementing the framework listeners; doing the following does not add the second listener:

```
int count1(0);
int count2(0);
auto listener_callback_counter1 = [&count1](const FrameworkEvent& evt) { ++count1; std::cout << "listener_callback_counter1: call count " << count1 << std::endl; };
auto listener_callback_counter2 = [&count2](const FrameworkEvent& evt) { ++count2; std::cout << "listener_callback_counter2: call count " << count2 << std::endl; };
f->GetBundleContext()->AddFrameworkListener(listener_callback_counter1);
f->GetBundleContext()->AddFrameworkListener(listener_callback_counter2);
```

The same happens when adding bundle listeners. 
In both cases the listener comparison functions in `usServicveListeners.cpp` returns true for the second listener as compared to the first. 

If the `std::bind` flavor of `AddBundleListener` is used, there is no problem.

For users, it is normal to express std::function objects as lambdas and expect this to work.

It is difficult to compare the equality of std::function objects as evidenced by [this stackoverflow thread](https://stackoverflow.com/questions/3629835/why-is-stdfunction-not-equality-comparable). 
We'll have to define what equality really means for listeners.

jeffdiclemente:
Some more info from further investigation...

Since ```std::function::target<void(const BundleEvent&)>()``` is used as one of the criteria to compare listeners for equality lambdas and callable objects created from ```std::bind``` will never be compared correctly because ```target<T>()``` will return a pointer only if the stored callable object is of type ```T```, otherwise it returns ```nullptr```. For lambdas and callable objects created from ```std::bind```, the type is not ```void(const BundleEvent& )```.

The only reason listeners created from ```std::bind``` can be added is because the pointer to the class instance is also used to compare the listeners. As long as two different class instance pointers are used the listeners aren't equal.

jeffdiclemente:
fixed in PR #179

PR #179, opened 2017-02-26
Support multiple listeners.

ksubramz:
Fix issues #83 and #95.

Change the BundleContext add listeners functions to return ListenerToken
objects.

Add API function RemoveListeners which can use these
ListenerToken objects to remove the listeners corresponding to the
tokens. Removing listeners using their tokens is the recommended
approach to clients going forward.

Introduce the move-only ListenerToken type. This is a wrapper to
ListenerTokenId, an uint64_t type, which is incremented every-time
a listener of any type is added to the BundleContext.

Deprecate existing API functions RemoveFrameworkListener,
RemoveBundleListener and RemoveServiceListener. These will be
removed in the next major release. These functions also emit
diagnostic log messages that they are deprecated.

Change framework code to use tokens in-place of the deprecated
RemoveServiceListener calls.

Change ServiceListenerEntry to contain ListenerTokenId.

Enable the previously commented out multiple framework listeners
test block in FrameworkListenerTest.

Add a new test file named MultipleListenersTest which exercises
this new functionality in a variety of ways.


framework/src/service/ServiceListeners.cpp
```diff
@@ -82,82 +93,170 @@ void ServiceListeners::Clear()
...
 }
 
+/**
+ * Called by RemoveListener(listenerToken)
+ * Iterates over the set of FrameworkListeners, BundleListeners and ServiceListeners to find the tokenId and
+ * remove the listener corresponding to the tokenId.
+ */
+void ServiceListeners::RemoveListener(const std::shared_ptr<BundleContextPrivate>& context, const ListenerToken& token)
+{
+  {
```
saschazelzer: We should check for an invalid token first and return early.

saschazelzer: Having to go through all the maps feels wasteful to me. At least, we should check the service listener map first because service listeners are more likely to be added / removed during the life time of a program.

If it is not too much work, it would be interesting to compare the performance implication when having lots of listeners registered (of different type) and we want to remove a single one. Comparison would be between 3.0 and your branch. I would say lets have 100 framework listeners, 1000 bundle listeners, and 10.000 service listeners. Then add and remove an additional one of each type in a loop (e.g. 1000 times) in a dedicated thread which is created 10 times (they should start looping only when all 10 threads have been started) and measure the average time for the addition and removal. Be sure to use a release build.

If there is a big performance hit, we might re-consider using a bit flag in the token id to store the listener type. In case you are too busy, I could help  out with this kind of testing.

framework/src/service/ServiceListeners.cpp
```diff
@@ -82,82 +93,171 @@ void ServiceListeners::Clear()
...
+    RemoveServiceListenerEntry(it, sle);
   }
   if (!sle.IsNull())
   {
     coreCtx->serviceHooks.HandleServiceListenerUnreg(sle);
   }
 }
 
-void ServiceListeners::AddBundleListener(const std::shared_ptr<BundleContextPrivate>& context, const BundleListener& listener, void* data)
+void ServiceListeners::RemoveServiceListener(const std::shared_ptr<BundleContextPrivate>& context,
```
jeffdiclemente: The only difference between the two ```RemoveServiceListener``` functions are the token ids. Is there a way to remove the code duplication?

framework/test/MultipleListenersTest.cpp
```diff
@@ -0,0 +1,322 @@
...
+    fCtx.RemoveListener(token2);
+    fCtx.RemoveListener(token3);
+    fCtx.RemoveListener(token4);
+    fCtx.RemoveListener(token5);
+    fCtx.RemoveListener(token6);
+    fCtx.RemoveListener(token7);
+    fCtx.RemoveListener(token8);
+    fCtx.RemoveListener(token9);
+    fCtx.RemoveListener(token10);
+    fCtx.RemoveListener(token11);
```
karthikreddy09: same as the previous comment

framework/src/service/ServiceListeners.cpp
```diff
@@ -38,33 +38,44 @@ US_MSVC_PUSH_DISABLE_WARNING(4180) // qualifier applied to function type has no
...
 
-struct BundleListenerCompare : std::binary_function<std::pair<BundleListener, void*>,
-                                                    std::pair<BundleListener, void*>, bool>
+/**
+ * Predicate functions used for comparison when adding and removing listeners.
+ * The functions ending in "Name" are legacy functions to support backwards compatibility.
+ * These will be removed when we implement functionality to remove listeners only via
+ * their listener tokens.
+ */
+auto BundleListenerCompareName = [](const BundleListener& l1, void* d1,
```
karthikreddy09: I think the function name is misleading. It is comparing function addresses not names.

framework/test/MultipleListenersTest.cpp
```diff
@@ -0,0 +1,322 @@
...
+    auto add_listener = [&fCtx, &count]()
+    {
+      auto listener = [&count](const FrameworkEvent&){ ++count; };
+      auto token = fCtx.AddFrameworkListener(listener);
+      return token;
+    };
+
+    // 1. Add num_listeners number of listeners asynchronously.
+    // 2. Gather the resulting tokens.
+    // 3. Remove the first remove_count number of listeners using the tokens.
```
jeffdiclemente: This comment is not very helpful. Its describing what can be easily observed in the code. 
What would be most helpful to maintainers is to explain why this is being done.

framework/src/service/ServiceListeners.cpp
ksubramz: Great catch. (silver lining is we now know that passing invalid token works as intended even if it's not caught early :) )

framework/include/cppmicroservices/BundleContext.h
```diff
@@ -668,11 +826,11 @@ class US_Framework_EXPORT BundleContext
    * @see RemoveServiceListener()
    */
   template<class R>
-  void AddServiceListener(R* receiver, void(R::*callback)(const ServiceEvent&),
-                          const std::string& filter = std::string())
+  ListenerToken AddServiceListener(R* receiver, void(R::*callback)(const ServiceEvent&),
```
ksubramz: Agreed, this makes it a single interface to add listeners for the clients. 

Note that as per the design document, currently, this is the only instance where if an attempt is made to add a listener (with or without a different filter) of this type to the system, it'll replace the existing listener in the system. 

The motivation was that the existing client code functionality should work as intended, unless it's an obvious bug which is getting fixed in this PR. Not doing this would mean a lot of tests such as ServiceHooksTest would have to change. Please refer to the design doc for more information.

framework/src/service/ServiceListeners.cpp
ksubramz: Your comment about changing the Bundle and Framework listeners to be an `unordered_map` with the `ListenerTokenId`s as the keys sent me in the right path. With this, the amortized removal when using tokens to remove is `O(1)`. Therefore, I reckon we don't even need the bit flag in the token id - because the information that the token id not present in the list of bundle or framework listeners is got after `O(1)`

I am now figuring out if we can follow similar approach for the ServiceListeners.

framework/src/service/ServiceListeners.h
```diff
@@ -50,17 +51,24 @@ class ServiceListeners : private detail::MultiThreaded<>
 
 public:
 
-  typedef std::unordered_map<std::shared_ptr<BundleContextPrivate>, std::list<std::pair<BundleListener,void*> > > BundleListenerMap;
+  typedef std::tuple<BundleListener, void*, ListenerTokenId> BundleListenerEntry;
+  typedef std::unordered_map<std::shared_ptr<BundleContextPrivate>, std::list<BundleListenerEntry>> BundleListenerMap;
```
ksubramz: Great suggestion.

framework/src/service/ServiceListeners.cpp
saschazelzer: Amortized constant time is great in theory, but if that constant is large, the price can still be non-negligible. But unless we measure and compare it, we cannot tell.

ksubramz: Sounds good...

framework/include/cppmicroservices/BundleContext.h
```diff
@@ -607,15 +609,169 @@ class US_Framework_EXPORT BundleContext
...
+   * value <code>1</code>. Subsequently, when the service is modified
+   * setting property <code>x</code> to the value <code>1</code>, the
+   * filter will match and the callable will be called with a
+   * <code>ServiceEvent</code> of type <code>SERVICE_MODIFIED</code>. Thus, the
+   * callable will not be called with a <code>ServiceEvent</code> of type
+   * <code>SERVICE_REGISTERED</code>.
+   *
+   * @param callable Any type of callable.
+   * @param filter The filter criteria.
+   * @returns a ListenerToken object which can be used to remove the callable from the
```
ksubramz: Jeff, it looks like "callable object" is the best word to describe the type of listeners which can be passed to the function. So, it makes more sense not to name the function parameter as "callable" and I'll name it something appropriate and less confusing.

framework/include/cppmicroservices/BundleContext.h
ksubramz: Tagging @jeffdiclemente @karthikreddy09 @saschazelzer 

I am having second thoughts about deprecating these three Add.*Listener functions (that take member functions). Reasons:

1. Though the functionality can be replicated by the other three Add.*Listener functions, these three seem to be convenient wrappers for adding member functions.
2. Internally, I see the usage of these functions to be more prevalent. If deprecated, all the invocations have to be changed to std::bind form, that maybe inconvenient to clients i.e. If these are deprecated, the negatives outweigh the benefits. Do you feel about this strongly, Sascha and others?

saschazelzer: IMHO, having only one way to add listeners is preferable. I didn't come across any other (more or less well-known) library which would expose a second API function for convenience member function registration. `std::bind` is a common and accepted technique from what I have seen.

ksubramz:
Sounds good!

framework/include/cppmicroservices/ListenerFunctors.h
```diff
@@ -97,18 +151,56 @@ namespace cppmicroservices {
...
+   * an object of type <code>R</code> and returns a <code>ServiceListener</code> object.
+   * This object can then be passed into <code>AddServiceListener()</code>.
+   *
+   * @tparam R The type containing the member function.
+   * @param receiver The object of type R.
+   * @param callback The member function pointer.
+   * @returns a ServiceListener object.
+   */
+  template<class R>
+  ServiceListener ServiceListenerMemberFunctor(R* receiver, void (R::*callback)(const ServiceEvent&))
```
saschazelzer: Shouldn't we deprecate this as well?

framework/include/cppmicroservices/ListenerFunctors.h
```diff
@@ -97,18 +151,56 @@ namespace cppmicroservices {
...
+   * an object of type <code>R</code> and returns a <code>BundleListener</code> object.
+   * This object can then be passed into <code>AddBundleListener()</code>.
+   *
+   * @tparam R The type containing the member function.
+   * @param receiver The object of type R.
+   * @param callback The member function pointer.
+   * @returns a BundleListener object.
+   */
+  template<class R>
+  BundleListener BundleListenerMemberFunctor(R* receiver, void (R::*callback)(const BundleEvent&))
```
saschazelzer: Deprecate?

framework/include/cppmicroservices/ListenerFunctors.h
```diff
@@ -97,18 +151,56 @@ namespace cppmicroservices {
...
+   * an object of type <code>R</code> and returns a <code>FrameworkListener</code> object.
+   * This object can then be passed into <code>AddFrameworkListener()</code>.
+   *
+   * @tparam R The type containing the member function.
+   * @param receiver The object of type R.
+   * @param callback The member function pointer.
+   * @returns a FrameworkListener object.
+   */
+  template<class R>
+  FrameworkListener BindFrameworkListenerToFunctor(R* receiver, void (R::*callback)(const FrameworkEvent&))
```
saschazelzer: Deprecate?

framework/src/service/ServiceListeners.cpp
```diff
@@ -82,32 +84,72 @@ void ServiceListeners::Clear()
   frameworkListenerMap.Lock(), frameworkListenerMap.value.clear();
 }
 
-void ServiceListeners::AddServiceListener(const std::shared_ptr<BundleContextPrivate>& context, const ServiceListener& listener,
-                                          void* data, const std::string& filter)
+ListenerToken ServiceListeners::MakeListenerToken()
+{
+  auto l = this->Lock(); US_UNUSED(l);
```
saschazelzer: The lock is unnecessary, please remove.

framework/test/CMakeLists.txt
```diff
@@ -100,6 +102,25 @@ add_executable(${_test_driver} ${_srcs} ${_additional_srcs} ${_third_party_srcs}
 add_dependencies(${_test_driver} ${US_RCC_EXECUTABLE_TARGET})
 set_property(TARGET ${_test_driver} APPEND PROPERTY COMPILE_DEFINITIONS US_BUNDLE_NAME=main)
 set_property(TARGET ${_test_driver} PROPERTY US_BUNDLE_NAME main)
+
+# Disable deprecation warnings in the test code.
+if (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") OR
+    ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang") OR
+    ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU"))
+  #target_compile_optinons(${_test_driver} PUBLIC "-Wno-deprecated-declarations")
+  set_source_files_properties(BundleActivatorTest.cpp PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
```
saschazelzer: Can't we just do this
```
set_source_files_properties(${_srcs} PROPERTIES COMPILE_FLAGS -Wno-error=deprecated-declarations)
```
instead of the repetitive lines and the complete disablement of the warning?

framework/include/cppmicroservices/ListenerFunctors.h
ksubramz: Sure. I had two choices - either deprecate these or add documentation to these convenience functions. But, std::bind is straightforward enough that clients wouldn't need these convenient functions I guess.

framework/include/cppmicroservices/ListenerFunctors.h
```diff
@@ -151,55 +151,16 @@ namespace cppmicroservices {
...
-   * This object can then be passed into <code>AddServiceListener()</code>.
-   *
-   * @tparam R The type containing the member function.
-   * @param receiver The object of type R.
-   * @param callback The member function pointer.
-   * @returns a ServiceListener object.
-   */
   template<class R>
-  ServiceListener ServiceListenerMemberFunctor(R* receiver, void (R::*callback)(const ServiceEvent&))
+  US_DEPRECATED ServiceListener ServiceListenerMemberFunctor(R* receiver, void (R::*callback)(const ServiceEvent&))
```
saschazelzer: Thanks, but can we get the documentation back :-) You invested some time in it already and it would be good to have it. Just add a short "deprecated" note there as well.

framework/test/CMakeLists.txt
```diff
@@ -107,18 +107,7 @@ set_property(TARGET ${_test_driver} PROPERTY US_BUNDLE_NAME main)
...
-  set_source_files_properties(BundleTest.cpp PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
-  set_source_files_properties(ConcurrencyTest.cpp PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
-  set_source_files_properties(FrameworkListenerTest.cpp PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
-  set_source_files_properties(FrameworkTest.cpp PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
-  set_source_files_properties(MultipleListenersTest.cpp PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
-  set_source_files_properties(ServiceListenerTest.cpp PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
-  set_source_files_properties(ServiceHooksTest.cpp PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
-  set_source_files_properties(ServiceRegistryPerformanceTest.cpp PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
-  set_source_files_properties(StaticBundleTest PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
+  set_source_files_properties(${_srcs} PROPERTIES COMPILE_FLAGS -Wno-deprecated-declarations)
```
saschazelzer: Did `-Wno-error=deprecated-declarations` not work, or do think suppressing the warnings completely is preferable?

framework/src/service/ServiceListeners.h
```diff
@@ -190,7 +210,14 @@ class ServiceListeners : private detail::MultiThreaded<>
 
 private:
 
-  void RemoveServiceListener(const ServiceListenerEntry& entryToRemove);
+  /**
+   * Factory method for return an unique ListenerToken object.
```
jeffdiclemente: Grammar: 
"Factory method that returns an unique ListenerToken object."

framework/src/service/ServiceListeners.cpp
```diff
@@ -24,6 +24,8 @@
 
 US_MSVC_PUSH_DISABLE_WARNING(4180) // qualifier applied to function type has no meaning; ignored
 
+#include <cassert>
```
ksubramz: I did use asserts in the interim but not anymore. Didn't remove the needless #include though.

framework/include/cppmicroservices/ListenerFunctors.h
```diff
@@ -24,15 +24,69 @@
...
+
+  /**
+   * \brief The token returned when a listener is registered with BundleContext.
+   *
+   * The token object enables the clients to remove the listeners from the BundleContext.
+   * This is a move-only type, with the intention that the transfer of ownership will
+   * be handled explicitly by the clients.
+   *
+   */
+  class US_Framework_EXPORT ListenerToken
```
ksubramz: The way I tackled both this and the suggestion below is to 
- separate out the ListenerToken declaration into ListenerToken.h and export it
- rename ListenerFunctors.cpp to ListenerToken.cpp (where the definition of ListenerToken resides)

framework/src/service/ServiceListeners.cpp
```diff
@@ -122,48 +155,94 @@ void ServiceListeners::RemoveServiceListener(const ServiceListenerEntry& entryTo
...
+void ServiceListeners::RemoveListener(const std::shared_ptr<BundleContextPrivate>& context, ListenerToken token)
+{
+  if (!token)
+  {
+    return;
+  }
+
+  auto tokenId = token.Id();
+  {
+    auto l = frameworkListenerMap.Lock(); US_UNUSED(l);
```
karthikreddy09: This block and the folowing block have the same functionality. can we create one function/lambda and call it twice instead of duplicating code with different variables.

framework/src/service/ServiceListeners.cpp
```diff
@@ -122,48 +127,103 @@ void ServiceListeners::RemoveServiceListener(const ServiceListenerEntry& entryTo
...
+      listener.target<void(const BundleEvent&)>() == std::get<0>(pair.second).target<void(const BundleEvent&)>();
+  };
+
   auto l = bundleListenerMap.Lock(); US_UNUSED(l);
-  bundleListenerMap.value[context].remove_if(std::bind(BundleListenerCompare(), std::make_pair(listener, data), std::placeholders::_1));
+  auto& listeners = bundleListenerMap.value[context];
+  auto it = std::find_if(listeners.begin(), listeners.end(), std::bind(BundleListenerCompareListenerData, listener, data, std::placeholders::_1));
+  if (it != listeners.end())
+  {
+      listeners.erase(it);
```
jeffdiclemente: spacing

framework/include/cppmicroservices/ListenerToken.h
```diff
@@ -32,12 +32,10 @@ namespace cppmicroservices {
   typedef std::uint64_t ListenerTokenId;
 
   /**
-   * \brief The token returned when a listener is registered with BundleContext.
-   *
-   * The token object enables the clients to remove the listeners from the BundleContext.
-   * This is a move-only type, with the intention that the transfer of ownership will
-   * be handled explicitly by the clients.
+   * \brief The token returned when a listener is registered with framework.
```
jeffdiclemente: should be "with the framework"

framework/include/cppmicroservices/ListenerToken.h
```diff
@@ -32,12 +32,10 @@ namespace cppmicroservices {
...
    *
+   * The token object is a move-only type and enables the clients to remove
```
jeffdiclemente: should be "enables clients to remove listeners from the framework"

saschazelzer:
This is great, thanks! One last thing, sorry for that: Please add / amend the final commit with a signed-off field. After that, it will be merged immediately.

PR #189, opened 2017-03-10
Bug fixes for tests and error handling

jeffdiclemente:
In our build environment (i.e. not travis ci), using a hard coded ```/tmp``` during tests occasionally produced a "permission denied" error while creating the persistent framework storage area. This lead to sporadic build failures. Changes were made to use ```TMPDIR``` instead which should always be a write-able location. ```/tmp``` was left in to satisfy travis ci. 

While making the above change, it was noticed that ```strerror``` in ```Utils.cpp``` could return a ```nullptr```. Fixed this as well to avoid constructing a ```std::string``` with a ```nullptr```.

### DIFF
--- a/framework/src/util/Utils.cpp
+++ b/framework/src/util/Utils.cpp
@@ -392,7 +392,8 @@ int GetLastErrorNo()
 std::string GetLastErrorStr()
 {
 #ifdef US_PLATFORM_POSIX
-  return std::string(strerror(errno));
+  char* errorString = strerror(errno);
+  return std::string(((errorString == nullptr)?"":errorString));
 #else
   // Retrieve the system error message for the last-error code
   LPVOID lpMsgBuf;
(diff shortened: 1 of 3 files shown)


framework/test/TestUtils.cpp
```diff
@@ -163,7 +163,8 @@ std::string GetTempDirectory()
   if (buffer) CoTaskMemFree(static_cast<void*>(buffer));
   return std::string(temp_dir.cbegin(), temp_dir.cend());
 #else
-  return std::string("/tmp");
+  char* tempdir = getenv("TMPDIR");
+  return std::string(((tempdir == nullptr)?"/tmp":tempdir));
```
kevinleeMW: Nice!

framework/src/util/Utils.cpp
```diff
@@ -392,7 +392,8 @@ int GetLastErrorNo()
 std::string GetLastErrorStr()
 {
 #ifdef US_PLATFORM_POSIX
-  return std::string(strerror(errno));
+  char* errorString = strerror(errno);
```
saschazelzer: I just realized that we are using the thread-unsafe `strerror` function here, but `GetLastErrorStr()` may be called concurrently. I will open a separate issue for this.

jeffdiclemente: Yep, I saw this as well. Thanks for opening the issue. 
The thread-safe versions of strerror are not part of standard POSIX I believe. To support MacOS, we'll have to double check which versions of thread-safe strerror are supported.

PR #190, opened 2017-03-10
travisci xcode8

jeffdiclemente:
add support for xcode8.2 in travis ci and upgrade the minimum supported xcode and Mac OS versions since Apple has released macOS Sierra.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,9 +4,9 @@
 +-------------+-------------------------+--------------------------+
 | Branch      | GCC 4.6                 | Visual Studio 2013       |
 |             +-------------------------+--------------------------+
-|             | XCode 6.4               | Visual Studio 2015       |
+|             | XCode 7.3               | Visual Studio 2015       |
 |             +-------------------------+--------------------------+
-|             | XCode 7.3               | MinGW-w64                |
+|             | XCode 8.2               | MinGW-w64                |
 +=============+=========================+==========================+
 | master      | |Linux Build Status|    | |Windows Build status|   |
 +-------------+-------------------------+--------------------------+
@@ -56,7 +56,7 @@ Minimum required compiler versions:
 
 - GCC 4.6
 - Clang 3.1
-- Clang from XCode 6.4
+- Clang from XCode 7.3
 - Visual Studio 2013
 
 Below is a list of tested compiler/OS combinations:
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -4,9 +4,9 @@
 +-------------+-------------------------+--------------------------+
 | Branch      | GCC 4.6                 | Visual Studio 2013       |
 |             +-------------------------+--------------------------+
-|             | XCode 6.4               | Visual Studio 2015       |
```
kevinleeMW: Can we change all instances of "XCode" to "Xcode" in this file?

mrlegowatch: Sara would be very happy about this!

saschazelzer:
Adding support for Xcode 8.2 is great. I am unsure about modifying minimum supported compiler versions in patch releases, if this PR is supposed to be part of the upcoming 3.1 release. Is there a reason why we couldn't still use the xcode6.4 image from travis?

jeffdiclemente:
@saschazelzer yes, this is supposed to be for 3.1. There is no reason we couldn't still use xcode 6.4. I moved ahead since Apple moves Xcode and their SDKs ahead frequently and it helps to stay current.

I can add xcode 6.4 back and keep it as the minimum supported macos compiler. Is that alright?

saschazelzer:
I have not much experience with Xcode versions and Apples support policies. If Xcode 6.4 is still officially supported than I would say yes, please add it back. Otherwise, we should update (if required) to the oldest still supported version if possible.

jeffdiclemente:
The more important part is to stay up to date with Xcode and the current SDK (i.e. Xcode 8.2 and macOS Sierra 10.12).

Xcode 6.4 is no longer supported starting with Mac OS X 10.11 (El Capitan). I suggest we remove Xcode 6.4 for the CppMicroServices 3.1 release. What do you think?

My understanding is that Apple supports the current Xcode version shipped with the current macOS and its SDK, Apple typically supports the previous Xcode and SDK version as well. So Apple is currently supporting Xcode 8.2 on 10.12 (macOS Sierra) and 7.4 on 10.11 (El Capitan). Apple moves at a quick cadence.

I agree with not changing the minimum compiler support for patch releases.
Is it acceptable to change it for a minor point release (e.g. for version 3.1)? What's the procedure in this case?

saschazelzer:
I guess we are defining the policy right now. In my view, we should strive to not change minimum compiler versions in between minor releases except if they are not officially supported any more *and* the code would need modifications in order to still support that old version. If it doesn't cost us anything keeping the current minimum version, why change it?

jeffdiclemente:
That seems reasonable to me. I'll put Xcode 6.4 back.

jeffdiclemente:
If there are no objections, I will merge this into development later today.

PR #195, opened 2017-03-20
Improve BadAnyCastException message

### LINKED ISSUE #181: Exception messages for BadAnyCastException are unhelpful
The messages returned from BadAnyCastException::what() are unhelpful while debugging failures. The message appears to be just a static string:

> Failed to convert between const Any types

any_cast has access both to the actual type contained within the Any as well as the C++ type that was the requested target type. The message should contain both of these in order to help developers troubleshoot failed casts.

jeffdiclemente:
Fixes #181
 
### DIFF
--- a/framework/include/cppmicroservices/Any.h
+++ b/framework/include/cppmicroservices/Any.h
@@ -455,11 +455,7 @@ private:
 
 namespace detail
 {
-US_Framework_EXPORT std::string GetDemangledName(const std::type_info& typeInfo);
-}
-
 /**
- * \ingroup gr_any
  *
  * A utility function used to throw a BadAnyCastException object
  * containing an exception message containing the source and target type names.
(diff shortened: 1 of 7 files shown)


framework/test/AnyTest.cpp
```diff
@@ -149,5 +149,44 @@ int AnyTest(int /*argc*/, char* /*argv*/[])
   US_TEST_CONDITION(anyMap3.ToString() == "{map : {1 : 0.3, 3 : bye}, number : 5, vector : [9,8,7]}", "Any[std::map<std::string,Any>].ToString()")
   US_TEST_CONDITION(anyMap3.ToJSON() == "{\"map\" : {\"1\" : 0.3, \"3\" : \"bye\"}, \"number\" : 5, \"vector\" : [9,8,7]}", "Any[std::map<std::string,Any>].ToJSON()")
 
+  // Test BadAnyCastException exceptions
+  const Any uncastableConstAny(0.0);
+  Any uncastableAny(0.0);
+
+  US_TEST_FOR_EXCEPTION(cppmicroservices::BadAnyCastException, any_cast<std::string>(uncastableConstAny))
+  
+  try
```
jeffdiclemente: This capability could be added. I don't know if there is any real value in doing it with gtest unit tests coming soon.

framework/include/cppmicroservices/Any.h
```diff
@@ -452,6 +452,31 @@ class BadAnyCastException : public std::bad_cast
...
+ *
+ * A utility function used to throw a BadAnyCastException object
+ * containing an exception message containing the source and target type names.
+ *
+ * \param funcName The throwing function's name.
+ * \param anyTypeName A string representing the Any object's underlying type.
+ * \throws cppmicroservices::BadAnyCastException
+ */
+template <typename TargetType>
+void ThrowBadAnyCastException(const std::string& funcName, const std::string& anyTypeName)
```
saschazelzer: I think this should go into the `detail` namespace as well and not be part of the public API docs. Passing both the target and source std::type_info objects as arguments to `ThrowBadAnyCastException` would also allow us to hide its implementation in a cpp file as well as not having to export `GetDemangledName`.

jeffdiclemente: I'll look into doing that. Thanks!

PR #198, opened 2017-03-29
Fix for unsafe_any_cast

karthikreddy09:
Signed-off-by: Mathworks Inc <Roy.Lurie@mathworks.com>

### DIFF
--- a/framework/test/AnyTest.cpp
+++ b/framework/test/AnyTest.cpp
@@ -29,6 +29,14 @@
 
 using namespace cppmicroservices;
 
+template <typename T>
+void TestUnsafeAnyCast(Any& anyObj, T val)
+{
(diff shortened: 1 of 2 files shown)


framework/test/AnyTest.cpp
```diff
@@ -29,6 +29,14 @@
 
 using namespace cppmicroservices;
 
+template <typename T>
+void TestUnsafeAnyCast(Any& anyObj, T val)
+{
```
ksubramz: Maybe assign the unsafe cast to a local ptr?
```cpp
T* ptr = unsafe_any_cast<T>(&anyObj);
US_TEST_CONDITION(ptr, "unsafe_any_cast");
```
and similarly for other 2 `US_TEST_CONDITION`s

PR #206, opened 2017-05-16
Support non standard shared library extensions

### LINKED ISSUE #205: support bundles that do not have .DLL/.so/.dylib extensions
There is a use case to support shared libraries which don't end in the traditional shared library file extension (i.e. DLL, DYLIB, SO). 

The simplest solution would be to remove the check within CppMicroServices when installing a bundle and loading the shared library.

jeffdiclemente:
Fixes #205


framework/src/bundle/BundlePrivate.cpp
```diff
@@ -405,7 +405,7 @@ void BundlePrivate::FinalizeActivation(LockType& l)
 
 void BundlePrivate::Uninstall()
 {
-  if (!IsSharedLibrary(GetLocation()))
+  if (BundleUtils::GetExecutablePath() == GetLocation())
```
jeffdiclemente: While code reviewing with Krish, I realized that either this throw is no longer needed OR the condition needs to also check for statically linked bundles inside of shared libraries.

Thinking about this, I think we should no longer throw since we removed the throw if a client installs a separate EXE as a bundle. This would provide symmetrical behavior between installing and uninstalling bundles (i.e. they don't throw).

jeffdiclemente:
Thanks for the code reviews. 
I will merge this later today, once I've addressed my comment in BundlePrivate.cpp.

PR #207, opened 2017-05-23
204 unicode filepath support

karthikreddy09:


third_party/miniz.c
```diff
@@ -2850,17 +2850,44 @@ void *tdefl_write_image_to_png_file_in_memory(const void *pImage, int w, int h,
...
   #if defined(_MSC_VER) || defined(__MINGW64__)
+    #include <Windows.h>
+    #include <stringapiset.h>
+    #include <wchar.h>
+    static wchar_t* utf8_to_wchar(const char* inStr)
+    {
+      if (inStr == NULL)
+        return NULL;
+      int wchars_count = MultiByteToWideChar(CP_UTF8, 0, inStr, -1, NULL, 0);
+      wchar_t* wstr = malloc(sizeof(wchar_t)*wchars_count);
```
jeffdiclemente: If ```MultiByteToWideChar``` fails, this will allocate a 0 byte buffer.

third_party/miniz.c
```diff
@@ -2850,17 +2850,44 @@ void *tdefl_write_image_to_png_file_in_memory(const void *pImage, int w, int h,
...
       FILE* pFile = NULL;
-      if (freopen_s(&pFile, pPath, pMode, pStream))
-        return NULL;
+      wchar_t* pFilenameW = utf8_to_wchar(pPath);
+      wchar_t* pModeW = utf8_to_wchar(pMode);
+      if (_wfreopen_s(&pFile, pFilenameW, pModeW, pStream))
+      {
+        pFile = NULL;
+      }
+      free(pFilenameW);
```
jeffdiclemente: ```utf8_to_wchar``` can return a nullptr. nullptr checks should happen before executing ```free```

framework/src/util/Utils.cpp
```diff
@@ -90,8 +110,28 @@ std::string library_suffix()
 
 }
 
+
 namespace cppmicroservices {
+  //-------------------------------------------------------------------
+  // Unicode Utility functions
+  //-------------------------------------------------------------------
+#ifdef US_PLATFORM_WINDOWS
+  std::wstring ToWString(const std::string& inStr)
```
jeffdiclemente: what's the expected behavior of ```ToWString``` and ```ToUTF8String``` if ```MultiByteToWideChar``` and ```WideCharToMultiByte``` return ```0```?

framework/src/util/Utils.cpp
```diff
@@ -90,8 +110,28 @@ std::string library_suffix()
...
+  {
+    int wchars_count = MultiByteToWideChar(CP_UTF8, 0, inStr.c_str(), -1, NULL, 0);
+    std::unique_ptr<wchar_t[]> wBuf(new wchar_t[wchars_count]);
+    MultiByteToWideChar(CP_UTF8, 0, inStr.c_str(), -1, wBuf.get(), wchars_count);
+    return wBuf.get();
```
ksubramz: Wouldn't this result in `use-after-free`? My concern is if because the unique_ptr goes out of scope, wouldn't the memory be reclaimed?

(if it's concerning, maybe use std::vector in the intermediate and return a std::wstring off that? something like)
```cpp
 std::wstring ToWString(const std::string& inStr)
 {
     int wchars_count = MultiByteToWideChar(CP_UTF8, 0, inStr.c_str(), -1, NULL, 0);
     std::vector<wchar_t> vBuf(wchars_count);
     MultiByteToWideChar(CP_UTF8, 0, inStr.c_str(), -1, vBuf.data(), wchars_count);
     return std::wstring(vBuf.begin(), vBuf.end());
}
```

third_party/miniz.c
karthikreddy09: if null is passed to free, it's a no-op. 
http://www.cplusplus.com/reference/cstdlib/free/

third_party/miniz.c
karthikreddy09: good catch.

framework/src/util/Utils.cpp
karthikreddy09: The returned object is copied before the std::unique_ptr is destroyed.
https://stackoverflow.com/questions/275214/scope-and-return-values-in-c

Thanks for the vector suggestion! :-)

framework/test/BundleTest.cpp
```diff
@@ -609,12 +610,46 @@ void TestAutoInstallEmbeddedBundles()
...
+    std::string path_utf8 = testing::LIB_PATH + testing::DIR_SEP + u8"くいりのまちとこしくそ" + testing::DIR_SEP + US_LIB_PREFIX + "TestBundleU" + US_LIB_EXT;
+    auto bundles = frameworkCtx.InstallBundles(path_utf8);
+    US_TEST_CONDITION(bundles.size() == 1, "Install bundle from unicode path");
+    auto bundle = bundles.at(0);
+    US_TEST_CONDITION(bundle.GetLocation() == path_utf8, "Bundle location is the same as the path used to install");
+    US_TEST_NO_EXCEPTION(bundle.Start());
+    US_TEST_CONDITION(bundle.GetState() == Bundle::State::STATE_ACTIVE, "Bundle check start state");
+    US_TEST_NO_EXCEPTION(bundle.Stop());
+  }
+  catch (const std::exception& ex)
```
karthikreddy09: There is already other test point to test invalid bundle paths. Not sure how much value is added with an invalid unicode path. I will wait for others to chime in.

framework/test/bundles/libU/resources/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+  "bundle.symbolic_name" : "TestBundleU",
+  "bundle.activator" : true,
+  "unicode.sample" : "电脑 くいりのまちとこしくそ"
```
saschazelzer: The property does not seem to be used. It would be nice to extend the unit tests and compare the manifest header with an embedded `u8"..."` string literal.

framework/test/bundles/libU/TestBundleU.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+struct TestBundleU : public TestBundleUService
+{
+
+  TestBundleU() {}
+  virtual ~TestBundleU() {}
+
+};
+
+class TestBundleUActivator : public BundleActivator
```
saschazelzer: The test bundle does not seem to need services, so those files and the related code can be removed. It even wouldn't need an activator, but you need some code to generate a bundle library...

framework/src/util/Utils.cpp
```diff
@@ -90,8 +110,35 @@ std::string library_suffix()
...
+  //-------------------------------------------------------------------
+  // Unicode Utility functions
+  //-------------------------------------------------------------------
+#ifdef US_PLATFORM_WINDOWS
+
+  // function returns empty string if inStr is empty or if the conversion failed
+  std::wstring ToWString(const std::string& inStr)
+  {
+    int wchar_count = MultiByteToWideChar(CP_UTF8, 0, inStr.c_str(), -1, NULL, 0);
+    std::unique_ptr<wchar_t[]> wBuf(new wchar_t[wchar_count+1]);
```
karthikreddy09: Yes, the extra character is to accomodate the case when wchar_count == 0 (when the api call fails). When the API succeeds we set the extra character at the end with '\0' so the string ends with "\0\0". I was trying to keep the code simple (without another if clause or throwing an exception) at the cost of one character per function call. 

I like the vector solution Krish suggested, however, that solution failed in MinGW with SEG_FAULT. If throwing an exception makes sense, the extra character can be removed.

framework/test/bundles/libU/resources/manifest.json
karthikreddy09: sure, makes sense!

framework/src/util/Utils.cpp
saschazelzer: Either the current solution or using a vector instead is fine with me.

But I think we should really differentiate the "invalid utf8" and "empty string" case. Those are too different to be handled in the same way.

framework/src/util/Utils.cpp
```diff
@@ -164,12 +213,8 @@ bool Exists(const std::string& path)
     else throw std::invalid_argument(GetLastErrorStr());
   }
 #else
-  DWORD attr(::GetFileAttributes(path.c_str()));
```
karthikreddy09: The reason for change is PathFileExistsW supports UNC paths and GetFileAttributes does not. I was unaware of the performance difference. This particular change is not really needed to support unicode so I am reverting it.

jeffdiclemente:
Abhinay is out all this week and unable to make any additional changes.
I suggest we push this change into 3.2 release instead of trying to force it for 3.1.
Thoughts?

saschazelzer:
Agreed, no need to force it for 3.1.

karthikreddy09:
I started working on this again. Should have a new pull request soon.

karthikreddy09:
Merged PR#245. Discarding this one.

Issue #213, opened 2017-06-09
The Framework must not recursively call ServiceFactory::GetService and ServiceFactory::UngetService on the same thread for the same bundle


jeffdiclemente:
According to the OSGi spec for a ServiceFactory, the ```getService(Bundle bundle,ServiceRegistration<S> registration)``` method states:

>  If this method is recursively called for the specified bundle, a
framework event of type FrameworkEvent.ERROR is fired containing a service exception of type
ServiceException.FACTORY_RECURSION and null is returned to the bundle.

The Framework currently doesn't enforce this. This issue is to track the implementation of this constraint.

jeffdiclemente:
Sascha described in #202 how this behavior could happen:

> When the framework calls GetService, it effectively calls into external code and the service factory implementation could (in theory) call GetService or UngetService on the same factory instance again (either explicitly or implicitly via calling back into the framwork API). This needs to be detected and handled.


Test code needs to be written to trigger this behavior and verify the implementation fixes the issue.

saschazelzer:
I am reading paragraph 10.1.25.1 of the OSGi Core 6.0 specs about service factories. The specified error handling is not in line with our code in `ServiceReferenceBasePrivate::GetServiceFromFactory` and I am thinking about using the correct `FrameworkEvent::Type` enum (error instead of warning) and using our existing `ServiceException` class in the created framework event. Do you disagree?

jeffdiclemente:
That all sounds good to me. I noticed that code coverage for the ```ServiceException``` is low and this is a great opportunity to increase it. 

It does beg the question on whether we should implement the ```BundleException``` class and use it consistently.

Also, I think that the work to convert over to consistently using ```ServiceException``` should be a separate issue independent of this one.

saschazelzer:
Good points. I think we should implement BundleException and use it consistently, ideally in 3.x in a backwards compatible manner. I am going to create new issues.

jeffdiclemente:
I agree on using BundleExceptions consistently and in a backwards compatible way for 3.x. 

Thanks!

saschazelzer:
The implementation in #222 makes use of a thread-local variable. Segmentation faults of tests from builds based on MingGW revealed a problem with the TLS implementation in libwinpthreads, which is also noted here:

https://github.com/Alexpux/MINGW-packages/issues/2519
https://sourceforge.net/p/mingw-w64/bugs/527/

[...]

PR #214, opened 2017-06-13
Fix for ServiceReferenceBasePrivate race condition

### LINKED ISSUE #202: A race condition in ServiceReferenceBasePrivate causes a seg fault
A race can occur when a client retrieves a service from a ```ServiceFactory``` using ```BundleContext::GetService(ref)```. The race leads to a seg fault which occurs as a result of trying to access a ```nullptr``` service object returned from an empty ```std::unordered_map``` of cached service objects, maintained by CppMicroServices.

The following function, when added to and called in ```framework/test/ServiceFactoryTest.cpp```, will reproduce the issue on Linux and Mac. The issue doesn't reproduce on Windows.

```
#ifdef US_ENABLE_THREADING_SUPPORT

// test that concurrent calls to ServiceFactory::GetService and ServiceFactory::UngetService
// don't cause race conditions.
void TestConcurrentServiceFactory()
{
  auto framework = FrameworkFactory().NewFramework();
  framework.Start();

  auto bundle = testing::InstallLib(framework.GetBundleContext(), "TestBundleH");
  bundle.Start();

  std::vector<std::thread> worker_threads;
  for (size_t i = 0; i < 100; ++i)
  {
    worker_threads.push_back(std::thread([framework]()
        {
          for (int i = 0; i < 1000; ++i)
          {
              auto frameworkCtx = framework.GetBundleContext();
              if (frameworkCtx)
              {
                auto ref = frameworkCtx.GetServiceReference<cppmicroservices::TestBundleH>();
                if (ref)
                {
                  std::shared_ptr<cppmicroservices::TestBundleH> svc = frameworkCtx.GetService(ref);
                  US_TEST_CONDITION_REQUIRED(svc, "test for valid service object");
                }
              }
          }
        }));
  }

  for (auto& t : worker_threads) t.join();
}
#endif
```

The stack trace from the resulting seg fault is

```
0x00000000005d5929 in std::_Rb_tree<std::string, std::pair<std::string const, std::shared_ptr<void> >, std::_Select1st<std::pair<std::string const, std::shared_ptr<void> > >, std::less<std::string>, std::allocator<std::pair<std::string const, std::shared_ptr<void> > > >::_M_begin (this=0x0) at gcc-4.9.3/include/c++/4.9.3/bits/stl_tree.h:524
524		  (this->_M_impl._M_header._M_parent);
(gdb) bt
#0  0x00000000005d5929 in std::_Rb_tree<std::string, std::pair<std::string const, std::shared_ptr<void> >, std::_Select1st<std::pair<std::string const, std::shared_ptr<void> > >, std::less<std::string>, std::allocator<std::pair<std::string const, std::shared_ptr<void> > > >::_M_begin (this=0x0) at gcc-4.9.3/include/c++/4.9.3/bits/stl_tree.h:524
#1  0x00000000005d4a57 in std::_Rb_tree<std::string, std::pair<std::string const, std::shared_ptr<void> >, std::_Select1st<std::pair<std::string const, std::shared_ptr<void> > >, std::less<std::string>, std::allocator<std::pair<std::string const, std::shared_ptr<void> > > >::find (this=0x0, __k=...) at gcc-4.9.3/include/c++/4.9.3/bits/stl_tree.h:1926
#2  0x00000000005d3a60 in std::map<std::string, std::shared_ptr<void>, std::less<std::string>, std::allocator<std::pair<std::string const, std::shared_ptr<void> > > >::find (this=0x0, __x=...)
    at gcc-4.9.3/include/c++/4.9.3/bits/stl_map.h:875
#3  0x00007ffff769d6ae in cppmicroservices::ExtractInterface (map=..., interfaceId=...)
    at CppMicroServices/framework/include/cppmicroservices/ServiceInterface.h:325
#4  0x00007ffff769bf81 in cppmicroservices::ServiceReferenceBasePrivate::GetService (this=0x7fffe4000970, bundle=0x931a80)
    at CppMicroServices/framework/src/service/ServiceReferenceBasePrivate.cpp:106
#5  0x00007ffff76c5151 in cppmicroservices::BundleContext::GetService (this=0x7fffdfffee10, reference=...)
    at CppMicroServices/framework/src/bundle/BundleContext.cpp:282
#6  0x00000000005d4958 in cppmicroservices::BundleContext::GetService<cppmicroservices::TestBundleH> (this=0x7fffdfffee10, reference=...)
    at CppMicroServices/framework/include/cppmicroservices/BundleContext.h:581
#7  0x00000000005d0f25 in <lambda()>::operator()(void) const (__closure=0x935708) at ./CppMicroServices/framework/test/ServiceFactoryTest.cpp:198
#8  0x00000000005d3656 in std::_Bind_simple<TestConcurrentServiceFactory()::<lambda()>()>::_M_invoke<>(std::_Index_tuple<>) (this=0x935708)
    at gcc-4.9.3/include/c++/4.9.3/functional:1700
#9  0x00000000005d34be in std::_Bind_simple<TestConcurrentServiceFactory()::<lambda()>()>::operator()(void) (this=0x935708)
    at gcc-4.9.3/include/c++/4.9.3/functional:1688
#10 0x00000000005d33ad in std::thread::_Impl<std::_Bind_simple<TestConcurrentServiceFactory()::<lambda()>()> >::_M_run(void) (this=0x9356f0)
    at gcc-4.9.3/include/c++/4.9.3/thread:115
#11 0x00007ffff70d4970 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#12 0x00007ffff68f2064 in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#13 0x00007ffff662762d in clone () from /lib/x86_64-linux-gnu/libc.so.6
```

jeffdiclemente:
Fixes #202 

This fixes a race condition involving how ServiceReferenceBasePrivate tracks the cached service instance and the list of dependent bundles.

### DIFF
--- a/framework/src/service/ServiceReferenceBasePrivate.cpp
+++ b/framework/src/service/ServiceReferenceBasePrivate.cpp
@@ -140,6 +140,13 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceInterfaceMap(BundleP
       }
       return s;
     }
+
+    if (registration->bundleServiceInstance.end() != registration->bundleServiceInstance.find(bundle))
+    {
+      s = registration->bundleServiceInstance.at(bundle);
+      ++registration->dependents.at(bundle);
+      return s;
+    }
   }
 
   // Calling into a service factory could cause re-entrancy into the
--- a/framework/src/service/ServiceReferenceBasePrivate.cpp
+++ b/framework/src/service/ServiceReferenceBasePrivate.cpp
@@ -130,7 +130,7 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceInterfaceMap(BundleP
   if (!serviceFactory)
   {
     auto l = registration->Lock(); US_UNUSED(l);
-          s = registration->service;
+    s = registration->service;
     if (s && !s->empty()) ++registration->dependents[bundle];
   }
   else
@@ -225,7 +225,7 @@ bool ServiceReferenceBasePrivate::UngetService(const std::shared_ptr<BundlePriva
 
   {
     auto l = registration->Lock(); US_UNUSED(l);
-    int count= registration->dependents[bundle.get()];
+    int count = registration->dependents[bundle.get()];
     if (count > 0)
     {
       hadReferences = true;
--- a/framework/src/service/ServiceReferenceBasePrivate.cpp
+++ b/framework/src/service/ServiceReferenceBasePrivate.cpp
@@ -228,6 +228,11 @@ bool ServiceReferenceBasePrivate::UngetService(const std::shared_ptr<BundlePriva
 
   {
     auto l = registration->Lock(); US_UNUSED(l);
+    if (registration->dependents.end() == registration->dependents.find(bundle.get()))
+    {
+      return hadReferences && removeService;
+    }
+
     int count = registration->dependents.at(bundle.get());
     if (count > 0)
     {
--- a/framework/src/service/ServiceReferenceBasePrivate.cpp
+++ b/framework/src/service/ServiceReferenceBasePrivate.cpp
@@ -47,7 +47,9 @@ ServiceReferenceBasePrivate::ServiceReferenceBasePrivate(ServiceRegistrationBase
 ServiceReferenceBasePrivate::~ServiceReferenceBasePrivate()
 {
   if (registration && !--registration->ref)
+  {
     delete registration;
+  }
 }
 
 InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceFromFactory(
@@ -62,7 +64,10 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceFromFactory(
                                                     ServiceRegistrationBase(registration));
     if (!smap || smap->empty())
     {
-      registration->bundle->coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING, MakeBundle(bundle->shared_from_this()), std::string("ServiceFactory returned an empty or nullptr interface map."), std::make_exception_ptr(std::logic_error("ServiceFactory produced null"))));
+      registration->bundle->coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING, 
+                                                                                  MakeBundle(bundle->shared_from_this()), 
+                                                                                  std::string("ServiceFactory returned an empty or nullptr interface map."), 
+                                                                                  std::make_exception_ptr(std::logic_error("ServiceFactory returned an invalid interface map"))));
       return smap;
     }
     std::vector<std::string> classes = (registration->properties.Lock(),
@@ -72,7 +77,10 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceFromFactory(
       if (smap->find(clazz) == smap->end() && clazz != "org.cppmicroservices.factory")
       {
         std::string message("ServiceFactory produced an object that did not implement: " + clazz);
-        registration->bundle->coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING, MakeBundle(bundle->shared_from_this()), message, std::make_exception_ptr(std::logic_error(message.c_str()))));
+        registration->bundle->coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING, 
+                                                                                    MakeBundle(bundle->shared_from_this()), 
+                                                                                    message, 
+                                                                                    std::make_exception_ptr(std::logic_error(message.c_str()))));
         return nullptr;
       }
     }
@@ -81,7 +89,10 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceFromFactory(
   catch (...)
   {
     s.reset();
-    registration->bundle->coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR, MakeBundle(bundle->shared_from_this()), std::string("ServiceFactory threw an unknown exception."), std::current_exception()));
+    registration->bundle->coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR, 
+                                                                                MakeBundle(bundle->shared_from_this()), 
+                                                                                std::string("ServiceFactory threw an unknown exception."), 
+                                                                                std::current_exception()));
   }
   return s;
 }
@@ -106,7 +117,7 @@ std::shared_ptr<void> ServiceReferenceBasePrivate::GetService(BundlePrivate* bun
   auto s = ExtractInterface(GetServiceInterfaceMap(bundle), interfaceId);
   if (!s)
   {
-    registration->Lock(), --registration->dependents[bundle];
+    registration->Lock(), --registration->dependents.at(bundle);
   }
   return s;
 }
@@ -115,52 +126,42 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceInterfaceMap(BundleP
 {
   InterfaceMapConstPtr s;
   if (!registration->available) return s;
-
   std::shared_ptr<ServiceFactory> serviceFactory;
-  int count = 0;
 
+  auto l = registration->Lock(); US_UNUSED(l);
+  if (!registration->available) return s;
+  serviceFactory = std::static_pointer_cast<ServiceFactory>(
+        registration->GetService_unlocked("org.cppmicroservices.factory"));
+
+  if(registration->dependents.end() == registration->dependents.find(bundle))
   {
-    auto l = registration->Lock(); US_UNUSED(l);
-    if (!registration->available) return s;
-    serviceFactory = std::static_pointer_cast<ServiceFactory>(
-          registration->GetService_unlocked("org.cppmicroservices.factory"));
-    count = registration->dependents[bundle];
+    registration->dependents.insert(std::make_pair(bundle, 0));
   }
 
   if (!serviceFactory)
   {
-    auto l = registration->Lock(); US_UNUSED(l);
     s = registration->service;
-    if (s && !s->empty()) ++registration->dependents[bundle];
+    if (s && !s->empty())
+    {
+      ++registration->dependents.at(bundle);
+    }
+    return s;
+  }
+
+  if (0 == registration->dependents.at(bundle))
+  {
+    // No cached service instance exists. Get a new one and cache it.
+    s = GetServiceFromFactory(bundle, serviceFactory);
+    registration->bundleServiceInstance.insert(std::make_pair(bundle, s));
+    if (s && !s->empty()) ++registration->dependents.at(bundle);
   }
   else
   {
-    if (count == 0)
-    {
-      s = GetServiceFromFactory(bundle, serviceFactory);
-
-      auto l = registration->Lock(); US_UNUSED(l);
-      if (registration->dependents[bundle] == 0)
-      {
-        registration->bundleServiceInstance.insert(std::make_pair(bundle, s));
-      }
-      else
-      {
-        // There was a race and we now have one instance too much. Return the
-        // already produced instance and ignore the additional one.
-        s = registration->bundleServiceInstance[bundle];
-      }
-
-      if (s && !s->empty()) ++registration->dependents[bundle];
-    }
-    else
-    {
-      auto l = registration->Lock(); US_UNUSED(l);
-      // return the already produced instance
-      s = registration->bundleServiceInstance[bundle];
-      if (s && !s->empty()) ++registration->dependents[bundle];
-    }
+    // Return the cached service instance.
+    s = registration->bundleServiceInstance.at(bundle);
+    if (s && !s->empty()) ++registration->dependents.at(bundle);
   }
+
   return s;
 }
 
@@ -196,7 +197,9 @@ bool ServiceReferenceBasePrivate::UngetPrototypeService(const std::shared_ptr<Bu
       catch (...)
       {
         std::string message("ServiceFactory threw an exception");
-        registration->bundle->coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING, MakeBundle(bundle->shared_from_this()), message, std::current_exception()));
+        registration->bundle->coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING, 
+                                                                                    MakeBundle(bundle->shared_from_this()), message, 
+                                                                                    std::current_exception()));
       }
 
       auto l = registration->Lock(); US_UNUSED(l);
@@ -225,7 +228,7 @@ bool ServiceReferenceBasePrivate::UngetService(const std::shared_ptr<BundlePriva
 
   {
     auto l = registration->Lock(); US_UNUSED(l);
-    int count = registration->dependents[bundle.get()];
+    int count = registration->dependents.at(bundle.get());
     if (count > 0)
     {
       hadReferences = true;
@@ -235,7 +238,7 @@ bool ServiceReferenceBasePrivate::UngetService(const std::shared_ptr<BundlePriva
     {
       if (count > 1)
       {
-        registration->dependents[bundle.get()] = count - 1;
+        --registration->dependents.at(bundle.get());
       }
       else if(count == 1)
       {
@@ -249,7 +252,11 @@ bool ServiceReferenceBasePrivate::UngetService(const std::shared_ptr<BundlePriva
 
     if (removeService)
     {
-      sfi = registration->bundleServiceInstance[bundle.get()];
+      if (registration->bundleServiceInstance.end() != registration->bundleServiceInstance.find(bundle.get()))
+      {
+        sfi = registration->bundleServiceInstance.at(bundle.get());
+      }
+
       if (sfi && !sfi->empty())
       {
         sf = std::static_pointer_cast<ServiceFactory>(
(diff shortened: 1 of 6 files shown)


framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -106,7 +117,7 @@ std::shared_ptr<void> ServiceReferenceBasePrivate::GetService(BundlePrivate* bun
   auto s = ExtractInterface(GetServiceInterfaceMap(bundle), interfaceId);
   if (!s)
   {
-    registration->Lock(), --registration->dependents[bundle];
+    registration->Lock(), --registration->dependents.at(bundle);
```
saschazelzer: If `GetServiceInterfaceMap` above returned early, the `.at(bundle)` might throw now, which would propagate into client code. I think it would be better to `find` the bundle.

Also (this problem is not new), if `GetServiceInterfaceMap` returns before incrementing the count, we might decrement it "too much".

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -115,52 +126,42 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceInterfaceMap(BundleP
 {
   InterfaceMapConstPtr s;
   if (!registration->available) return s;
-
   std::shared_ptr<ServiceFactory> serviceFactory;
-  int count = 0;
 
+  auto l = registration->Lock(); US_UNUSED(l);
```
saschazelzer: Now this lock is held while calling into client code (the service factoy), which must be avoided. The original code avoided that, but had that race condition you found.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -115,52 +126,42 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceInterfaceMap(BundleP
...
 
   if (!serviceFactory)
   {
-    auto l = registration->Lock(); US_UNUSED(l);
-          s = registration->service;
-    if (s && !s->empty()) ++registration->dependents[bundle];
+    s = registration->service;
+    if (s && !s->empty())
+    {
+      ++registration->dependents.at(bundle);
```
saschazelzer: Why the `.at`? If the bundle is not know to the map, we should insert it, using`[]`.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -115,52 +126,42 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceInterfaceMap(BundleP
...
+    }
+    return s;
+  }
+
+  if (0 == registration->dependents.at(bundle))
```
saschazelzer: The `.at` could throw again. `[]` is find in my opinion (having a bundle entry with default `0` count is fine).

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -115,52 +126,42 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceInterfaceMap(BundleP
...
+  {
+    // No cached service instance exists. Get a new one and cache it.
+    s = GetServiceFromFactory(bundle, serviceFactory);
+    registration->bundleServiceInstance.insert(std::make_pair(bundle, s));
+    if (s && !s->empty()) ++registration->dependents.at(bundle);
```
saschazelzer: Same `.at` question here.

framework/test/bundles/libH/TestBundleH.cpp
```diff
@@ -66,19 +67,20 @@ class TestProduct2 : public TestProduct, public TestBundleH2
 class TestBundleHPrototypeServiceFactory : public PrototypeServiceFactory
```
saschazelzer: The service factory implementation could call back into the framework (by e.g. getting another service). This would test if you get a deadlock, which should happen with the proposed changes.

framework/src/service/ServiceReferenceBasePrivate.cpp
jeffdiclemente: I'm missing something. I don't see a case where ```GetServiceInterfaceMap``` could return a valid ```InterfaceMapConstPtr``` object without ensuring that ```registration->dependents``` has a valid element in it with an incremented count.

framework/src/service/ServiceReferenceBasePrivate.cpp
jeffdiclemente: The insertion happens on lines 136-139.

framework/src/service/ServiceReferenceBasePrivate.cpp
jeffdiclemente: I agree, we shouldn't hold locks while calling client code. In this case, I think the only way a deadlock occurs is if the client's ```ServiceFactory::GetService``` implementation calls ```GetServiceReference/GetService``` on itself since the lock is on the ```ServiceRegistrationBasePrivate``` object associated with the ```ServiceFactory``` object. I'm sure a deadlock could occur, I don't know how likely that is to happen.

An alternative solution is to scope the locks such that ```GetServiceFromFactory(...)``` is unconditionally called and the return value is discarded if a cached service object already exists. This seems wasteful.

These are the only two options I could come up with that fixed the race condition. What are your thoughts?

framework/src/service/ServiceReferenceBasePrivate.cpp
jeffdiclemente: read/writes to ```registration->dependents``` are guarded in ```ServiceReferenceBasePrivate::GetServiceInterfaceMap``` and ```ServiceReferenceBasePrivate::UngetService```. At this point in the code ```.at()``` should be guaranteed to find an element. Did I miss a spot where ```registration->dependents``` is not guarded?

framework/test/bundles/libH/TestBundleH.cpp
jeffdiclemente: excellent idea. I'm going to take a little time to do this and see if I can break the framework.

framework/src/service/ServiceReferenceBasePrivate.cpp
saschazelzer: For a valid `InterfaceMapConstPtr` the code looks fine, but I am concerned about the invalid case.

First, if `GetServiceInterfaceMap` returns a `null` map (in case `registration->available` if `false` or due to some other condition (e.g. the race you found), then `ExtractInterface` will dereference an nullptr (we should fix that).

Second, in the invalid case, the counter was not incremented in `GetServiceInterfaceMap` (I believe) but we decrement it here. The decrement is necessary if `GetServiceInterfaceMap` returns a valid map (and hence incremented the counter) but `ExtractInterface` fails (because e.g. `interfaceId` is empty. We currently do not differentiate the two cases.

framework/src/service/ServiceReferenceBasePrivate.cpp
saschazelzer: I agree, something needs to call `GetService` on the same service registration. But since a service factory can call into arbitrary framework API, e.g. starting bundles, which then can again do arbitrary things in their activator etc., I think we should make that bullet proof.

Discarding a second instance was the strategy used before. This is an exceptional case and not likely to occur. I would favor no deadlocks instead of optimizing service creation for corner cases :)

To avoid the race you found, would it be enough to check (in the first registration lock scope) if we have a service factory and the usage count is > 0, then return the cached instance? If not, go out of scope and execute the original code.

framework/src/service/ServiceReferenceBasePrivate.cpp
saschazelzer: Ah, you are right. But then we can't we just use `[]` instead?

framework/src/service/ServiceReferenceBasePrivate.cpp
saschazelzer: If think you are right. I would still prefer the usage of `[]` in that area. If existence is guaranteed anyway, we can avoid the bounds checking done in `.at`. Even if it is not guaranteed, creating an entry does not do harm and we do not need to reason about existence and possible `out_of_range` exceptions.

framework/src/service/ServiceReferenceBasePrivate.cpp
jeffdiclemente: Understood. I'll fix the logic.

framework/src/service/ServiceReferenceBasePrivate.cpp
jeffdiclemente: Agreed on removing the deadlock. I prematurely optimized for service creation.

I'm not sure if that will be sufficient to eliminate the race condition. I will take a closer look. 
In general, since ```ServiceReferenceBasePrivate::GetServiceInterfaceMap``` and ```ServiceReferenceBasePrivate::UngetService``` both modify the ```registration->dependents``` and ```registration->bundleServiceInstance``` data structures concurrently, all access (i.e. read and write) must be done within locks to guarantee no race conditions.

framework/src/service/ServiceReferenceBasePrivate.cpp
jeffdiclemente: Thanks for the suggestion, I've used it to simplify this logic. I will have an update to this PR shortly.

I prefer using ```at()``` over ```operator[]``` since I feel it makes the code easier to maintain. While all uses of ```at()``` are guaranteed not to throw, an exception being thrown would indicate a bug in the logic. If ```operator[]``` is used a potential bug of this kind would be harder to see.

jeffdiclemente:
FYI, failures are from travis CI timing out on memcheck_ServiceFactoryTest

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -103,64 +114,51 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetPrototypeService(const Bund
 
 std::shared_ptr<void> ServiceReferenceBasePrivate::GetService(BundlePrivate* bundle)
 {
-  auto s = ExtractInterface(GetServiceInterfaceMap(bundle), interfaceId);
-  if (!s)
-  {
-    registration->Lock(), --registration->dependents[bundle];
-  }
-  return s;
+  return ExtractInterface(GetServiceInterfaceMap(bundle), interfaceId);
```
saschazelzer: The impl of `ExtractInterface` should still be improved. When it gets a null shared_ptr, it will dereference it. I think this is were the code originally crashed due to the race in `GetServiceInterfaceMap`.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -103,64 +114,51 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetPrototypeService(const Bund
...
 }
 
 InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceInterfaceMap(BundlePrivate* bundle)
 {
   InterfaceMapConstPtr s;
   if (!registration->available) return s;
-
   std::shared_ptr<ServiceFactory> serviceFactory;
-  int count = 0;
```
saschazelzer: I think we should keep that variable, see comment below.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -103,64 +114,51 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetPrototypeService(const Bund
...
+      return s;
     }
   }
+
+  // Calling into a service factory could cause re-entrancy into the
+  // framework and even, theoretically, into this function. Ensuring
+  // we don't hold a lock while calling into the service factory eliminates
+  // the possibility of a deadlock. It does not however eliminate the
+  // possbilituy of infinite recursion.
+  s = GetServiceFromFactory(bundle, serviceFactory);
```
saschazelzer: At this point, we are unlocked and know, that we have a service factory. Previously, we would try to reuse a previously cached service instance but we were missing a check for the race condition. I think we can still avoid calling `GetServiceFromFactory` every time and reuse the cached version while checking a possible race:

Case 1 `count < 1`: This case was handled fine in the original code. We call `GetServiceFromFactory`, then lock and check the count again. If it is still < 1 we insert the new instance into our cache (the map) and are done. If It is >= 1 now, we discard the newly created instance and return the (recently added) instance from the cache.

Case 2 `count >= 1`: Lock and get the cached instance. Now (this was missing originally) if the instance is `null` (because of the race) we need to go back to case 1 (probably do case 1 and 2 in a `while(!s)` loop. If it is not `null`, we increase the count and return.

Did I miss something?

framework/test/bundles/libH/TestBundleH.cpp
```diff
@@ -88,30 +101,106 @@ class TestBundleHPrototypeServiceFactory : public PrototypeServiceFactory
 class TestBundleHServiceFactory : public ServiceFactory
 {
   std::map<long, std::shared_ptr<TestProduct>> fcbind;   // Map calling bundle with implementation
+  std::mutex fcbindLock;
 public:
 
   InterfaceMapConstPtr GetService(const Bundle& caller, const ServiceRegistrationBase& /*sReg*/)
   {
-    std::cout << "GetService in H" << std::endl;
+    // This code causes infinite recursion into this function. It can be used as a test case
```
saschazelzer: Nice. I will enable this when working on #213 .

framework/src/service/ServiceReferenceBasePrivate.cpp
jeffdiclemente: good catch. The code changes should guarantee that ```nullptr``` is never passed to ```ExtractInterface``` (in all failure modes, it should be a default constructed ```InterfaceMapConstPtr```). Nevertheless its a good idea to ensure it never can dereference a ```nullptr```. I'll fix it.

framework/src/service/ServiceReferenceBasePrivate.cpp
jeffdiclemente: My goal was to simplify the code and reduce the number of conditional branches (at the cost of multiple ```GetServiceFromFactory``` calls).
 
It is possible to avoid calling ```GetServiceFromFactory``` each time and still have no race condition. The drawback is that duplicate code would be required in both locked regions of code. 
Once the code exits the first lock, no assumptions can be made about the state of ```registration->dependents``` and ```registration->bundleServiceInstance``` when entering the second lock. Both locked regions of code need to do similar, if not the same, checks after acquiring the lock.

Is this drawback (i.e. duplicate code) acceptable?

saschazelzer:
In my opinion, yes. We are only talking about a few lines of code here. But service creation via a factory can be arbitrarily costly, because it is not under our control.

On June 23, 2017 4:37:27 PM GMT+02:00, Jeff DiClemente <notifications@github.com> wrote:
>jeffdiclemente commented on this pull request.
>
>
>
>>      }
>   }
>+
>+  // Calling into a service factory could cause re-entrancy into the
>+  // framework and even, theoretically, into this function. Ensuring
>+  // we don't hold a lock while calling into the service factory
>eliminates
>+  // the possibility of a deadlock. It does not however eliminate the
>+  // possbilituy of infinite recursion.
>+  s = GetServiceFromFactory(bundle, serviceFactory);
>
>My goal was to simplify the code and reduce the number of conditional
>branches (at the cost of multiple ```GetServiceFromFactory``` calls).
> 
>It is possible to avoid calling ```GetServiceFromFactory``` each time
>and still have no race condition. The drawback is that duplicate code
>would be required in both locked regions of code. 
>Once the code exits the first lock, no assumptions can be made about
>the state of ```registration->dependents``` and
>```registration->bundleServiceInstance``` when entering the second
>lock. Both locked regions of code need to do similar, if not the same,
>checks after acquiring the lock.
>
>Is this drawback (i.e. duplicate code) acceptable?
>
>-- 
>You are receiving this because your review was requested.
>https://github.com/CppMicroServices/CppMicroServices/pull/214#discussion_r123763118

-- 

jeffdiclemente:
Ok. I will make that change

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -103,64 +114,58 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetPrototypeService(const Bund
...
+
+  auto l = registration->Lock(); US_UNUSED(l);
+
+  registration->dependents.insert(std::make_pair(bundle, 0));
+
+  if (s && !s->empty())
```
saschazelzer: There is a really strange but possible scenario: What if the call to the service factory above returned a null service object, but another thread created a valid instance a micro second earlier and the framework cached it. Then we would return a null service even though we have a non-null cached instance. I think we should check for an existing cached service instance in an else branch and then return it if available.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -103,64 +114,58 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetPrototypeService(const Bund
...
+  {
+    auto insertResultPair = registration->bundleServiceInstance.insert(std::make_pair(bundle, s));
+    s = insertResultPair.first->second;
```
saschazelzer: The original code had a comment about discarding a newly created service instance in case of a race here. It would probably be good to have that comment back in.

jeffdiclemente: sure. will do.

framework/src/service/ServiceReferenceBasePrivate.cpp
jeffdiclemente: I agree, it is a strange yet possible scenario. 

If one thread has a null service object and another has a valid service object, what happens all depends on which thread enters the lock first.

If a service factory does return a null service object, I don't feel its the Framework's responsibility to try and guarantee that a valid one is always returned (excluding cases where one exists in the cache already). If a client's service factory returns a null service object on one thread and then a valid one on another, that seems like a bug; either with the client code querying for the service object or the client's service factory implementation. 

I'm not sure how the else condition really helps ensure that a non-null service is returned. This feels like we're overcompensating for possible bugs in client code. To check in an else branch we still have to hold the lock.

saschazelzer: I agree that we cannot control the order, but we do not have to. The only case I think we should cover is that if we get a null service but have a non-null service cached because of a race, return the non-null service (just because we can). A service factory is allowed to return null and there is nothing we can do against that.

jeffdiclemente: Ok. That case sounds reasonable to me. I'll make the change.

PR #216, opened 2017-06-23
Added new framework working dir property.

### LINKED ISSUE #209: Current working directory handling is not thread safe
We read the current working directory during framework initialization. This is not safe in a multi-threaded program and should be avoided.

The working directory should be passed as a framework property instead.

saschazelzer:
This fixes #209 by adding a new framework launch property for setting the current working directory. The unsafe working directory related OS functions are only called once during static initialization.

As a side effect, changed the framework properties map from `std::map` to `std::unordered_map`. The ordered map is a legacy from our pre-C++11 development.

### DIFF
--- a/framework/src/util/Utils.cpp
+++ b/framework/src/util/Utils.cpp
@@ -129,7 +129,7 @@ std::vector<std::string> SplitString(const std::string& str, const std::string& 
   return token;
 }
 
-std::string GetCurrentWorkingDirectory()
+std::string InitCurrentWorkingDirectory()
 {
 #ifdef US_PLATFORM_WINDOWS
   DWORD bufSize = ::GetCurrentDirectoryA(0, NULL);
@@ -152,6 +152,13 @@ std::string GetCurrentWorkingDirectory()
   }
 #endif
   return std::string();
+}
+
+static const std::string s_CurrWorkingDir = InitCurrentWorkingDirectory();
+
+std::string GetCurrentWorkingDirectory()
+{
+  return s_CurrWorkingDir;
 }
 
 bool Exists(const std::string& path)
@@ -206,9 +213,9 @@ bool IsRelative(const std::string& path)
 #endif
 }
 
-std::string GetAbsolute(const std::string& path)
+std::string GetAbsolute(const std::string& path, const std::string& base)
 {
-  if (IsRelative(path)) return GetCurrentWorkingDirectory() + DIR_SEP + path;
+  if (IsRelative(path)) return base + DIR_SEP + path;
   return path;
 }
 
@@ -315,7 +322,7 @@ bool IsBundleFile(const std::string& location)
       std::vector<uint32_t> indices;
 
       resContainer.GetChildren(dir + "/", true, names, indices);
-      return std::any_of(names.begin(), names.end(), [](const std::string& resourceName) -> bool { 
+      return std::any_of(names.begin(), names.end(), [](const std::string& resourceName) -> bool {
         return resourceName == std::string("manifest.json");
       });
     });
@@ -351,7 +358,7 @@ std::string GetFileStorage(CoreBundleContext* ctx, const std::string& name, bool
   {
     return fwdir;
   }
-  const std::string dir = fs::GetAbsolute(fwdir) + DIR_SEP + name;
+  const std::string dir = fs::GetAbsolute(fwdir, ctx->workingDir) + DIR_SEP + name;
   if (!dir.empty())
   {
     if (fs::Exists(dir))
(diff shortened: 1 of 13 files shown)


framework/src/util/Utils.cpp
```diff
@@ -154,6 +154,13 @@ std::string GetCurrentWorkingDirectory()
   return std::string();
 }
 
+static const std::string s_CurrWorkingDir = InitCurrentWorkingDirectory();
```
jeffdiclemente: global statics are problematic. This doesn't look like it will cause any static destruction issues now. As the code evolves, it may in future. This is more of a future looking code maintenance concern than anything else.
Can a function local static be used instead? That is thread-safe and guaranteed to be called once (for C++11).

saschazelzer: We need (from my understanding) to fulfill the following requirement here:

1. Get the current working directory in a thread-safe way

This affects all calls to the underlying OS provided functions, from any piece of code running in the current process. The only safe way to avoid synchronization issues I know of is to call such functions during static initialization of the library, which means constructing a static global. I would be happy to see another solution. Function local statics do not help because they cannot synchronize calls to the underlying OS function.

saschazelzer: Hm, I am not sure we are talking about the same thing. Some OS API (in this particular case the Windows `GetCurrentDirectory` function) must not be called concurrently. How would a static local or call_once approach prevent clients from calling a framework function which initializes the static local (or calls call_once) while concurrently calling `GetCurrentDirectory` through some other code path? Whenever multiple threads can be created in the process, there is no way for any library to synchronize on OS functions (except for some techniques we really do not want to use I would say).

So static initialization - when the process is effectively in "single-threading-mode" - is the only viable solution as far as I can see. Or not call the function at all and require a framework property at framework initialization.

Regarding library dependencies and static globals, the concerns are valid in case of static linking. However, the POSIX `getcwd` is a syscall on Linux (and possibly others) and the Windows `GetCurrentDirectory` is provided by kernel32.dll and cannot be statically linked. If on some systems the current directory cannot be retrieved during initialization, it can still be set as a framework property externally.

jeffdiclemente: It is possible that I have some hidden assumptions in my head 😄 
My assumptions are:

- ```InitCurrentWorkingDirectory``` is only ever called once for the lifetime of the process
- C++11 magic statics can be used
- The function local static is constructed in  ```GetCurrentWorkingDirectory()``` 
- ```InitCurrentWorkingDirectory()``` is only ever called by ```GetCurrentWorkingDirectory()``` 

For a function local static in C++11 (i.e. magic statics), the following would guarantee that ```GetCurrentDirectoryA``` was called only once and only on one thread, even if GetCurrentWorkingDirectory() is called multiple times from multiple threads:

```
std::string& GetCurrentWorkingDirectory()
{
  static std::string s_CurrWorkingDir(InitCurrentWorkingDir());
  return s_CurrWorkingDir;
}
```
On construction of ```s_CurrWorkingDir```, ```InitCurrentWorkingDir()``` will only be called once, no matter how many times ```GetCurrentWorkingDirectory()``` is called since function local statics are only constructed once in a thread-safe way.


As for ```std::call_once``` - after writing some example code for ```std::call_once```, it seems like an inferior choice to function local statics since the current working directory```std::string``` needs to be an in/out parameter to the ```Callable``` object. From what I see, the only significant difference between this and function local statics is that ```std""call_once``` handles exceptions.

saschazelzer: I am not sure why you think it is about `InitCurrentWorkingDirectory`. We need to ensure that we do not call `GetCurrentDirectory` in an unsafe way (provided by kernel32.dll). We cannot control who calls that function and when or in what thread it is called during a progams lifetime. How can a magic function local static help here? I fail to see that.

jeffdiclemente: Sorry. I finally understand your point. Yes, the Win32 API ```GetCurrentDirectory``` reads a global variable and yes, the only way to guarantee that CppMicroServices never calls into ```GetCurrentDirectory``` concurrently is to call it during static initialization or not at all.

I'm wondering whether there is an even simpler approach now...
Considering how the current working directory is being used in CppMicroServices to construct absolute paths to folders, what about doing away with calling ```GetCurrentDirectory``` altogether and use the Win32 API ```SearchPath``` instead to construct an absolute path?

saschazelzer: By reading msdn for `SearchPath`, it seems that you have to provide a file name and it tries to find an existing path in the filesystem based on some search logic. To me, it doesn't look like it would be useful for deriving e.g. the framework directory path using this function. In this case, we do not have a file name (only a partial path) and it isn't required to exist. But I haven't used `SearchPath` on my own yet.

PR #219, opened 2017-07-05
Codecov

karthikreddy09:
Enable code coverage reports for all branches.
Reports are uploaded to https://codecov.io/gh/cppmicroservices


fixcoveragefilepaths.py
```diff
@@ -0,0 +1,67 @@
+import xml.etree.ElementTree as xml
+import glob, os, re, sys, getopt
```
ksubramz: Needn't necessarily address it now but [argparse](https://docs.python.org/2/howto/argparse.html) is the idiomatic and the recommended way to do cmdline option parsing.

CMakeLists.txt
```diff
@@ -381,6 +405,10 @@ if(US_ENABLE_THREADING_SUPPORT AND US_TSAN_LINK_FLAGS)
   set(US_LINK_FLAGS "${US_LINK_FLAGS} ${US_TSAN_LINK_FLAGS}")
 endif()
 
+if(US_ENABLE_CCOV AND US_CXX_CCOV_FLAGS)
+  set(US_LINK_FLAGS "${US_LINK_FLAGS} ${US_CXX_CCOV_FLAGS}")
```
jeffdiclemente: ```US_CXX_CCOV_FLAGS``` is used to both set the CXX flags and the linker flags. Should the same variable be used? Do the flags for the compiler and linker ever differ?

fixcoveragefilepaths.py
```diff
@@ -0,0 +1,67 @@
+import xml.etree.ElementTree as xml
```
saschazelzer: Could you add a brief high-level description why this file is necessary and who it solves / works around some problem? Is there a known bug report or similar which we can link to?

opencppcoverage_mergereports.bat
```diff
@@ -0,0 +1,13 @@
+@echo off
```
saschazelzer: Same here, could you add a briefly describe the purpose of this file inline?

appveyor.yml
```diff
@@ -151,8 +151,9 @@ before_build:
 - echo %PATH%
 - if "%platform%"=="mingw-w64" set PATH=%PATH:C:\Program Files\Git\usr\bin;=%
 - if "%platform%"=="mingw-w64" set PATH=C:\msys64\mingw64\bin;%PATH%
+- if [%WITH_COVERAGE%]==[1] set PATH="%PATH%:C:\Program Files\OpenCppCoverage;"
```
saschazelzer: Typo, should be `;` instead of `:`.

.travis.yml
```diff
@@ -15,11 +14,11 @@ matrix:
             apt:
               packages: valgrind
           compiler: gcc
-          env: BUILD_CONFIGURATION=0 WITH_COVERAGE=1
+          env: BUILD_CONFIGURATION=0 WITH_COVERAGE=1 BUILD_DIR="/tmp/cppmicroservices_builds"
```
saschazelzer: Is the default build path `/tmp/us builds` troublesome because of the space? Even though the `BUILD_DIR` env var is not global now, it is used in all builds except osx builds. So we would be missing CI failures due to spaces in the build path.

karthikreddy09: Yes, the space is causing the final upload script to trip. I had to reconfirmed by running some builds in CI. So I made the environment variable specific to code coverage builds. We cover the  use case with a space in the directory path, in the builds without code coverage.

appveyor.yml
karthikreddy09: Thanks Sascha! I will make the change.

appveyor.yml
```diff
@@ -149,14 +151,23 @@ before_build:
 - echo %PATH%
 - if "%platform%"=="mingw-w64" set PATH=%PATH:C:\Program Files\Git\usr\bin;=%
```
jeffdiclemente: shouldn't this be ```%PATH%```, not ```%PATH```?

also, what does the ```=%``` at the end do?

appveyor.yml
```diff
@@ -149,14 +151,23 @@ before_build:
...
 - if "%platform%"=="mingw-w64" set PATH=C:\msys64\mingw64\bin;%PATH%
```
jeffdiclemente: why not combine this with the above line?

appveyor.yml
saschazelzer: This is the syntax for Windows batch file string replacements. The line replaces `C:\Program Files\Git\usr\bin` with the empty string in `%PATH%`.

appveyor.yml
saschazelzer: The first line does string replacement. The second prepends a new entry to PATH.

jeffdiclemente: Ok. I see now. Thanks for the explanation.

appveyor.yml
karthikreddy09: Thanks for the explanation @saschazelzer. I was confused by this as well.

karthikreddy09:
The codecov check failures are expected since there is no existing report for development branch.

Issue #217, opened 2017-06-29
Use correct framework event severity and exception types for service factory errors


saschazelzer:
Review the severity used when sending framework events. At least in the service factory area, we are sending warnings instead of errors.

Also we use `std::logic_error` instead of the existing `ServiceException` class when transporting exceptions via a framework event.

PR #222, opened 2017-07-26
Handle recursive service factory calls.

saschazelzer:
This checks if a service factory is called recursively from the same thread for the same bundle.

The documentation and cmake changes are actually larger then the code itself. Code changes include some minor improvements as well.

The minimum compiler versions did not change. However, a recommended list of compiler versions was added and if a lower version is used, warnings are printed during cmake configuration.


framework/include/cppmicroservices/Bundle.h
```diff
@@ -329,11 +329,11 @@ class US_Framework_EXPORT Bundle
...
    *
-   * If this bundle is not in the <code>STATE_STARTED</code> state, then this
-   * bundle has no valid <code>BundleContext</code> and this method will
-   * return a default constructed \c BundleContext object.
+   * If this bundle is not in the \c STATE_STARTING, \c STATE_ACTIVE, or
+   * \c STATE_STOPPING states, then this bundle has no valid \c BundleContext
+   * and this method will an invalid \c BundleContext object.
    *
-   * @return A valid or invalid <code>BundleContext</code> for this bundle or.
+   * @return A valid or invalid <code>BundleContext</code> for this bundle.
```
saschazelzer: Maybe... although I think repeating it here doesn't hurt.

framework/include/cppmicroservices/ServiceFactory.h
```diff
@@ -70,25 +65,35 @@ class ServiceFactory
...
    *
-   * <p>
-   * The framework caches the value returned (unless the InterfaceMap is empty),
-   * and will return the same service object on any future call to
-   * <code>BundleContext::GetService</code> for the same bundles. This means the
-   * framework does not allow this method to be concurrently called for the
-   * same bundle.
+   * The framework caches the valid service object, and will return the same
+   * service object on any future call to \c BundleContext::GetService for the
+   * specified bundle. This means the framework does not allow this method to
```
saschazelzer: The wording is basically taken from the specs (the API documentation). "Whom"  are the clients getting service objects. They indirectly call this function via the framework API.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -119,41 +122,83 @@ std::shared_ptr<void> ServiceReferenceBasePrivate::GetService(BundlePrivate* bun
 
 InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceInterfaceMap(BundlePrivate* bundle)
 {
+  /*
```
saschazelzer: The specs only talk about recursion from the same thread.

framework/test/bundles/libH/TestBundleH.cpp
```diff
@@ -187,6 +168,50 @@ class TestBundleHServiceFactoryReturnsNullPtr : public ServiceFactory
...
 };
 
+// Simulate an error condition whereby the service factory recursively tries to get the same service for the same bundle
+class TestBundleHServiceFactoryRecursion : public ServiceFactory
+{
+public:
+
+  InterfaceMapConstPtr GetService(const Bundle& caller, const ServiceRegistrationBase& sReg)
+  {
+    static bool validRecursion = false;
```
saschazelzer: I will add some more comments, but I think you misread the code. A call from the framework context will cause infinite recursion. A call from a non-framework context will recurse once, by retrieving the service on behalf of the framework context. The `return nullptr` is a no-op, it is never reached.

karthikreddy09: Thanks for the explanation @saschazelzer. I think I understand the difference now.

The returned `nullptr` I was referring to is the value returned when the framework breaks out of the infinite recursion, not the `return nullptr` statement in this function. (See line#204 of ServiceFactoryTest.cpp).

More comments would certainly help. Also, It would be cleaner if we could get rid of the unreachable statement `return nullptr`. The loop  and the unreachable return statement can be avoided with the following block.
```
if (caller.GetBundleId() > 0)
{
  auto systemBundle = caller.GetBundleContext().GetBundle(0);
  return systemBundle.GetBundleContext().GetService(sReg.GetReference());
}
```

framework/include/cppmicroservices/ServiceFactory.h
karthikreddy09: sounds reasonable, no concern from me.

framework/src/service/ServiceReferenceBasePrivate.cpp
karthikreddy09: Maybe I have missed it in the spec. Do you mind pointing me to the right section?

I found the snippet "If this method is recursively called for the specified bundle, a
framework event of type FrameworkEvent.ERROR is fired containing a service exception of type ServiceException.FACTORY_RECURSION and null is returned to the bundle." on [page#255 of the OSGi core spec](https://osgi.org/download/r6/osgi.core-6.0.0.pdf#page=255). It does not call out recursion on the same thread. 

Also, I was thinking when we add code to satisfy a requirement from the spec, it would be helpful if we had a comment with the link to spec page where the requirement came from. Would be easier to track in the future.

framework/test/bundles/libH/TestBundleH.cpp
saschazelzer: Okay, yes the return value of `GetService` is different because of these two test cases being covered. Your snippet is much nicer, I will use it!

framework/src/service/ServiceReferenceBasePrivate.cpp
saschazelzer: Ah, I probably read or inferred that from the source code of a Java implementation. But I have thought about your use case as well and I do not see how we could differentiate such a case from a legitimate concurrent access which is not recursive. If you want to shoot yourself in the foot, you certainly can. Btw, the two Java implementations I looked at implemented this in different ways, none of them according to the spec as far as I can tell...

karthikreddy09: Yeh, I have been thinking about this use case and could not come up with any strategy to detect recursion when multiple threads are involved. It's an edge case and users most likely won't do that. I am OK with this implementation.

jeffdiclemente:
FYI; reviewing this is on my todo list. I'm working on submitting the gtest integration PR first.

CMakeLists.txt
```diff
@@ -65,37 +66,42 @@ set(CMAKE_CXX_STANDARD 11)
...
+
+set(US_COMPILER_GNU_MINIMUM_RECOMMENDED_VERSION         4.8   )
+set(US_COMPILER_Clang_MINIMUM_RECOMMENDED_VERSION       3.3   )
+set(US_COMPILER_AppleClang_MINIMUM_RECOMMENDED_VERSION  8.0   )
+set(US_COMPILER_MSVC_MINIMUM_RECOMMENDED_VERSION        19.00 )
+
+set(US_COMPILER_GNU_MINIMUM_RECOMMENDED_VERSION_PRODUCT        "GCC ${US_COMPILER_GNU_MINIMUM_RECOMMENDED_VERSION}")
+set(US_COMPILER_Clang_MINIMUM_RECOMMENDED_VERSION_PRODUCT      "Clang ${US_COMPILER_CLANG_MINIMUM_RECOMMENDED_VERSION}")
+set(US_COMPILER_AppleClang_MINIMUM_RECOMMENDED_VERSION_PRODUCT "Xcode 8.0 (Clang ${US_COMPILER_AppleClang_MINIMUM_RECOMMENDED_VERSION})")
+set(US_COMPILER_MSVC_MINIMUM_RECOMMENDED_VERSION_PRODUCT       "Visual Studio 2013 Update 4 (MSVC ${US_COMPILER_MSVC_MINIMUM_RECOMMENDED_VERSION})")
```
saschazelzer: Good catch.

framework/include/cppmicroservices/Bundle.h
```diff
@@ -329,11 +329,11 @@ class US_Framework_EXPORT Bundle
...
    *
-   * If this bundle is not in the <code>STATE_STARTED</code> state, then this
-   * bundle has no valid <code>BundleContext</code> and this method will
-   * return a default constructed \c BundleContext object.
+   * If this bundle is not in the \c STATE_STARTING, \c STATE_ACTIVE, or
+   * \c STATE_STOPPING states, then this bundle has no valid \c BundleContext
+   * and this method will return an invalid \c BundleContext object.
    *
-   * @return A valid or invalid <code>BundleContext</code> for this bundle or.
+   * @return A valid or invalid <code>BundleContext</code> for this bundle.
```
saschazelzer: Yeah, I didn't do a thorough cleanup, just simplified what I came across and might have missed some. I suggest to do a doc cleanup separately.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -64,9 +67,9 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceFromFactory(
                                                     ServiceRegistrationBase(registration));
     if (!smap || smap->empty())
     {
-      registration->bundle->coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING, 
-                                                                                  MakeBundle(bundle->shared_from_this()), 
-                                                                                  std::string("ServiceFactory returned an empty or nullptr interface map."), 
+      registration->bundle->coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING,
```
saschazelzer: True, on the other hand, I didn't want to change this because it is not related to the actual task being worked on. I will change the documentation instead and review the event codes later in #217.

framework/include/cppmicroservices/Bundle.h
jeffdiclemente: a separate doc cleanup task sounds good to me.

framework/src/service/ServiceReferenceBasePrivate.cpp
saschazelzer: Actually, I would like to keep it as it is for now, even though it is slightly inconsistent. I would have to rewrite the whole doc block, just to change it back soon after when working on #217.

saschazelzer:
@jeffdiclemente What do you think?

PR #225, opened 2017-08-02
200 gtest integration

### LINKED ISSUE #200: Add GTest to CppMicroServices
Add the ability to write GTest unit tests.

jeffdiclemente:
Integrate gtest into CppMicroServices.

Fixes #200


framework/test/gtest/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+if (WIN32 AND US_USE_SYSTEM_GTEST)
+  foreach(lib_fullpath ${GTEST_BOTH_LIBRARIES})
+    get_filename_component(dir ${lib_fullpath} DIRECTORY)
+    get_filename_component(name_no_ext ${lib_fullpath} NAME_WE)
+    set(dll_file "${dir}/${name_no_ext}${CMAKE_SHARED_LIBRARY_SUFFIX}")
+    add_custom_command(TARGET ${us_gtest_test_exe_name} POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different
+        "${dll_file}"
+        $<TARGET_FILE_DIR:${us_gtest_test_exe_name}>)
+  endforeach(lib_fullpath)
```
saschazelzer: What happens here if the external gtest build is a static build? Shouldn't `dll_file` be non-existent then? I am thinking if we should leave it up to the user (on Windows) to configure his `PATH` env variable or copy files around if he uses an external shared gtest build.

jeffdiclemente: What will happen is that the static libs will be copied, which is irrelevant in terms of running the EXE. At worst it copies the gtest static libs where they are not needed.

I can only speak to our use case, and copying the files is preferable because it doesn't pollute the ```PATH``` and keeps everything isolated. This is useful when doing large parallel builds of many different software projects.

framework/test/gtest/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
+#-----------------------------------------------------------------------------
+# Build and run the GTest Suite of tests
+#-----------------------------------------------------------------------------
+
+set(us_gtest_test_exe_name FrameworkTests)
```
jeffdiclemente: will do.

jeffdiclemente:
> What do you think about trimming the GTest sources, e.g. by removing old docs, unit tests, project files, etc?

I think it will make it more difficult to update the sources in the future (how will anyone know what to add/remove, unless we write it down somewhere). However, the real problem for me is that I don't know which parts to keep and which to remove. Do you have any guidance on what can be removed?

> The Git history of the branch looks quite complex. Could you rebase the commits on development and squash them?

Absolutely.

framework/test/gtest/CMakeLists.txt
saschazelzer: Okay, this is fine with me,

saschazelzer:
> I think it will make it more difficult to update the sources in the future (how will anyone know what to add/remove, unless we write it down somewhere). However, the real problem for me is that I don't know which parts are the old docs and old project files. Do you have any guidance on what specific files/folders can be removed?

I thought that the update question would come up. A good compromise could be to add a few rules to our `.gitignore` file. I suggest the following:
```
third_party/googletest
!third_party/googletest/CMakeLists.txt
!third_party/googletest/README.md
!third_party/googletest/googletest/LICENSE
!third_party/googletest/googletest/README.md
!third_party/googletest/googletest/CMakeLists.txt
!third_party/googletest/googletest/cmake
!third_party/googletest/googletest/include
!third_party/googletest/googletest/src
```
This leaves just the files we need right now for build GTest, and updating is simplified because the gitignore rules will filter the files out. If we miss out on something, the CI builds will fail at least.

CMakeLists.txt
```diff
@@ -219,7 +219,8 @@ us_cache_var(DOC_INSTALL_DIR ${default_doc_install_dir} STRING "Relative install
 us_cache_var(MAN_INSTALL_DIR ${default_man_install_dir} STRING "Relative install location for man pages" ADVANCED)
 set(AUXILIARY_CMAKE_INSTALL_DIR ${AUXILIARY_INSTALL_DIR}/cmake)
 
-set(BUILD_SHARED_LIBS ${US_BUILD_SHARED_LIBS})
+set(BUILD_SHARED_LIBS ${US_BUILD_SHARED_LIBS} FORCE)
```
saschazelzer: This should be set in the CMake cache, e.g. by also using `us_cache_var`.

saschazelzer:
Not sure if this is still on your todo list: to get rid of the superfluous gtest files, we need rewrite the commit which added the gtest files to include the `.gitignore` changes and then force-push the branch again. Otherwise, we will carry these gtest files around in the history.

jeffdiclemente:
Ok. I didn't see that the gtest files should be removed in addition to the .gitignore file changes.

Is it sufficient to remove the gtest files in another commit? I'm not familiar with re-writing commits in git.

saschazelzer:
It won't be as clean, because the file contents is still recorded in some git object (you could always go back to the version which still had the full gtest sources). If you want, I can rewrite that single commit which introduced gtest. The author of that commit will stay the same.

jeffdiclemente:
Sure. I'd appreciate it if you could rewrite the commit. Thanks!

jeffdiclemente:
Thanks for fixing the commits Sascha. Is this PR ready to be merged?

saschazelzer:
Looks good to me, thanks for all the effort!

PR #226, opened 2017-08-08
Pass CMAKE_CXX_FLAGS through to the add_custom_command invocation of the compiler in usFunctionEmbedResources.cmake

moraaljj:
Updating usFunctionEmbedResources.cmake to use CMAKE_CXX_FLAGS rather than building the flags itself. This should allow changes to compilation flags in other cmake files to affect the compiled empty resource file for Darwin-based platforms.

### DIFF
--- a/cmake/usFunctionEmbedResources.cmake
+++ b/cmake/usFunctionEmbedResources.cmake
@@ -161,12 +161,12 @@ function(usFunctionEmbedResources)
        # when it should really be:
        #  >> clang ... -c ... -o stub.o
        # (pardon the elipsis for abbreviation)
-       
-       separate_arguments(US_RESOURCE_CXX_FLAGS UNIX_COMMAND ${CMAKE_CXX_FLAGS})
+
+       separate_arguments(_us_resource_cxx_flags UNIX_COMMAND ${CMAKE_CXX_FLAGS})
       # section name is "us_resources" because max length for section names in Mach-O format is 16 characters.
       add_custom_command(
         OUTPUT ${_source_output}
-        COMMAND ${CMAKE_CXX_COMPILER} ${US_RESOURCE_CXX_FLAGS} -c ${US_CMAKE_RESOURCE_DEPENDENCIES_CPP} -o stub.o
+        COMMAND ${CMAKE_CXX_COMPILER} ${_us_resource_cxx_flags} -c ${US_CMAKE_RESOURCE_DEPENDENCIES_CPP} -o stub.o
         COMMAND ${CMAKE_LINKER} -r -sectcreate __TEXT us_resources ${_zip_archive_name} stub.o -o ${_source_output}
         DEPENDS ${_zip_archive}
         WORKING_DIRECTORY ${_zip_archive_path}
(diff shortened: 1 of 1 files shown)


cmake/usFunctionEmbedResources.cmake
```diff
@@ -154,21 +154,22 @@ function(usFunctionEmbedResources)
...
-      endif()
+        # This resolves an issue where passing in CMAKE_CXX_FLAGS using
+       # -D CMAKE_CXX_FLAGS:STRING="..." results in a quoted string being passed to
+       # the command below which results in the command being:
+       #  >> clang "..." -c ... -o stub.o
+       # when it should really be:
+       #  >> clang ... -c ... -o stub.o
+       # (pardon the elipsis for abbreviation)
+       separate_arguments(TEMP_CXX_FLAGS UNIX_COMMAND ${CMAKE_CXX_FLAGS})
+       foreach(flag ${TEMP_CXX_FLAGS})
```
karthikreddy09: Any reason why this for loop is needed? passing the output of separate_arguments to add_custom_command works fine for me.

```
add_custom_command(
        OUTPUT ${_source_output}
        COMMAND ${CMAKE_CXX_COMPILER} ${TEMP_CXX_FLAGS} -c ${US_CMAKE_RESOURCE_DEPENDENCIES_CPP} -o stub.o
        COMMAND ${CMAKE_LINKER} -r -sectcreate __TEXT us_resources ${_zip_archive_name} stub.o -o ${_source_output}
        DEPENDS ${_zip_archive}
        WORKING_DIRECTORY ${_zip_archive_path}
        COMMENT "Linking resources zip file for ${US_RESOURCE_TARGET}"
        VERBATIM
       )
```

moraaljj:
Hi,

The loop is necessary due to an odd behavior with add_custom_command and
passing in CMAKE_CXX_FLAGS using the -D option to cmake.  The loop ensures
that the quotation marks around the arguments are removed so they can be
properly parsed.

-J

cmake/usFunctionEmbedResources.cmake
```diff
@@ -154,21 +154,22 @@ function(usFunctionEmbedResources)
...
-        list(APPEND US_OSX_CXX_FLAGS -mmacosx-version-min=${CMAKE_OSX_DEPLOYMENT_TARGET})
-      endif()
+        # This resolves an issue where passing in CMAKE_CXX_FLAGS using
+       # -D CMAKE_CXX_FLAGS:STRING="..." results in a quoted string being passed to
+       # the command below which results in the command being:
+       #  >> clang "..." -c ... -o stub.o
+       # when it should really be:
+       #  >> clang ... -c ... -o stub.o
+       # (pardon the elipsis for abbreviation)
+       separate_arguments(TEMP_CXX_FLAGS UNIX_COMMAND ${CMAKE_CXX_FLAGS})
```
saschazelzer: Style: please use `_temp_cxx_flags`

cmake/usFunctionEmbedResources.cmake
```diff
@@ -154,21 +154,22 @@ function(usFunctionEmbedResources)
...
+       foreach(flag ${TEMP_CXX_FLAGS})
+         set(US_RESOURCE_CXX_FLAGS ${US_RESOURCE_CXX_FLAGS} ${flag})
```
saschazelzer: Style: please use `_resource_cxx_flags`

moraaljj:
Hi, I updated and removed the for loop as suggested by Abhinay.  It appears to have the same effect and cmake appears to expand the list automatically in the add_custom_command call.

saschazelzer:
Thanks!

Issue #227, opened 2017-08-14
LDAPFilter::Match throws std::runtime_error if a bundle contains a key that is a leading substring of another key


jeffdiclemente:
Calling LDAPFilter::Match on a bundle that contains the keys: 

```
"apple" : "granny smith" 
"applesauce" : "yum" 
```

throws a std::runtime_error when creating an instance of cppmicroservices::Properties because Properties::Find_unlocked only does a strncasecmp of key.size() between key and keys[i]. 

Properties::Find_unlocked should check that the sizes of the two strings are equal before reporting duplicate keys and any exceptions thrown by Properties should be swallowed by LDAPFilter::Match before returning false since Match isn't supposed to throw exceptions.

ksubramz:
The way this bug manifests is worth documenting - it's dependent on 
 - if a key is a substring of another key, and 
 - the AnyMap iteration order in `Properties` constructor.. specifically the line:

`
  for (auto& iter : p)
`
in `Properties::Properties(const AnyMap& p)`

Suppose the range-based iterator gives out the key `apple` first and `applesauce` next, there is no exception triggered. However, if the iterator gives out the key `applesauce` first and then the key `apple`, the `Properties::Find_unlocked` call in the constructor does a strncasecmp of the key in question `apple` and the first 5 characters of `applesauce`, which is `apple` and throws an exception because they are equal.

It turns out I can't reproduce the bug with the keys specified in the Issue, however I can reproduce the bug with different sets of keys i.e. `hosed`, `hosedd`, `hose`

ksubramz:
Merged.

PR #228, opened 2017-08-17
Fix LDAPFilter::Match bug

ksubramz:
Addresses #227

### DIFF
--- a/framework/include/cppmicroservices/LDAPFilter.h
+++ b/framework/include/cppmicroservices/LDAPFilter.h
@@ -130,10 +130,8 @@ public:
    *        in the match.
    * @return <code>true</code> if the bundle's properties match this
    *         <code>LDAPFilter</code> <code>false</code> otherwise.
-   * @throws std::runtime_error If the number of keys of the <code>bundle</code>
+   * @throws std::runtime_error If the number of keys of the bundle's properties
    *         exceeds the value returned by std::numeric_limits<int>::max().
-   * @throws std::runtime_error If the <code>bundle</code> contains case variants
-   *         of the same key name.
    */
   bool Match(const Bundle& bundle) const;
 
@@ -146,7 +144,7 @@ public:
    *        in the match.
    * @return <code>true</code> if the <code>AnyMap</code>'s values match this
    *         filter; <code>false</code> otherwise.
-   * @throws std::runtime_error If the number of keys of the <code>dictionary</code>
+   * @throws std::runtime_error If the number of keys in the <code>dictionary</code>
    *         exceeds the value returned by std::numeric_limits<int>::max().
    * @throws std::runtime_error If the <code>dictionary</code> contains case variants
    *         of the same key name.
@@ -162,7 +160,7 @@ public:
    *        in the match.
    * @return <code>true</code> if the <code>AnyMap</code>'s values match this
    *         filter; <code>false</code> otherwise.
-   * @throws std::runtime_error If the number of keys of the <code>dictionary</code>
+   * @throws std::runtime_error If the number of keys in the <code>dictionary</code>
    *         exceeds the value returned by std::numeric_limits<int>::max().
    * @throws std::runtime_error If the <code>dictionary</code> contains case variants
    *         of the same key name.
--- a/framework/include/cppmicroservices/LDAPFilter.h
+++ b/framework/include/cppmicroservices/LDAPFilter.h
@@ -120,18 +120,18 @@ public:
   bool Match(const ServiceReferenceBase& reference) const;
 
  /**
-   * Filter using a bundle's manifest properties.
+   * Filter using a bundle's manifest headers.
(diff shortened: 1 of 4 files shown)


ksubramz:
Addressing Jeff's comment, the reason all the `Match` methods are throwing because they all construct temporary `Properties` objects. And both the exceptions may get thrown in the constructor of `Properties`.

jeffdiclemente:
alright, closing the loop; I talked to Krish offline and we resolved my confusion on why some ```match``` methods throw and others don't. Its because a throw happens in the ```BundleContext::RegisterService``` method if case variants are found.

framework/include/cppmicroservices/LDAPFilter.h
```diff
@@ -130,6 +130,10 @@ class US_Framework_EXPORT LDAPFilter {
    *        in the match.
    * @return <code>true</code> if the bundle's properties match this
    *         <code>LDAPFilter</code> <code>false</code> otherwise.
+   * @throws std::runtime_error If the number of keys of the <code>bundle</code>
```
jeffdiclemente: the keys are part of the bundle's properties and not the bundle itself.

framework/include/cppmicroservices/LDAPFilter.h
```diff
@@ -130,6 +130,10 @@ class US_Framework_EXPORT LDAPFilter {
...
+   *         exceeds the value returned by std::numeric_limits<int>::max().
+   * @throws std::runtime_error If the <code>bundle</code> contains case variants
```
jeffdiclemente: the keys are part of the bundle's properties and not the bundle itself.

framework/include/cppmicroservices/LDAPFilter.h
```diff
@@ -142,6 +146,10 @@ class US_Framework_EXPORT LDAPFilter {
    *        in the match.
    * @return <code>true</code> if the <code>AnyMap</code>'s values match this
    *         filter; <code>false</code> otherwise.
+   * @throws std::runtime_error If the number of keys of the <code>dictionary</code>
```
jeffdiclemente: > If the number of keys of the <code>dictionary</code>

sounds a little awkward. I'd suggest:

> If the number of keys in the <code>dictionary</code>

framework/include/cppmicroservices/LDAPFilter.h
```diff
@@ -154,6 +162,10 @@ class US_Framework_EXPORT LDAPFilter {
...
```
jeffdiclemente: I'd suggest:

> If the number of keys in the <code>dictionary</code>

framework/include/cppmicroservices/LDAPFilter.h
saschazelzer: This can never happen, since the Match() function operates on the bundles headers, which is a case insensitive map already.

framework/include/cppmicroservices/LDAPFilter.h
```diff
@@ -130,10 +130,8 @@ class US_Framework_EXPORT LDAPFilter {
    *        in the match.
    * @return <code>true</code> if the bundle's properties match this
    *         <code>LDAPFilter</code> <code>false</code> otherwise.
-   * @throws std::runtime_error If the number of keys of the <code>bundle</code>
+   * @throws std::runtime_error If the number of keys of the bundle's properties
```
saschazelzer: We need to be careful about the terminology here. It is the bundle's manifest headers that are checked and matched. The return doc above also users the wrong properties term.

framework/include/cppmicroservices/LDAPFilter.h
```diff
@@ -126,12 +126,12 @@ class US_Framework_EXPORT LDAPFilter {
...
-   * @param bundle The bundle whose properties are used
+   * @param bundle The bundle whose manifest's properties are used
    *        in the match.
-   * @return <code>true</code> if the bundle's properties match this
+   * @return <code>true</code> if the bundle's manifest properties match this
    *         <code>LDAPFilter</code> <code>false</code> otherwise.
-   * @throws std::runtime_error If the number of keys of the bundle's properties
-   *         exceeds the value returned by std::numeric_limits<int>::max().
+   * @throws std::runtime_error If the number of keys of the bundle's manifest
+   *         properties exceeds the value returned by std::numeric_limits<int>::max().
```
saschazelzer: I am sorry, but elsewhere we really use the term *headers* (the `Match()` implementation really calls `bundle.GetHeaders()`) for the manifest data, not *properties*. properties are accessed via the bundle context and are framework properties. *manifest properties* now is even more confusing in my opinion.

ksubramz: No issues :). The reason I chose the wordings was because the short blurb of this API method says `Filter using a bundle's manifest properties`. I'll change as per your feedback.

Issue #235, opened 2017-08-24
Improve code coverage


ksubramz:
This is an umbrella issue to track the work done to improve the code coverage of our test suite. (Landing page: https://codecov.io/gh/cppmicroservices/CppMicroServices)

rasane:
Hello, I Would like to help.. I added 1 test and confirmed it ran successfully and code path was executed. I Wanted to confirm if it affects code metrics, how can I do that? 

Apart from raising a pr and running code coverage at about link, how can I set up on my personal branch?
Ps: haven't forked officially yet as i m trying to understand my capability to contribute positively. Let me know if I missed some obvious note, apologies in advance.

Thanks, rasane

jeffdiclemente:
Hi rasane,

Thanks for writing a test!

There are a few ways you could determine whether the test affects code metrics...
### Manual Inspection
Use the codecov.io link above to navigate to the source file(s) your test uses. Visually inspect which lines in the source aren't hit by a test and then check your test code to see if it walks any of those code paths. If it does, you can be sure that the test improves code coverage.

### Run gcov and lcov locally on your machine
In a nutshell, gcov and lcov are used to gather and create the code coverage data. You can always run these manually on a Linux box.

### Fork CppMicroServices and send a PR
This is typically very quick and easy and doesn't require running tools manually.

rasane:
Thanks for the detailed explanation. I don't have a Linux box at the moment, will install one this weekend.. I haven't used gcov/lcov before, sound like handy tools..

PR #236, opened 2017-08-29
Improve code coverage of LDAP* and AnyMap

ksubramz:
Addresses part of the umbrella issue #235 

For the following sources the coverage is expanded to all the reachable lines, which results in the coverage percentage increase (compared to the "development" branch) indicated by the numbers next to the files.
AnyMap.cpp : 58.18 -> 93.81%
LDAPExpr.cpp : 72.12 -> 97.77%
LDAPProp.cpp :  72.52 ->100.00 %
LDAPFilter.cpp : 98.36 -> 98.36 %


saschazelzer:
Because the branch history is not that nice to look at, I recommend to squash it this time before merging.

framework/src/util/LDAPExpr.cpp
```diff
@@ -331,11 +331,6 @@ bool LDAPExpr::IsNull() const
   return !d;
 }
 
-bool LDAPExpr::Query( const std::string& filter, const PropertiesHandle& pd)
```
jeffdiclemente: I don't remember, why was this function removed?

ksubramz: This function was removed because it's unused in our code (and it's not a public interface)

Issue #220, opened 2017-07-12
[Linux] Test failure caused by temporary directory name collisions


jeffdiclemente:
I recently saw a use case whereby a BundleTest test point failed while creating the framework's temporary storage directory. 
The failure was a seg fault and when it was traced back, the fault was that the temp storage directory already existed with permissions that did not allow r/w access to the current user. 
For example, ```/tmp/data``` already existed and the "local" sub directory couldn't be created because ```/tmp/data``` wasn't r/w for the user running the test (the machine is used by many different users).

For Linux, the test code should use a standard POSIX system API, like [mkdtemp](http://pubs.opengroup.org/onlinepubs/9699919799/functions/mkdtemp.html) instead of querying the ```TMPDIR``` env var. This helps to avoid directory name collisions.

saschazelzer:
I am going to fix this as part of my work on #191.

Issue #191, opened 2017-03-11
GetLastErrorStr() uses thread-unsafe strerror function


saschazelzer:
On POSIX systems, the GetLastErrorStr() function internally uses `strerror` which is not thread-safe. We should use `strerror_r` instead.

Issue #126, opened 2016-08-12
Rework GlobalConfig.h.in


lygstate:
US_BUILD_SHARED_LIBS can be added as a compiler option.

US_ENABLE_THREADING_SUPPORT default to ON, and we can disable it by compiler option.
And should not appear in GlobalConfig.in.h.
And it's should only affect Cpp files and not in headers.

US_HAVE_VISIBILITY_ATTRIBUTE can be detect by compiler version.

US_HAVE_CXXABI_H should be always exist beyond Windows system.

```
#define CppMicroServices_MAJOR_VERSION @CppMicroServices_MAJOR_VERSION@
#define CppMicroServices_MINOR_VERSION @CppMicroServices_MINOR_VERSION@
#define CppMicroServices_PATCH_VERSION @CppMicroServices_PATCH_VERSION@
#define CppMicroServices_VERSION @CppMicroServices_VERSION@
#define CppMicroServices_VERSION_STR "@CppMicroServices_VERSION@"

These could be able to export as a function. at expose it at the runtime.
Beyond that, Windows also need to include these in it's .rc files.
```
# cmakedefine US_BIG_ENDIAN
# cmakedefine US_LITTLE_ENDIAN

Should be detected at the runtime. Or, if we are cross-compiling, things gonna be wrong.

US_HAVE_REGEX are used for test. We could use the newest version of C++ compiler to do test.
And remove it from GlobalConfig.in.h

saschazelzer:
Thank you for your suggestions. I am curious, what is your motivation to suggest these changes?

Some of the defines you mention are useful for client code (e.g. `US_BUILD_SHARED_LIBS` and the version macros) during compile time, so I'd want to make those available. Not having them in a generated header means that clients either need to remember adding them manually or we rely on the CMake files, which ties clients to it even more. Or did I miss something?

I agree with the `US_ENABLE_THREADING_SUPPORT` comment.

For `US_HAVE_VISIBILITY_ATTRIBUTE` and `US_HAVE_CXXABI_H`, I prefer to rely on feature tests instead of hard coding compiler versions. The latter one is used only internally and we can move it out of GlobalConfig.h.in.

The version macros are useful to conditionally compile client code depending on the version number. Making them functions would prevent this.

The `US_BIG_ENDIAN` (and little variant) macros are determined using compile tests. So this should be correct already in a cross-compilation setup.

I agree, `US_HAVE_REGEX` should be removed from the header file.

[...]

PR #238, opened 2017-09-03
System error code handling and utiltity function cleanup

saschazelzer:
This PR addresses issues #191 and #220 and also contains further cleanup.

The first commit contains the actual fixes:

- Correctly dinstinguish between C runtime library errors and Win32 API errors
- Set `errno` to 0 before calling C runtime library functions
- Use thread-safe `strerrror` variants if available

It also improves code sharing between the legacy and the new GTest test executable:

- Pull out `TestUtils.cpp` for re-use in legacy and GTest tests
- `TestUtils.cpp` changes:
  - Use temporary directory instead of "ProgramData" path on Windows
  - Added cross-platform `mkdtemp` and `mkstemp` functions for creating unique files and directories (originated from public domain and contributed by me with slight changes to another project a couple of years ago; now just copied here)
  - Added helper functions for creating temporary files and directories

The commit also adds unit tests for our internal filesystem related functions:

- Create tests for `cppmicroservices::fs` functions. The tests are compiled for static builds only to avoid exporting internal functions. 
- Make the gtest-based test executable a proper bundle, in order to compile `TestUtils.cpp`.

For the above changes, I had to duplicate a little more code from the internal file-system code. Seeing that we then had almost the same content duplicated in the framework and the unit test code I wanted to get rid of that and make it reusable.

This is what the second commit contains:

- Create an *object library* called "util" that contains framework independent utility code chared between bundles and tests. This includes file-system related functions, error handling routines, string manipulation routines.
- Add the object library to all bundles and the two test drivers by default

Please note that virtually no new code was written, except for disentangling our C / Win32 error handling, using `strerror_r` etc. where appropriate, and writing new unit tests (GTest based).


framework/src/bundle/CoreBundleContext.cpp
```diff
@@ -149,12 +151,19 @@ void CoreBundleContext::Init()
...
+  {
+    auto const execPath = util::GetExecutablePath();
+    if (bundleRegistry.GetBundles(execPath).empty() &&
+        IsBundleFile(execPath))
+    {
+      // auto-install all embedded bundles inside the executable
+      bundleRegistry.Install(execPath, systemBundle.get());
+    }
+  }
+  catch (const std::exception& e)
```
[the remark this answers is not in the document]
saschazelzer: Good point. That was not intended.

util/src/Error.cpp
```diff
@@ -0,0 +1,129 @@
...
+  #include <crtdbg.h>
+#endif
+
+namespace cppmicroservices {
+
+namespace util {
+
+std::string GetLastWin32ErrorStr()
+{
+#ifdef US_PLATFORM_WINDOWS
```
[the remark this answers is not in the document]
saschazelzer: I will look at that.

util/src/Error.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+std::string GetLastCErrorStr()
+{
+  char errorString[128];
+#if ((_POSIX_C_SOURCE >= 200112L) && !  _GNU_SOURCE) || defined(US_PLATFORM_APPLE)
+  // This is the XSI strerror_r version
+  if (strerror_r(errno, errorString, sizeof errorString))
+  {
+    return "Unknown error";
```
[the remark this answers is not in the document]
saschazelzer: We could return `errno` as a number in the returned string (by saving it before calling `strerror_r`). I think returning the error code from a failing `strerror_r` call would be misleading, because it will mask the original error for which we wanted to retrieve the textual description.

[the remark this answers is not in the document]
jeffdiclemente: Sorry, I meant returning a string indicating why strerror_r failed. 
According to the documentation for strerror_r it only returns a failure if errno is invalid or the string buffer isn't big enough.

util/src/Error.cpp
```diff
@@ -0,0 +1,129 @@
...
+  }
+  return errorString;
+#else
+  return strerror_r(errno, errorString, sizeof errorString);
+#endif
+}
+
+US_MSVC_PUSH_DISABLE_WARNING(4715) // 'function' : not all control paths return a value
+std::string GetExceptionStr(const std::exception_ptr& exc)
```
karthikreddy09: The warning push pop can be removed with the following block.

```
std::string GetExceptionStr(const std::exception_ptr& exc)
{
  std::string exceptionStr;
  try
  {
    if(exc)
    {
      std::rethrow_exception(exc);
    }
  }
  catch (const std::exception& e)
  {
    exceptionStr = e.what();
  }
  catch (...)
  {
    exceptionStr=  "unknown";
  }
  return exceptionStr;
}
```

util/src/FileSystem.cpp
```diff
@@ -0,0 +1,332 @@
...
+    if (getcwd(buf.data(), bufSize) != 0 && errno != ERANGE)
+    {
+      return std::string(buf.data());
+    }
+  }
+#endif
+  return std::string();
+}
+
+static const std::string s_CurrWorkingDir = InitCurrentWorkingDirectory();
```
karthikreddy09: Why not make this magic static inside the function?

util/src/Error.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include <cppmicroservices/GlobalConfig.h>
+
+#ifdef US_PLATFORM_POSIX
+  #include <errno.h>
+  #include <string.h>
+
+#else
+  #ifndef WIN32_LEAN_AND_MEAN
+    #define WIN32_LEAN_AND_MEAN
```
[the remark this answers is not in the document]
saschazelzer: I think we need #126 resolved and have an *internal* global `.h` file (which is not included in public `.h` files) where we can place such defines.

util/src/FileSystem.cpp
saschazelzer: This code was just moved from a different file and not changed. We had a discussion about the actual code in the original PR. Basically, we need to call `InitCurrentWorkingDirectory` at a time when there is only the main thread running (during static init).

util/src/Error.cpp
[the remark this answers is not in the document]
jeffdiclemente: that sounds good to me.

util/src/FileSystem.cpp
jeffdiclemente: @karthikreddy09 take a look at PR #216 for more information.

karthikreddy09: Ah ... now I understand. Thanks for pointing to the previous discussion.

framework/src/bundle/CoreBundleContext.cpp
```diff
@@ -151,19 +151,28 @@ void CoreBundleContext::Init()
...
-    {
-      // auto-install all embedded bundles inside the executable
-      bundleRegistry.Install(execPath, systemBundle.get());
-    }
+    execPath = util::GetExecutablePath();
   }
   catch (const std::exception& e)
   {
     DIAG_LOG(*sink) << e.what();
+    // Let the exeception propagate all the way up to the
```
jeffdiclemente: typo: "exeception"

framework/test/util/TestUtils.cpp
```diff
@@ -0,0 +1,490 @@
...
+    }
+  }
+
+  /* We got out of the loop because we ran out of combinations to try.  */
+  errno = EEXIST;
+  return -1;
+}
+
+// A cross-platform version of the POSIX mkdtemp function
+char* mkdtemps_compat(char *tmpl, int suffixlen)
```
ksubramz: It looks like there's a lot of common code between mkdtemps_compat and mkstemps_compat that can be factored out to remove the duplication?

framework/test/util/TestUtils.cpp
```diff
@@ -0,0 +1,490 @@
...
+  return std::string(temp_dir.cbegin(), temp_dir.cend());
+#else
+  char* tempdir = getenv("TMPDIR");
+  return std::string(((tempdir == nullptr)?"/tmp":tempdir));
+#endif
+}
+
+static const char validLetters[] = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+
+// A cross-platform version of the mkstemps function
```
ksubramz: Sascha, curious about why we need to be compatible with mkstemp (or mkdtemp), AFAIU, these are only called by our tests and if we drop the compatibility criterion, we could avoid passing "char *" and avoid using strlen() etc. i.e. operate on std::string's instead of (less safe) C-style strings, use std::string methods and throw an exception from this function itself in-case of any error.

saschazelzer: The answer is a historical one. The originates from the public domain on the Internet. As such it was C code, written to be compatible witk mkstemp and friends by the original author. I only did very small changes originally (many years ago) and used the code as is in another project. It did its job well enough and I thought I would spend my time elsewhere instead of making it prettier for CppMicroServices.

Feel free to send a PR for C++ification or other improvements, but I think there is not much gain unless there is a bug of course.

framework/test/util/TestUtils.cpp
saschazelzer: Mostly for the same reason as above. Laziness, working on bugs or features, afraid of introducing regressions, ...

framework/test/util/TestUtils.cpp
ksubramz: Thanks. Sounds good.

PR #242, opened 2017-10-11
Fix IsConvertibleTo bug

### LINKED ISSUE #240: IsConvertibleTo method doesn't check for validity of member
## The Issue
The method in question:
```cpp
bool ServiceReferenceBasePrivate::IsConvertibleTo(const std::string& interfaceId) const
{
  return registration ? registration->Lock(), registration->service->find(interfaceId) != registration->service->end() : false;
}
```
The member `service` of `registration` is an `InterfaceMapConstPtr`, which could be `reset()` by another thread calling Unregister() on the `ServiceRegistration` object [[1](https://github.com/CppMicroServices/CppMicroServices/blob/development/framework/src/service/ServiceRegistrationBase.cpp#L260)]. When the current thread dereferences `service` i.e. `registration->service->find()`, it results in a crash (access violation).

## Minimal Test Case

Minimal test case to illustrate the crash

```cpp
void TestUnregisterRace(BundleContext context)
{
  struct TestServiceA : public ITestServiceA
  {
  };

  auto s1 = std::make_shared<TestServiceA>();

  std::atomic_bool done(false);
  auto copyRefs = [](ServiceReference<ITestServiceA> ref, std::atomic_bool& done) {
	  while (!done)
	  {
	      (void)ref.IsConvertibleTo("IBooService");
	  }
  };
  
  ServiceRegistration<ITestServiceA> registration = context.RegisterService<ITestServiceA>(s1);
  ServiceReference<ITestServiceA> reference = context.GetServiceReference<ITestServiceA>();
  std::thread thread(copyRefs, reference, std::ref(done));
  std::this_thread::sleep_for(std::chrono::milliseconds(10));
  registration.Unregister();
  std::this_thread::sleep_for(std::chrono::milliseconds(10));
  done = true;
  thread.join();
}
```

* The main thread registers a service with ServiceReference reference, spawns a thread copyRefs, waits for a bit and calls Unregister() on the ServiceRegistration object registration.
* The spawned thread keeps calling IsConvertibleTo() on reference object until the main thread flips the bool "done", which in turn signals the thread to stop.
* The crash happens as soon as the Unregister() is invoked on the registration object.

## Fix Proposal

The fix simply checks for the validity of `service` member before dereferencing it.

```cpp
bool ServiceReferenceBasePrivate::IsConvertibleTo(const std::string& interfaceId) const
{
  if (registration)
  {
	  auto l = registration->Lock(); US_UNUSED(l);
	  return registration->service ? registration->service->find(interfaceId) != registration->service->end() : false;
  }
  return false;
}
```
This has been tested and the minimal test case program runs successfully.

### A tangential question
In general use case, it seems to me that the check for `registration` in `IsConvertibleTo` will always evalute to `True` because it's a non-null pointer. Is my assumption correct? Is there a case where `registration` is nullptr?

ksubramz:
Fix #240

### DIFF
--- a/framework/src/service/ServiceReferenceBasePrivate.cpp
+++ b/framework/src/service/ServiceReferenceBasePrivate.cpp
@@ -361,7 +361,12 @@ PropertiesHandle ServiceReferenceBasePrivate::GetProperties() const
 
 bool ServiceReferenceBasePrivate::IsConvertibleTo(const std::string& interfaceId) const
 {
-  return registration ? registration->Lock(), registration->service->find(interfaceId) != registration->service->end() : false;
+  if (registration)
+  {
+	  auto l = registration->Lock(); US_UNUSED(l);
+	  return registration->service ? registration->service->find(interfaceId) != registration->service->end() : false;
+  }
+  return false;
 }
 
 }
--- a/framework/src/service/ServiceReferenceBasePrivate.cpp
+++ b/framework/src/service/ServiceReferenceBasePrivate.cpp
@@ -363,8 +363,8 @@ bool ServiceReferenceBasePrivate::IsConvertibleTo(const std::string& interfaceId
 {
   if (registration)
   {
-	  auto l = registration->Lock(); US_UNUSED(l);
-	  return registration->service ? registration->service->find(interfaceId) != registration->service->end() : false;
+    auto l = registration->Lock(); US_UNUSED(l);
+    return registration->service ? registration->service->find(interfaceId) != registration->service->end() : false;
   }
   return false;
 }
--- a/framework/test/driver/ServiceRegistryTest.cpp
+++ b/framework/test/driver/ServiceRegistryTest.cpp
@@ -91,7 +91,7 @@ void TestUnregisterFix(BundleContext context)
   {
   };
 
-  std::atomic_bool done(false);
+  std::atomic_bool done = ATOMIC_VAR_INIT(false);
   auto invokeIsConvertibleTo = [](ServiceReference<ITestServiceA> ref, std::atomic_bool& done) {
     while (!done)
     {
--- a/framework/test/driver/ServiceRegistryTest.cpp
+++ b/framework/test/driver/ServiceRegistryTest.cpp
@@ -31,6 +31,9 @@
 #include "TestingMacros.h"
 
 #include <stdexcept>
+#include <chrono>
+#include <thread>
+#include <atomic>
 
 using namespace cppmicroservices;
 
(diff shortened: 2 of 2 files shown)


framework/test/driver/ServiceRegistryTest.cpp
```diff
@@ -80,6 +83,32 @@ void TestMultipleServiceRegistrations(BundleContext context)
   US_TEST_CONDITION_REQUIRED(!ref, "Testing for invalid service reference")
 }
 
+// Spawned thread repeatedly calls the IsConvertibleTo method and when the main thread
+// unregisters the service, it used to result in a crash.
+void TestUnregisterFix(BundleContext context)
```
saschazelzer: This function should be guarded by a `#ifdef US_ENABLE_THREADING_SUPPORT`. One of the Travis test failures is related to the code executing without the synchronization primitives "enabled".

ksubramz: Thanks Sascha. I guarded this test with a US_ENABLE_THREADING_SUPPORT and in the `#else` part, I have the single-threaded test. (And yes, the bug manifested before the fix and not after the fix.)

framework/test/driver/ServiceRegistryTest.cpp
```diff
@@ -80,6 +83,46 @@ void TestMultipleServiceRegistrations(BundleContext context)
...
+  ServiceReference<ITestServiceA> reference = context.GetServiceReference<ITestServiceA>();
+  std::thread thread(invokeIsConvertibleTo, reference, std::ref(done));
+  std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  registration.Unregister();
+  std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  done = true;
+  thread.join();
+}
+#else
+void TestUnregisterFix(BundleContext context)
```
saschazelzer: What I meant is that the original bug you are fixing (the `registration->service` nullptr access) is not specific to a multi-threaded test case. We do not need multiple threads to test the correct behavior, but the simple test below is enough. So the multi-threaded version above could be removed completely.

framework/test/driver/ServiceRegistryTest.cpp
```diff
@@ -80,6 +83,46 @@ void TestMultipleServiceRegistrations(BundleContext context)
...
+{
+  struct TestServiceA : public ITestServiceA
```
saschazelzer: We do not need a struct here. Below, you can just do
```
context.RegisterService<int>(std::make_shared<int>());
```

framework/test/driver/ServiceRegistryTest.cpp
```diff
@@ -80,6 +83,46 @@ void TestMultipleServiceRegistrations(BundleContext context)
...
+  {
+  };
+
+  ServiceRegistration<ITestServiceA> registration = context.RegisterService<ITestServiceA>(std::make_shared<TestServiceA>());
+  ServiceReference<ITestServiceA> reference = context.GetServiceReference<ITestServiceA>();
+  registration.Unregister();
+  (void)reference.IsConvertibleTo("IBooService");
```
saschazelzer: Why not check that the return value is `false`?

framework/test/driver/ServiceRegistryTest.cpp
ksubramz: Sure.. The reason for two threads test was that it represented a crash we were encountering internally (multi-threaded use case). I'll make the changes.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -361,7 +361,12 @@ PropertiesHandle ServiceReferenceBasePrivate::GetProperties() const
 
 bool ServiceReferenceBasePrivate::IsConvertibleTo(const std::string& interfaceId) const
 {
-  return registration ? registration->Lock(), registration->service->find(interfaceId) != registration->service->end() : false;
+  if (registration)
+  {
+	  auto l = registration->Lock(); US_UNUSED(l);
+	  return registration->service ? registration->service->find(interfaceId) != registration->service->end() : false;
```
jeffdiclemente: fix spacing

saschazelzer:
What is the status of this PR? It looks like it can be merged.

PR #244, opened 2017-10-13
182 bundle manifest validation

### LINKED ISSUE #182: The resource compiler should perform at least minimal validation of a manifest prior to embedding
The resource compiler should perform at least minimal validation (perhaps simple syntax checking) of JSON manifests prior to embedding them into bundles. At it stands right now, even simple typos in a manifest result in total runtime failure of the product-- see #180. Even when #180 is fixed, it is still extremely wasteful to not be able to detect basic issues with manifest files until runtime.

jeffdiclemente:
Fixes #182


framework/doc/bundle_properties.rst
```diff
@@ -58,9 +59,18 @@ Services library specifies the following standard keys for re-use in
    :any:`GetSymbolicName() <cppmicroservices::Bundle::GetSymbolicName>` or
    :any:`GetVersion() <cppmicroservices::Bundle::GetVersion>`.
 
+.. attention::
+
+   JSON is case-sensitive. However C++ Micro Services stores bundle properties in
+   a case in-sensitive way to accommodate :any:`LDAPFilter <cppmicroservices::LDAPFilter>`
```
ksubramz: "case in-sensitive" -> "case-insensitive"

framework/test/driver/ResourceCompilerTest.cpp
```diff
@@ -422,28 +536,569 @@ static void testrcFailureModes(const std::string& rcbinpath, const std::string&
...
+
+  std::ostringstream cmd;
+  cmd << rcbinpath;
+  cmd << " --bundle-name " << "mybundle";
+  cmd << " --out-file " << tempdir << "json_comment_syntax.zip";
+  cmd << " --manifest-add " << tempdir << "manifest.json";
+  US_TEST_CONDITION(EXIT_SUCCESS == runExecutable(cmd.str()), "Test embedding a manifest containing JSON comments.");
+}
+
+void testManifestAddWithDuplicateKeys(const std::string& rcbinpath, const std::string& tempdir)
```
ksubramz: A lot of common code between this test and testManifestAddWithInvalidJSON (considering you are testing for the same failure modes for the same commands).  maybe attempt factoring out the common code to remove the duplication?

framework/test/driver/ResourceCompilerTest.cpp
```diff
@@ -422,28 +536,569 @@ static void testrcFailureModes(const std::string& rcbinpath, const std::string&
...
+  cmd4 << " --res-add manifest.json";
+
+  testing::ChangeDirectory(tempdir);
+  US_TEST_CONDITION(BUNDLE_MANIFEST_VALIDATION_ERROR_CODE == runExecutable(cmd4.str()), "Fail to embed manifest containing duplicate JSON key names.");
+  testing::ChangeDirectory(origdir);
+  US_TEST_CONDITION(false == containsBundleZipFile(tempdir + duplicate_keys_res_add_bundle_file), "Test that the invalid manifest.json file was not added");
+}
+
+// A zip file containing only a manifest.
+// The zip file is deleted when this object is destroyed.
```
ksubramz: `// The zip file is deleted when this object is destroyed.`. the deletion code is missing from the destructor?

framework/test/driver/ResourceCompilerTest.cpp
```diff
@@ -422,28 +536,569 @@ static void testrcFailureModes(const std::string& rcbinpath, const std::string&
...
+  cmd << " --manifest-add " << tempdir << "manifest_part2.json";
+  cmd << " --manifest-add " << tempdir << "duplicate_manifest.json";
+  cmd << " --manifest-add " << tempdir << "manifest_part3.json";
+  US_TEST_CONDITION(BUNDLE_MANIFEST_VALIDATION_ERROR_CODE == runExecutable(cmd.str()), "Test that a duplicate manifest json part fails to embed the manifest.");
+
+  ZipFile duplicate_merged_zip(tempdir + "duplicate_merged_zip.zip");
+  US_TEST_CONDITION(0 == duplicate_merged_zip.size(), "Test that no manifest.json was embedded since there was an duplicate manifest part.");
+}
+
+// return the contents of the manifest file for the giveen bundle_name
```
ksubramz: "giveen" -> "given"

framework/test/driver/ResourceCompilerTest.cpp
```diff
@@ -422,28 +536,569 @@ static void testrcFailureModes(const std::string& rcbinpath, const std::string&
...
+std::string manifestContent(const std::string& zipfile, std::string bundle_name)
```
ksubramz: "manifestContent" -> "getManifestContent" ?

framework/test/driver/ResourceCompilerTest.cpp
```diff
@@ -422,28 +536,569 @@ static void testrcFailureModes(const std::string& rcbinpath, const std::string&
...
+  std::cout << "Expected JSON:\n\n" << expectedJSON << "JSON concatenated by usResourceCompiler:\n\n" << concatenatedJSON << std::endl;
+
+  // line feed and new line characters may be lurking in the strings. I don't know how to use miniz and jsoncpp to embed
+  // these characters in a consistent manner, so I'm opting to remove them afterwards.
+  concatenatedJSON.erase(std::remove(concatenatedJSON.begin(), concatenatedJSON.end(), '\r'), concatenatedJSON.end());
+  concatenatedJSON.erase(std::remove(concatenatedJSON.begin(), concatenatedJSON.end(), '\n'), concatenatedJSON.end());
+  expectedJSON.erase(std::remove(expectedJSON.begin(), expectedJSON.end(), '\r'), expectedJSON.end());
+  expectedJSON.erase(std::remove(expectedJSON.begin(), expectedJSON.end(), '\n'), expectedJSON.end());
+  
+  US_TEST_CONDITION(0 == concatenatedJSON.compare(expectedJSON), "Test that the concatenated JSON content matches the expected JSON content.");
```
ksubramz: maybe `concatenedJSON == expectedJSON` ? :)

tools/rc/ResourceCompiler.cpp
```diff
@@ -110,6 +112,178 @@ std::string us_tempfile()
...
+
+/*
+ * @brief parses json content and returns the parsed json or throws.
+ * @tparam jsonContent json content to parse. The type must be convertible to std::istream.
+ * @param root The parsed Json root object.
+ * @throw InvalidManifest if the json is invalid. Parse error information is in the exception.
+ * If an exception is thrown, the root param is invalid.
+ */
+template <class T>
+void parseAndValidateJson(T& jsonContent, Json::Value& root)
```
ksubramz: Will `const T& jsonContent` work here?

tools/rc/ResourceCompiler.cpp
```diff
@@ -520,10 +783,27 @@ int main(int argc, char** argv)
       }
 
       std::unique_ptr<ZipArchive> zipArchive(new ZipArchive(zipFile, compressionLevel, bundleName));
+      
+      // map of manifest file to its JSON data
+      std::map<std::string, Json::Value> manifests;
```
ksubramz: `std::map` -> `std::unordered_map` ?

tools/rc/ResourceCompiler.cpp
```diff
@@ -110,6 +112,178 @@ std::string us_tempfile()
...
+    throw InvalidManifest(exceptionMsg);
+  }
+}
+
+/* 
+ * @brief Validate manifest files in an archive.
+ * @param archiveFile archive file path
+ * @throw std::InvalidManifest on the first invalid manifest found.
+ */
+void validateManifestsInArchive(const std::string archiveFile)
```
ksubramz: Probably doesn't matter at all but to be consistent with your other signatures

`const std::string archiveFile` -> `const std::string& archiveFile`

framework/src/bundle/BundleManifest.cpp
```diff
@@ -74,14 +74,14 @@ Any ParseJsonValue(const Json::Value& jsonValue, bool ci)
...
   else if (jsonValue.isIntegral())
   {
     return Any(jsonValue.asInt());
   }
+  else if (jsonValue.isDouble())
+  {
+    return Any(jsonValue.asDouble());
+  }
 
   return Any();
```
karthikreddy09: Unrelated to this change but the number of return statements in this function makes me uncomfortable.

framework/test/driver/ResourceCompilerTest.cpp
```diff
@@ -37,23 +39,34 @@
...
-*/
-static void createDirHierarchy(const std::string& tempdir)
+ * Execute a process with the given command line
+ * @param executable absolute path to the executable to run along with
+ *  and any command line parameters.
+ * @return the executable's return code.
+ */
+int runExecutable(const std::string& executable)
+{
+// WEXITSTATUS is only available on POSXI. Wrap std::system into a function
```
karthikreddy09: typo, POSIX

framework/test/driver/ResourceCompilerTest.cpp
```diff
@@ -422,28 +536,569 @@ static void testrcFailureModes(const std::string& rcbinpath, const std::string&
...
+  cmd << " --manifest-add " << tempdir << "manifest.json";
+  US_TEST_CONDITION(BUNDLE_MANIFEST_VALIDATION_ERROR_CODE == runExecutable(cmd.str()), "Fail to embed manifest containing JSON syntax errors.");
+
+  ZipFile invalid_syntax_zip(tempdir + "invalid_syntax.zip");
+  US_TEST_CONDITION(0 == invalid_syntax_zip.size(), "Test that the invalid manifest.json file was not added");
+
+  std::string invalid_syntax_bundle_name("invalid_syntax.bundle");
+  US_TEST_NO_EXCEPTION_REQUIRED(createDummyBundle(tempdir, invalid_syntax_bundle_name));
+
+  std::ostringstream cmd2;
```
karthikreddy09: I suggest creating a function that constructs the full command string and call it from multiple places instead of constructing them with the stream operators over and over.

tools/rc/ResourceCompiler.cpp
```diff
@@ -110,6 +112,178 @@ std::string us_tempfile()
...
+    mz_uint numBytes = mz_zip_reader_get_filename(&currZipArchive, currZipIndex, archiveName, sizeof archiveName);
+    std::clog << "\tValidating: " << archiveName << " (from " << archiveFile << ") "<< std::endl;
+    try
+    {
+      if (numBytes > 1 && archiveName[numBytes-2] != '/') // The last character is '\0' in the array
+      {
+        std::string archiveEntry(archiveName);
+        if (archiveEntry.find_first_of("/", 0) == archiveEntry.find_last_of("/") &&
+            std::string::npos != archiveEntry.find("/manifest.json"))
+        {
```
ksubramz: Maybe flatten this conditional? Something like
```cpp
mz_uint numBytes = mz_zip_reader_get_filename(&currZipArchive, currZipIndex, archiveName, sizeof archiveName);
if (numBytes <= 1)
{
  continue;
}
try
{
  std::string archiveEntry(archiveName);
  if (archiveEntry.back() != '/' &&
      archiveEntry.find_first_of("/", 0) == archiveEntry.find_last_of("/") &&
      std::string::npos != archiveEntry.find("/manifest.json"))
  ```

File: framework/test/driver/ResourceCompilerTest.cpp (hunk 3 above)
Comment (jeffdiclemente) [reply framework/test/driver/ResourceCompilerTest.cpp]: I forgot to remove this comment. There is no need to delete the zip file now that the temporary directory and everything in it is completely cleaned up after running the tests.

File: framework/test/driver/ResourceCompilerTest.cpp (hunk 6 above)
Comment (jeffdiclemente) [reply framework/test/driver/ResourceCompilerTest.cpp]: They are functionally the same (operator== and std::string::compare). Is there a strong case for one over the other?

File: tools/rc/ResourceCompiler.cpp (hunk 7 above)
Comment (jeffdiclemente) [reply tools/rc/ResourceCompiler.cpp]: It can't be ```const``` because of ```Json::parseFromStream```

File: framework/test/driver/ResourceCompilerTest.cpp (hunk 2 above)
Comment (jeffdiclemente) [reply framework/test/driver/ResourceCompilerTest.cpp]: I agree that using common code is good. In this case, I'd like to put this off until we can move all the resource compiler tests to use gtest, which can easily do parameterized tests. This will eliminate large amounts of test code.

File: framework/test/driver/ResourceCompilerTest.cpp (hunk 12 above)
Comment (jeffdiclemente) [reply framework/test/driver/ResourceCompilerTest.cpp]: I'm not sure how useful this would be. There is nothing special about how the command line string is constructed. I'd like to defer this for now.

File: tools/rc/ResourceCompiler.cpp [hunk 14]
```diff
@@ -110,6 +112,178 @@ std::string us_tempfile()
...
+          char manifestFileContents[MZ_ZIP_MAX_IO_BUF_SIZE];
+          memset(manifestFileContents, 0, MZ_ZIP_MAX_IO_BUF_SIZE);
+          if(MZ_TRUE == mz_zip_reader_extract_file_to_mem(&currZipArchive,
```
saschazelzer: If I read this correctly, the `mz_zip_reader_extract_file_to_mem` call will fail for files with an uncompressed size `> MZ_ZIP_MAX_IO_BUF_SIZE` which is defined to 64 KB. This doesn't look right to me.

tools/rc/ResourceCompiler.cpp
```diff
@@ -110,6 +112,178 @@ std::string us_tempfile()
...
+            {
+              std::string exceptionMsg(archiveFile);
+              exceptionMsg += " (" + archiveEntry + ") : " + e.what();
+              throw InvalidManifest(exceptionMsg);
+            } 
+          }
+        }
+      }
+    }
+    catch (const InvalidManifest&)
```
saschazelzer: This `catch` block is redundant with the one just below.

tools/rc/ResourceCompiler.cpp
```diff
@@ -281,9 +503,35 @@ void ZipArchive::AddResourcesFromArchive(const std::string &archiveFileName)
           {
             throw std::runtime_error("Found duplicate file with name " + std::string(archiveName));
           }
+          std::string archiveEntry(archiveName);
+          if (archiveEntry.find_first_of("/", 0) == archiveEntry.find_last_of("/") &&
+              std::string::npos != archiveEntry.find("/manifest.json"))
+          {
+            char manifestFileContents[MZ_ZIP_MAX_IO_BUF_SIZE];
+            memset(manifestFileContents, 0, MZ_ZIP_MAX_IO_BUF_SIZE);
+            if(MZ_TRUE == mz_zip_reader_extract_file_to_mem(&currZipArchive,
```
saschazelzer: Same concern about file and buffer sizes here.

tools/rc/ResourceCompiler.cpp
```diff
@@ -110,6 +112,178 @@ std::string us_tempfile()
...
+ * @param jsonFile path to a json file.
+ * @param root The parsed Json root object.
+ * @throw InvalidManifest if the json is invalid. Parse error information is in the exception.
+ * If an exception is thrown, the root param is invalid.
+ */
+void parseAndValidateJsonFromFile(const std::string& jsonFile, Json::Value& root)
+{
+  try
+  {
+    std::ifstream json(jsonFile);
```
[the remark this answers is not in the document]
jeffdiclemente: From jsoncpp.cpp (line 1997), it looks like it calls ```rdbuf()``` on the ```std::istream``` object. Its unclear to me from the documentation for ```std::ifstream``` what ```rdbuf()``` does if the file couldn't be opened. 

I think the safest thing to do would be to call ```std::basic_ifstream::is_open()``` after ```std::ifstream json(jsonFile);``` and fail if the file is not open.

tools/rc/ResourceCompiler.cpp
jeffdiclemente: good catch. This should be using ```mz_zip_reader_extract_to_heap``` instead.

tools/rc/ResourceCompiler.cpp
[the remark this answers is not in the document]
saschazelzer: Having an extra `is_open()` call is fine with me.

PR #248, opened 2017-11-01
Increase code coverage for Bundle.cpp

jeffdiclemente:
The goal is at least 80% code coverage for Bundle.cpp.


ksubramz:
Referencing #235

PR #249, opened 2017-11-06
Fix build failures on Linux and MinGW

jeffdiclemente:
Turn off new GCC 7 -Wimplicit-fallthrough warning, jsoncpp.cpp is tripping over this warning.

Fix ```warn_unused_result``` error from use of strtol

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -365,7 +365,7 @@ else()
   foreach(_cxxflag  -Werror -Wall -Wextra -Wpointer-arith -Winvalid-pch -Wcast-align
                     -Wwrite-strings -Woverloaded-virtual -Wnon-virtual-dtor -Wold-style-cast
                     -Wstrict-null-sentinel -Wsign-promo -fdiagnostics-show-option
-                    -fstack-protector-all)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -365,7 +365,7 @@ else()
   foreach(_cxxflag  -Werror -Wall -Wextra -Wpointer-arith -Winvalid-pch -Wcast-align
                     -Wwrite-strings -Woverloaded-virtual -Wnon-virtual-dtor -Wold-style-cast
                     -Wstrict-null-sentinel -Wsign-promo -fdiagnostics-show-option
-                    -fstack-protector-all)
+                    -fstack-protector-all -Wimplicit-fallthrough=0)
```
saschazelzer: I would prefer to not disable this warning for the whole project. We already have comments in our own code to signal to GCC that a fall through is intentional. I suggest to patch jsoncpp instead, see https://github.com/open-source-parsers/jsoncpp/pull/697 for an easy fix.

jeffdiclemente: Ok. i'll patch jsoncpp.cpp and submit.

PR #250, opened 2017-11-06
Patch jsoncpp for GCC 7 implicit-fallthrough warning

jeffdiclemente:
Suppress the warning in jsoncpp.cpp instead of turning it off entirely.


jeffdiclemente:
will do.

saschazelzer:
Thanks!

PR #253, opened 2017-11-13
Increase code coverage for ServiceReferenceBase.cpp

ksubramz:
Addressing #235 

### DIFF
--- a/framework/test/driver/ServiceRegistryTest.cpp
+++ b/framework/test/driver/ServiceRegistryTest.cpp
@@ -106,7 +106,8 @@ void TestServiceReferenceMemberFunctions(BundleContext context)
   // Test ServiceReference member functions GetPropertyKeys()
   std::vector<std::string> keys;
   ref1.GetPropertyKeys(keys);
-  US_TEST_CONDITION(keys.size() > 0, "Test keys size")
+  US_TEST_CONDITION(std::find(keys.begin(), keys.end(), "StringKey") != keys.end(), "Test existence of key StringKey")
+  US_TEST_CONDITION(std::find(keys.begin(), keys.end(), "Status") != keys.end(), "Test existence of key Status")
 
   auto keys_by_val = ref1.GetPropertyKeys();
   US_TEST_CONDITION(keys_by_val == keys, "Test keys equality")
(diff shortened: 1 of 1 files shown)


framework/test/driver/ServiceRegistryTest.cpp
```diff
@@ -88,6 +89,48 @@ void TestUnregisterFix(BundleContext context)
   US_TEST_CONDITION_REQUIRED(!reference.IsConvertibleTo("IBooService"), "Testing for IsConvertibleTo returning false");
 }
 
+void TestServiceReferenceMiscMembers(BundleContext context)
```
karthikreddy09: I would recommend using a separate test point for each member functions rather than clubbing them together.

framework/test/driver/ServiceRegistryTest.cpp
```diff
@@ -88,6 +89,48 @@ void TestUnregisterFix(BundleContext context)
...
+{
+	struct TestServiceA : public ITestServiceA
+	{
+	};
+
+	auto s1 = std::make_shared<TestServiceA>();
+	ServiceProperties props;
+	props["string"] = std::string("A std::string");
```
karthikreddy09: For a second I was wondering why the type is being stored in the properties. Simple names such as "foo", "bar" make it easier to read.

framework/test/driver/ServiceRegistryTest.cpp
ksubramz: After testing out creating an own test point, I my current approach is slightly more cleaner (because the boilerplate of creating/getting service reference seems to be less - even when I factor that out to a function). Let me know if you still think a change is warranted.

framework/test/driver/ServiceRegistryTest.cpp
```diff
@@ -88,6 +89,48 @@ void TestUnregisterFix(BundleContext context)
...
+  props["StringKey"] = std::string("A string value");
+  props["Status"] = false;
+
+  ServiceRegistration<ITestServiceA> reg1 = context.RegisterService<ITestServiceA>(s1, props);
+  ServiceReference<ITestServiceA> ref1 = context.GetServiceReference<ITestServiceA>();
+
+  // Test ServiceReference member functions GetPropertyKeys()
+  std::vector<std::string> keys;
+  ref1.GetPropertyKeys(keys);
+  US_TEST_CONDITION(keys.size() > 0, "Test keys size")
```
jeffdiclemente: Is it possible to write a stricter test? 
If the point of the test is to make sure that the service properties retrieved are correct, checking that more than 1 property exists doesn't seem sufficient.

PR #256, opened 2017-12-20
Update BundlePrivate.cpp and BundleThread.cpp code coverage

jeffdiclemente:
These classes were difficult to test due to the interactions between BundlePrivate and BundleThread.  In certain use cases, the result of the test varied even on the same OS platform. For example, where I would expect an exception to be thrown, it would not throw all the time. These raises a concern as to whether there are some hidden hard to see race conditions.

BundleThread.cpp falls short of 80% code coverage however I don't think its worth it to put more effort into reaching the code coverage goal. Effort would be better spent getting a few of the CppMicroServices developers together to thoroughly review and document these classes.


framework/test/driver/BundleTest.cpp
```diff
@@ -815,6 +816,209 @@ void TestBundleGetPropertyKeys()
...
+// Test installing a bundle with invalid meta-data in its manifest.
+void TestBundleManifestFailures()
+{
+  auto f = FrameworkFactory().NewFramework();
+  f.Start();
+  auto bc = f.GetBundleContext();
+
+  // throw if manifest.json bundle version key is not a string type
+  US_TEST_FOR_EXCEPTION(std::runtime_error, testing::InstallLib(bc, "TestBundleWithInvalidVersionType"));
+  US_TEST_CONDITION(2 == bc.GetBundles().size(), "Test that an invalid bundle.version type results in not installing the bundle.");
```
karthikreddy09: This is not forward compatible. Since the bundle size is hardcoded, If the main executable is changed to not being a bundle this test fails. I would suggest searching for a bundle with the symbolic name from the list of all bundles.

framework/test/driver/BundleTest.cpp
```diff
@@ -879,5 +1083,18 @@ int BundleTest(int /*argc*/, char* /*argv*/[])
   TestBundleGetPropertyKeys();
   TestBundleGetProperty();
 
+#if defined (US_BUILD_SHARED_LIBS)
```
karthikreddy09: Why are these tests invalid for static builds?

jeffdiclemente: I think only ```TestBundleManifestFailures``` should be in this ```#ifdef``` block since all statically linked bundles would be installed when the ```Framework``` starts and the bundle designed to have an invalid manifest would fail all the tests. I'll double check and update.

jeffdiclemente: ```TestBundleStopFail``` and ```TestBundleStartFail``` can't be built for static builds only because of the statically linked bundle tests in ```TestBundleStates()```.

framework/test/driver/BundleTest.cpp
jeffdiclemente: I can't check the bundle name for two of the bundles since they are designed not to have bundle names. I think the best thing to check is a hard count on the # of bundles and make sure that those bundles are named ```system_bundle``` and '''main'''.

jeffdiclemente:
@CppMicroServices/developers please review. There are two functional changes in ```BundlePrivate.cpp```. One fixes a bug (a std::runtime_error object wasn't being correctly thrown).

framework/test/driver/BundleTest.cpp
```diff
@@ -815,6 +816,254 @@ void TestBundleGetPropertyKeys()
...
+// Test installing a bundle with invalid meta-data in its manifest.
+void TestBundleManifestFailures()
+{
+  auto f = FrameworkFactory().NewFramework();
+  f.Start();
+  auto bc = f.GetBundleContext();
+
+  std::vector<std::string> validBundleNames({f.GetSymbolicName(), "main"});
```
ksubramz: If both `validBundleNames` and the container `names` inside the lambda `validateBundleNames` are `std::set`s instead, then there's no need to sort the containers before calling `std::includes`

ksubramz: Good work!

jeffdiclemente:
@CppMicroServices/developers If there are no objections, I will merge this tomorrow at 4pm US EST.

PR #261, opened 2018-02-20
Wait for framework to shutdown

jeffdiclemente:


saschazelzer:
Sorry for the late comment. The change is fine, but tries to work around a deeper problem, as you are probably aware of anyway. I have a reproduction case for a runaway bundle thread during shutdown (depending on timings) but the fix is not obvious. I am thinking about re-doing the whole bundle thread implementation to make it look like modern C++ and also easier to read and reason about.

jeffdiclemente:
@saschazelzer you're right, this is a workaround for a larger problem. Could you provide the reproduction case? I and others here could take a look at it and see if we can determine a fix.

Re-factoring the bundle thread implementation is a good idea.

PR #269, opened 2018-04-25
Fix failure caused by permission issues in temp dir

### LINKED ISSUE #266: Permissions issues in %TEMP%\data directory causes CppMicroServices to fail
# Reproduction Steps

Under Windows,
## First
1. Make the directory `%TEMP%\data (C:\Users\<username>\AppData\Local\Temp\data)` inaccessible. (Right click and deny all access control for the user)
2. Run the gtests.  `FrameworkTest.CustomConfiguration` test will fail (the test sets the `FRAMEWORK_STORAGE` configuration to the `%TEMP%` dir)

## Second
1. Delete the directory `%TEMP%\data` and touch a file `%TEMP%\data`.
2. Run the gtests.  `FrameworkTest.CustomConfiguration` will fail again.

# Sequence of calls inside CppMicroServices to trigger the exceptions

- `getFramework()` calls `initOnce()` which sets the persistent storage location (property `FRAMEWORK_STORAGE`) to `%TEMP%` directory.
- `CoreBundleContext` calls `GetFileStorage(this, "data"); `during initialization
   - `GetFileStorage()` takes the path set in `FRAMEWORK_STORAGE` and appends "data", which is `%TEMP%\data`
  - If this directory `%TEMP%\data` doesn't exist, `GetStorageFile()` tries to `_mkdir()` the directory.
    - If the call fails because the directory already exists, it's a no-op.
    - If the call fails due to other reasons, it throws an exception
  - If the path `%TEMP%\data` exists, but it's not accessible,` GetStorageFile()` throws an exception [1]
  - If the path `%TEMP%\data` exists, but it's not a directory, `GetStorageFile()` throws an exception

[1]  Interestingly, on Windows, the `util::Exists()` call succeeds on an inaccessible directory. The underlying `GetFileAttributesW()` returns `FILE_ATTRIBUTE_DIRECTORY 16 (0x10)`. However, the `util::IsDirectory()` returns false `(_stat returns an ENOENT errno)`, and we throw `std::runtime_error("Not a directory")...`

# Bug?

This is probably functioning as designed, but we have been bit by this "mis-behavior" and our customers have run into issues because they either had wrong permissions on `data` directory or they had a file under `%TEMP%` named `data`.

# Potential Solutions

IIUC, `CoreBundleContext` creates this directory for the persistent storage feature of CppMicroServices , which is WIP and not slated for completion in our near future. The following are the potential solutions/workarounds I can think of. Please suggest your preference or any alternate ideas.

- Change the throwing of these particular exceptions to not throw any exceptions.
  - Effort: Minimal.
  - Temporary until we implement persistent storage feature.
- Catch the exceptions in `CoreBundleContext::Init()` surrounding the `GetFileStorage()` call and log them. Change GetFileStorage() call to not create a directory if it doesn't exist.
  - Effort: Minimal.
  - Temporary until we implement persistent storage feature.
- Comment out the `GetFileStorage()` invocation in `CoreBundleContext::Init()`
  - Effort: Minimal.
  - Temporary until we implement persistent storage feature.
  - Added advantage of not creating the stray `%TEMP%\data` directory. 
- Remove all persistent storage code.
  - Effort: Medium.
  - Probably a cleaner solution because it appears that implementing persistent storage isn't in the plans for the near future.

I think the second solution is better for the time being, because the exceptions don't get thrown and the stray `%TEMP%\data` directory isn't created.

ksubramz:
Fix #266 by ignoring the exception thrown by the erstwhile
GetFileStorage (renamed to GetPersistentStoragePath)

Disable the creation of the "data" directory under the persistent
storage directory when the framework starts

Make minor refactoring changes

### DIFF
--- a/framework/src/bundle/CoreBundleContext.cpp
+++ b/framework/src/bundle/CoreBundleContext.cpp
@@ -139,7 +139,15 @@ void CoreBundleContext::Init()
 //  {
 //    dataStorage = GetFileStorage(this, "data");
 //  }
-  dataStorage = GetFileStorage(this, "data");
+  try
+  {
+    dataStorage = GetPersistentStoragePath(this, "data", /*create=*/ false);
+  }
+  catch (const std::exception& e)
+  {
+    DIAG_LOG(*sink) << "Ignored runtime exception with message'" << e.what() << "' from the GetPersistentStoragePath function.\n";
+  }
+
 
   systemBundle->InitSystemBundle();
   _us_set_bundle_context_instance_system_bundle(systemBundle->bundleContext.Load().get());
--- a/framework/test/gtest/FrameworkTest.cpp
+++ b/framework/test/gtest/FrameworkTest.cpp
@@ -220,9 +220,7 @@ TEST(FrameworkTest, FrameworkStartsWhenFileNamedDataExistsInTempDir)
     ScopedFile(std::string directory, std::string filename) :
       filePath(std::move(directory) + util::DIR_SEP + std::move(filename))
     {
-      std::fstream file(filePath, std::fstream::out);
-      file << "test";
-      file.close();
+      CreateFile();
     }
     ~ScopedFile()
     {
@@ -231,6 +229,13 @@ TEST(FrameworkTest, FrameworkStartsWhenFileNamedDataExistsInTempDir)
       std::remove(filePath.c_str());
     }
   private:
+    void CreateFile()
+    {
+      std::fstream file(filePath, std::fstream::out);
+      ASSERT_FALSE(file.fail()) << "Failbit of the file stream should not be set.";
+      file << "test";
+      file.close();
+    }
     const std::string filePath;
   };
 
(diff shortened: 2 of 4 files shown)


framework/test/gtest/FrameworkTest.cpp
```diff
@@ -208,6 +212,56 @@ TEST(FrameworkTest, CustomConfiguration)
...
 
+TEST(FrameworkTest, FrameworkStartsWhenFileNamedDataExistsInTempDir)
+{
+  class ScopedFile
+  {
+  public:
+    ScopedFile(std::string directory, std::string filename) :
+      filePath(std::move(directory) + util::DIR_SEP + std::move(filename))
+    {
+      std::fstream file(filePath, std::fstream::out);
```
saschazelzer: It is probably good to check the failbit of the stream and if it is bad, fail the test.

framework/src/bundle/CoreBundleContext.cpp
```diff
@@ -139,7 +139,15 @@ void CoreBundleContext::Init()
 //  {
 //    dataStorage = GetFileStorage(this, "data");
 //  }
-  dataStorage = GetFileStorage(this, "data");
+  try
```
saschazelzer: This is solution number two, which is also fine with me.

ksubramz: You are right. I forgot to update the issue with this information. Will do that.

PR #271, opened 2018-05-01
minimal support for cross compilation

### LINKED ISSUE #270: support cross compilation
Add support to cross compile CppMicroServices minimally for iOS, Android and linux-arm.

jeffdiclemente:
Fixes #270 

This is the minimum required to allow clients to cross compile CppMicroServices.

This change is meant to allow the host version of the usResourceCompiler to be used during a cross compilation build. 

There is probably more that can be done to support cross compilation for clients however without use cases, this seems the most obvious thing to support right now.


CMakeLists.txt
```diff
@@ -362,10 +374,15 @@ else()
     set(US_CXX_FLAGS "${US_CXX_FLAGS} -stdlib=libc++")
   endif()
 
+  # If not cross-compiling, turn on Stack Smashing Protection.
+  # This is disabled for ARM as it requires an additional dependecy on libssp.so.
```
jeffdiclemente: typo: dependecy

CMakeLists.txt
```diff
@@ -249,6 +249,18 @@ if(APPLE)
...
+# Cross-compiled builds need access to the host version of usResourceCompiler.
+# ImportExecutables.cmake is generated by the host and contains the cmake target
+# definition for the host version of usResourceCompiler. This file is then included by the
+# cross-compiled build so that it can access the host version of usResourceCompiler.
+#
+# Without this file, the crossbuild cannot use the host version of usResourceCompiler, because
+# cmake will try to create & build a target for a cross-compiled version of usResourceCompiler.
+# ImportExecutables.cmake allows us to define usResourceCompiler as a cmake target during the
+# host build and then include it during the crossbuild. This defines the usResourceCompiler
+# cmake target with the host location, which prevents cmake from trying to cross-compile it.
```
jeffdiclemente: clarify terminology (host / target)
use less words

CMakeLists.txt
```diff
@@ -602,9 +622,12 @@ add_subdirectory(doc)
 
 if(NOT US_NO_INSTALL)
 
+# Only install the target if we are not cross-compiling, because the target is only built during the host build
```
jeffdiclemente: ambiguous use of "target"

CMakeLists.txt
```diff
@@ -579,8 +596,11 @@ configure_file(
 
 add_subdirectory(tools/rc)
 
-export(TARGETS ${US_RCC_EXECUTABLE_TARGET}
-       FILE ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Targets.cmake)
+# Only export the target if we are not cross-compiling, because the target is only built during the host build
```
jeffdiclemente: ambiguous use of "target"

tools/rc/CMakeLists.txt
```diff
@@ -7,20 +7,32 @@ add_definitions(-DUS_RCC_EXECUTABLE_NAME=\"${US_RCC_EXECUTABLE_OUTPUT_NAME}\")
...
-add_executable(${US_RCC_EXECUTABLE_TARGET} ${srcs})
-set_property(TARGET ${US_RCC_EXECUTABLE_TARGET} PROPERTY OUTPUT_NAME ${US_RCC_EXECUTABLE_OUTPUT_NAME})
-if(WIN32)
-  target_link_libraries(${US_RCC_EXECUTABLE_TARGET} Shlwapi)
-endif()
+# When cross-compiling, import the host executable targets from a file
+if(CMAKE_CROSSCOMPILING)
+  # See https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/CrossCompiling
+  include(${IMPORT_EXECUTABLES})
+
```
jeffdiclemente: extra space

CMakeLists.txt
```diff
@@ -362,10 +374,15 @@ else()
     set(US_CXX_FLAGS "${US_CXX_FLAGS} -stdlib=libc++")
   endif()
 
+  # If not cross-compiling, turn on Stack Smashing Protection.
+  # This is disabled for ARM as it requires an additional dependecy on libssp.so.
+  if(NOT CMAKE_TOOLCHAIN_FILE)
```
saschazelzer: I think this should test for the target system name instead of testing for the presence of a toolchain file (which could be present without actually cross-compiling, I believe). Maybe somthing like
```
if (NOT CMAKE_SYSTEM_NAME STREQUAL "Android")
```
or
```
if (NOT CMAKE_SYSTEM MATCHES "arm")
```

CMakeLists.txt
```diff
@@ -579,8 +596,11 @@ configure_file(
 
 add_subdirectory(tools/rc)
 
-export(TARGETS ${US_RCC_EXECUTABLE_TARGET}
-       FILE ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Targets.cmake)
+# Only export the target if we are not cross-compiling, because the target is only built during the host build
+if(NOT CMAKE_CROSSCOMPILING)
```
saschazelzer: I believe the intent was to cross-compile bundles only, which if fine. However, we might want to cross-compile *and* install the complete project in the future, such that an install tree can be packaged and used on the target platform to  develop native bundles (I am thinking of e.g. creating .deb or .rpm packages in the future). This means we would have to buid and install the cross-compiled version of usResouceCompiler as well.

CMakeLists.txt
jeffdiclemente: I agree. I think I may be able to make the check more specific and determine whether -Wstack-protector would work through a compile time check, instead of checking against a target system name. I think its conceivable that clients could have stack smashing available in their ARM toolchains.

CMakeLists.txt
jeffdiclemente: good point. I agree.

jeffdiclemente:
@CppMicroServices/developers, let me know if there are any more issues to resolve. I'd like to merge this into development by this Friday. Thanks!

PR #283, opened 2018-06-25
static ServiceTracker object crashes in ServiceTracker::Close()

### LINKED ISSUE #281: static ServiceTracker object crashes in ServiceTracker::Close()
Maintaining a static ```ServiceTracker``` object can cause the program to crash on termination in ```ServiceTracker::Close()```.

The seg fault is caused by accessing the a ```nullptr``` ```CoreBundleContext``` pointer in ```BundleContext::GetLogSink()```

test case:
```
#include "cppmicroservices/Framework.h"
#include "cppmicroservices/FrameworkEvent.h"
#include "cppmicroservices/FrameworkFactory.h"
#include <cppmicroservices/ServiceInterface.h>
#include "cppmicroservices/ServiceTracker.h"

#include <memory>

// A dummy interface to use with the service tracker
namespace {
    class Foo {
    public:
        virtual ~Foo() = default;
    };
}
CPPMICROSERVICES_DECLARE_SERVICE_INTERFACE(Foo, "org.cppmicroservices.test.servicetracker.Foo");

int main()
{
    auto f = cppmicroservices::FrameworkFactory().NewFramework();
    ASSERT_TRUE(f);
    f.Start();

    static std::shared_ptr<cppmicroservices::ServiceTracker<Foo>> globalTracker(std::make_shared<cppmicroservices::ServiceTracker<Foo>>(f.GetBundleContext()));
    globalTracker->Open();

    f.Stop();
    f.WaitForStop(std::chrono::milliseconds::zero());
}
// seg fault access violation when this process destroys static objects...
```

Minimally ```ServiceTracker::Close()``` shouldn't crash if the ```Framework``` has been destroyed.

jeffdiclemente:
Fixes #281

### DIFF
--- a/framework/test/gtest/GlobalServiceTrackerTest.cpp
+++ b/framework/test/gtest/GlobalServiceTrackerTest.cpp
@@ -36,7 +36,7 @@ limitations under the License.
 namespace {
     class Foo {
     public:
-        virtual ~Foo() = default;
+        virtual ~Foo() {}
     };
 }
 
--- a/framework/test/gtest/GlobalServiceTrackerTest.cpp
+++ b/framework/test/gtest/GlobalServiceTrackerTest.cpp
@@ -1,0 +1,60 @@
+/*=============================================================================
+
+Library: CppMicroServices
+
+Copyright (c) The CppMicroServices developers. See the COPYRIGHT
+file at the top-level directory of this distribution and at
+https://github.com/CppMicroServices/CppMicroServices/COPYRIGHT .
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


framework/test/gtest/GlobalServiceTrackerTest.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include "gtest/gtest.h"
+
+#include <memory>
+
+// A dummy interface to use with service trackers
+namespace {
+    class Foo {
+    public:
+        virtual ~Foo() {}
```
karthikreddy09: = default;

In fact, the destructor does not have to be virtual. This class is for test purpose only and there is no subclass for it.

PR #285, opened 2018-07-10
Fix methods in HttpServletRequest.cpp and implement HandlePut, HandleDelete and HandlePost. 

m2cci-MJN:
* In HttpServletRequest.cpp :
    -
- Fix the GetQueryString() method.
- Fix m_Uri attribute, now contain the query uri.
- Implementation of GetRequestBody() method that return the request body. 
  
* In ServletContainer.cpp : 
   -
* Implementation of the HandlePost(), HandlePut() and HandleDelete() methods.

### DIFF
--- a/httpservice/include/cppmicroservices/httpservice/HttpServletRequest.h
+++ b/httpservice/include/cppmicroservices/httpservice/HttpServletRequest.h
@@ -10,7 +10,7 @@
   you may not use this file except in compliance with the License.
   You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+  http://www.apache.org/licenses/LICENSE-2.0
 
   Unless required by applicable law or agreed to in writing, software
   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


httpservice/include/cppmicroservices/httpservice/HttpServletRequest.h
```diff
@@ -38,15 +38,14 @@ struct HttpServletRequestPrivate;
 class US_HttpService_EXPORT HttpServletRequest
 {
 public:
-
   ~HttpServletRequest();
 
-  HttpServletRequest(const HttpServletRequest& o);
-  HttpServletRequest& operator=(const HttpServletRequest& o);
+  HttpServletRequest(const HttpServletRequest &o);
```
ksubramz: Hi, first, thanks for this PR!

The first thing that jumps out is the various changes that puts the pointer and reference identifiers (* and &) next to the variables. We don't explicitly mention - but we prefer letting these identifiers stay adjacent to the types. If you look at the framework code, the convention we follow is these identifiers should be next to the type.

saschazelzer:
Thank you for the pull request. I had a quick look and the changes look good. In order to get them merged, please have a look at our [CONTRIBUTING](../blob/development/CONTRIBUTING.rst) document. Especially, please

- Remove the white-space changes
- Squash your commits
- Sign-off your work

arcadien:
Thanks @saschazelzer and @ksubramz for your replies. We restarted our development branch, and opened a [ticket](https://github.com/CppMicroServices/CppMicroServices/issues/287) to better define what we are doing here. We will pay attention to the CONTRIBUTING.md content for the next coming PR ; this one can be cancelled in the meantime.

httpservice/include/cppmicroservices/httpservice/HttpServletRequest.h
arcadien: Hi, 
Is it possible to provide needed code convention? we are using clang-format to automatically format the code and hence we can provide the config file

arcadien: https://zed0.co.uk/clang-format-configurator/ can be used to quickly create/check a format file

ksubramz: Thanks, this is definitely in the cards for us - to create a project-wide clang-format.

Issue #284, opened 2018-07-10
Improve Performance of Demangling Interface Names in CppMicroServices


ksubramz:
CppMicroServices stores interface names in the ServiceRegistry as the keys. An example of a interface name is `Foo::Bar`. This interface name is found by demangling the typeid of T whenever a service is registered through a `BundleContext::RegisterService<T>` call.

Later, when the service is requested through a `BundleContext::GetServiceReference<T>` call, CppMicroServices converts the T into the interface name and queries the map in the service registry.

However, these demangling calls are costly - benchmark profile for a `GetServiceReference<T>` call suggests that the demangling is done thrice and takes about 50% of the time.

Registering and getting the service references are essential operations used by the clients. A decent performance improvement like this proposal is beneficial.

# Profiling Runs

## Run 1

|           |                 |
|-----------|-----------------|
| Platform | Win64 |
| Repetitions | 100  |
| Program   | Intel VTune Amplifier |
| Snippet | `(void)framework->GetBundleContext().GetServiceReference<NSBar::Bar>();` |

***Hotspots***

| Function Stack | CPU Time: Total |	CPU Time: Self |	
|----------------|-----------------|------------------------|
| benchmark::Fixture::Run	| 64.8419 |	0|
|          ServiceFixture_GetBarReferenceByInterface_Benchmark::BenchmarkCase |	64.4675 |	0.162012 |
| cppmicroservices::BundleContext::GetServiceReference |	45.6606 |	0.161393 | CppMicroServices3.dll | 
| cppmicroservices::detail::GetDemangledName |	25.133 |	2.14582 |

***Diving down into the demangling call-stack***

(The indentation corresponds to the callers)
```
| Function name | Time spent (including callees) in seconds |
1. us_service_interface_iid<struct cppmicroservices::ServiceFindHook> 11.4323s
  cppmicroservices::ServiceHooks::FilterServiceReferences	11.4323s
   cppmicroservices::ServiceRegistry::Get_unlocked	11.4323s
    cppmicroservices::ServiceRegistry::Get	11.4323s
     cppmicroservices::BundleContext::GetServiceReference	11.4323s
2. cppmicroservices::ServiceReference<class NSBar::Bar>::ServiceReference<class NSBar::Bar>	7.1137s
  ServiceFixture_GetBarReferenceByInterface_Benchmark::BenchmarkCase	7.1137s
[...]

PR #286, opened 2018-07-17
Improve performance by caching the demangling process

ksubramz:
Address #284. Also, refer to RFC PR https://github.com/CppMicroServices/rfcs/pull/2

### DIFF
--- a/framework/include/cppmicroservices/ServiceInterface.h
+++ b/framework/include/cppmicroservices/ServiceInterface.h
@@ -53,7 +53,7 @@
  * @tparam T The service interface type.
  * @return A unique id for the service interface type T.
  */
-template<class T> std::string us_service_interface_iid();
+template<class T> const std::string& us_service_interface_iid();
 
 namespace cppmicroservices {
 
@@ -140,13 +140,17 @@ namespace detail
 }
 
 /// \cond
-template<class T> std::string us_service_interface_iid()
+template<class T> const std::string& us_service_interface_iid()
 {
   static const std::string name = cppmicroservices::detail::GetDemangledName(typeid(T));
   return name;
 }
 
-template<> inline std::string us_service_interface_iid<void>() { return std::string(); }
+template<> inline const std::string& us_service_interface_iid<void>()
+{
+  static const std::string name("");
+  return name;
+}
 /// \endcond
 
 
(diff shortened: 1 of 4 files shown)


framework/test/gtest/ServiceReferenceTest.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// This test exercises the 2 ways to register a service 
+//   a. using the name of the interface i.e. "Foo::Bar"
+//   b. using the type of the interface i.e. <Foo::Bar>
+// multiplied with 2 ways to get the service reference
+//   a. using the name of the interface
+//   b. using the type of the interface
+// NOTE: We do have tests for this in different places,
+// but this test consolidates all these modes together.
```
jeffdiclemente: Since we already have tests in different places can the other, redundant, tests be removed in favor of these tests?

ksubramz: Jeff, what I meant was that the fragments exercising the same functionality do exist, but they are usually parts of a larger test.

Whereas this test just exercises the 2 variants of RegisterService x 2 variants of GetService (to ensure the functionality isn't affected). I agree that the test I added is redundant - i don't mind removing it, I guess.

jeffdiclemente: I'm not suggesting to remove the tests you've written. I'm asking if any of the tests which already exist can be removed in favor of this test.

ksubramz: I've gone through them and it seems they are integral to the existing tests i.e. they can't be removed.

framework/include/cppmicroservices/ServiceInterface.h
```diff
@@ -142,7 +142,8 @@ namespace detail
 /// \cond
 template<class T> std::string us_service_interface_iid()
```
saschazelzer: In this particular case, I think it makes more sense for the return type to be `const std::string&` to avoid a potential copy. If we do this, we should also look at all `us_service_interface_iiid()` calls and use something like
```
auto& clazz = us_service_interface_iid();
```

ksubramz: Thanks I have made this particular change.

> you could check if changing std::string to const std::string& is doable without much effort.

Did you mean: store the `string`s in the map in registry as `const std::string &`? Now that you've broached this topic, I am working on an RFC where the idea is to store the keys of the services as numeric IDs instead of the `std::string`s currently. This was suggested by one of our clients and his belief is that the string comparison used when doing map operations is significantly more expensive when compared to IDs. I'll have an RFC up soon.

saschazelzer: I meant exactly what you did :+1: 
Using integer ids internally sounds a good idea, let's see what you come up with!

PR #289, opened 2018-07-19
Initial version of clang format config

arcadien:
A first version of the format file. It seems to cope with the code you wrote. Adjustments may be needed, https://zed0.co.uk/clang-format-configurator/ is a good place to live-check the config.


saschazelzer:
Thanks for the config. I will merge it as it is, but as you wrote it will need some modifications to more closely match our style. We also need to update our docs, sanitize the sources according to the final format and maybe add a Git pre-commit hook.

PR #294, opened 2018-08-01
Code formatting

saschazelzer:
Source code formatting using clang-format.

The first commit contains the modified `.clang-format` file, some minor style-related changes and a shell script to run clang-format. The shell script formats all `.cpp` and `.h` files except for the `third_party` folder.

The second commit contains the changes by running the `format_code.sh` script.

I briefly looked at options to automatically check (but not format) commits to follow the code style (e.g. git hooks), but finally refrained from elaborate cross-platform solutions. So there is no mechanism in place to prevent commits from introducing non-conforming changes. I am fine with that.


cmake/try_compile/us_have_regex.cpp
```diff
@@ -1,5 +1,7 @@
 #include <regex>
-int main() {
+int
```
ksubramz: For the function definitions, my preference would be something like
```cpp
int main()
{
```
instead of 
```cpp
int
main()
{
```

( I mean if the consensus is we stick to the Mozilla default style i.e. option 2 going forward, I am fine but my preference is #1 :-).  )

I believe the configuration options are `AlwaysBreakAfterDefinitionReturnType`  and `AlwaysBreakAfterReturnType`

saschazelzer: Okay. I am also going to add indentation for pre-processor directives. Also, we can easily change the style later if we find we are not happy with something.

jeffdiclemente:
@saschazelzer Thanks!

PR #300, opened 2018-08-27
Remove Chrono.h

carneyweb:
No longer need Chrono.h which chose between two clock implementations depending on compiler version. We now require compiler versions that are all compatible with one version.


framework/src/bundle/BundleArchive.h
```diff
@@ -23,11 +23,12 @@
 #ifndef CPPMICROSERVICES_BUNDLEARCHIVE_H
 #define CPPMICROSERVICES_BUNDLEARCHIVE_H
 
-#include "cppmicroservices/detail/Chrono.h"
+#include <cppmicroservices/Bundle.h>
```
saschazelzer: Is that header actually needed here?

framework/include/cppmicroservices/Bundle.h
```diff
@@ -45,6 +45,8 @@ class ServiceReference;
 
 typedef ServiceReference<void> ServiceReferenceU;
 
+using clock_t = std::chrono::high_resolution_clock;
```
saschazelzer: Please stick to using `std::steady_clock`, because we want a monotonic clock here. If the high resolution clock was used to "fix" the failing MinGW BundleTest, increase the wait time in `BundleTest.cpp` from 10ms to 100ms instead.

carneyweb:
Hello again Sasha :-)

In answer to your questions:

Yes, I had changed the clock to “high_resolution_clock” clock because of the problems in the MinGW bundle tests. I’ll change that back to “steady_clock” and update the test case to 100ms.
The inclusion of that header is because I put the “using clock_t…” statement there because I had deleted the Chrono.h file and just used the “steady_clock” everywhere. When that failed the MinGW tests, I thought it would be prudent to put the “using clock_t…” declaration in one place again, but didn’t think it was worth re-introducing the Chrono.h header for one using statement. If you’d like header restored, I’ll be happy to.
- Mike.


On August 29, 2018 at 2:34:06 AM, Sascha Zelzer (notifications@github.com) wrote:

@saschazelzer requested changes on this pull request.

In framework/src/bundle/BundleArchive.h:

> @@ -23,11 +23,12 @@
 #ifndef CPPMICROSERVICES_BUNDLEARCHIVE_H
 #define CPPMICROSERVICES_BUNDLEARCHIVE_H
  
-#include "cppmicroservices/detail/Chrono.h"
+#include <cppmicroservices/Bundle.h>
Is that header actually needed here?

In framework/include/cppmicroservices/Bundle.h:

> @@ -45,6 +45,8 @@ class ServiceReference;
  
 typedef ServiceReference<void> ServiceReferenceU;
  
+using clock_t = std::chrono::high_resolution_clock;
Please stick to using std::steady_clock, because we want a monotonic clock here. If the high resolution clock was used to "fix" the failing MinGW BundleTest, increase the wait time in BundleTest.cpp from 10ms to 100ms instead.

framework/test/driver/BundleTest.cpp
```diff
@@ -545,13 +546,13 @@ void TestBundleStates()
   bundle.Stop();
   US_TEST_CONDITION((bundle.GetState() & Bundle::STATE_ACTIVE) == false,
                     "Test stopped bundle state")
-  std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
karthikreddy09: why change from 10 to 100?

carneyweb:
Hi Abhinay

It’s because of the resolution of the clock… actions were happening too quickly for the test to pass. Sasha had written this in his comments:

Please stick to using std::steady_clock, because we want a monotonic clock here. If the high resolution clock was used to "fix" the failing MinGW BundleTest, increase the wait time in BundleTest.cpp from 10ms to 100ms instead.

- Mike.

On October 15, 2018 at 8:46:04 AM, Abhinay Reddyreddy (notifications@github.com) wrote:

@karthikreddy09 approved this pull request.

In framework/test/driver/BundleTest.cpp:

> @@ -545,13 +546,13 @@ void TestBundleStates()
   bundle.Stop();
   US_TEST_CONDITION((bundle.GetState() & Bundle::STATE_ACTIVE) == false,
                     "Test stopped bundle state")
-  std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
why change from 10 to 100?

PR #303, opened 2018-10-03
Dlopen options

carneyweb:
Added new API for SharedLibrary::Load which takes a map of key/value frameworkProperties and uses the value of a new key called LIBRARY_LOAD_OPTIONS to pass to the SharedLibrary::Load(int) call, which in turn uses that value to pass to the dlopen call on POSIX systems. Then in BundlePrivate::Start0 I modified the call to lib.Load() to pass in the frameworkProperties to the new api.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,5 @@ third_party/googletest
 !third_party/googletest/googletest/cmake
 !third_party/googletest/googletest/include
 !third_party/googletest/googletest/src
+/maci64
+/glnxa64
--- a/.gitignore
+++ b/.gitignore
@@ -47,5 +47,3 @@ third_party/googletest
 !third_party/googletest/googletest/cmake
 !third_party/googletest/googletest/include
 !third_party/googletest/googletest/src
-/maci64
-/glnxa64
--- a/framework/include/cppmicroservices/Constants.h
+++ b/framework/include/cppmicroservices/Constants.h
@@ -422,9 +422,16 @@ US_Framework_EXPORT extern const std::string SCOPE_BUNDLE; // = "bundle"
  */
 US_Framework_EXPORT extern const std::string SCOPE_PROTOTYPE; // = "prototype"
 
+/**
+ * Service property that holds optional flags for dlopen calls on POSIX systems.
+ */
 US_Framework_EXPORT extern const std::string LIBRARY_LOAD_OPTIONS;
 
 }
 }
 
 #endif // CPPMICROSERVICES_CONSTANTS_H
+
+// Local Variables:
+// c-file-style: "cppms"
+// End:
--- a/framework/include/cppmicroservices/Constants.h
+++ b/framework/include/cppmicroservices/Constants.h
@@ -21,11 +21,11 @@
 =============================================================================*/
 
 #ifndef CPPMICROSERVICES_CONSTANTS_H
-#define CPPMICROSERVICES_CONSTANTS_H
+#  define CPPMICROSERVICES_CONSTANTS_H
 
-#include "cppmicroservices/FrameworkConfig.h"
+#  include "cppmicroservices/FrameworkConfig.h"
 
-#include <string>
+#  include <string>
 
 namespace cppmicroservices {
 
@@ -45,13 +45,15 @@ namespace Constants {
  * Location identifier of the OSGi <i>system bundle </i>, which is defined
  * to be &quot;System %Bundle&quot;.
  */
-US_Framework_EXPORT extern const std::string SYSTEM_BUNDLE_LOCATION; // = "System Bundle";
+US_Framework_EXPORT extern const std::string
+  SYSTEM_BUNDLE_LOCATION; // = "System Bundle";
 
 /**
  * Alias for the symbolic name of the OSGi <i>system bundle </i>. It is
  * defined to be &quot;system.bundle&quot;.
  */
-US_Framework_EXPORT extern const std::string SYSTEM_BUNDLE_SYMBOLICNAME; // = "system_bundle";
+US_Framework_EXPORT extern const std::string
+  SYSTEM_BUNDLE_SYMBOLICNAME; // = "system_bundle";
 
 /**
  * Manifest header identifying the bundle's activator.
@@ -62,7 +64,8 @@ US_Framework_EXPORT extern const std::string SYSTEM_BUNDLE_SYMBOLICNAME; // = "s
  *
  * The header value may be retrieved via the \c Bundle::GetProperty method.
  */
-US_Framework_EXPORT extern const std::string BUNDLE_ACTIVATOR; // = "bundle.activator";
+US_Framework_EXPORT extern const std::string
+  BUNDLE_ACTIVATOR; // = "bundle.activator";
 
 /**
  * Manifest header identifying the bundle's category.
@@ -70,7 +73,8 @@ US_Framework_EXPORT extern const std::string BUNDLE_ACTIVATOR; // = "bundle.acti
  * The header value may be retrieved from the \c AnyMap object
  * returned by the \c Bundle::GetHeaders() method.
  */
-US_Framework_EXPORT extern const std::string BUNDLE_CATEGORY; // = "bundle.category";
+US_Framework_EXPORT extern const std::string
+  BUNDLE_CATEGORY; // = "bundle.category";
 
 /**
  * Manifest header identifying the bundle's copyright information.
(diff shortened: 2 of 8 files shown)


carneyweb:
OK. I’ll fix that up. That was inadvertent. Sorry about that.

- Mike.

On October 3, 2018 at 9:33:00 AM, Jeff DiClemente (notifications@github.com) wrote:

@jeffdiclemente commented on this pull request.

In general, it looks like the indentations are not consistent with the code base's standard style. There is a .clang-format file at the root of the source tree which can be used to format all the source files automatically.

framework/include/cppmicroservices/SharedLibrary.h
```diff
@@ -35,186 +37,200 @@
...
+    
+    /**
+     * Loads the shared library pointed to by this SharedLibrary object.  On
+     * POSIX systems dlopen() is called with flags from the key
+     * LIBRARY_LOAD_OPTIONS in the frameworkProperties map.
+     *
+     * @throws std::logic_error If the library is already loaded.
+     * @throws std::runtime_error If loading the library failed.
+     */
+    void Load(const std::unordered_map<std::string, Any>& frameworkProperties);
```
jeffdiclemente: I'm not sure that another ```Load``` function is necessary. It seems like it only adds complexity. Since a client sets the ```LIBRARY_LOAD_OPTIONS``` config option using the actual POSIX ```RTLD_*``` values, they can be passed directly to ```Load(int flags)```.

framework/include/cppmicroservices/Constants.h
```diff
@@ -449,7 +421,17 @@ US_Framework_EXPORT extern const std::string SCOPE_BUNDLE; // = "bundle"
...
+US_Framework_EXPORT extern const std::string LIBRARY_LOAD_OPTIONS;
+
 }
 }
 
 #endif // CPPMICROSERVICES_CONSTANTS_H
+
+// Local Variables:
+// c-file-style: "cppms"
+// End:
```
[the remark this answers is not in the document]
carneyweb: removed.

framework/include/cppmicroservices/SharedLibrary.h
```diff
@@ -35,186 +37,200 @@
...
 
 #ifdef _MSC_VER
 #  pragma warning(pop)
 #endif
 
 #endif // CPPMICROSERVICES_SHAREDLIBRARY_H
+
+// Local Variables:
+// c-file-style: "cppms"
+// End:
```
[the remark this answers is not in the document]
carneyweb: removed.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -952,3 +953,7 @@ std::shared_ptr<BundlePrivate> GetPrivate(const Bundle& b)
   return b.d;
 }
 }
+
+// Local Variables:
+// c-file-style: "cppms"
+// End:
```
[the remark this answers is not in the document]
carneyweb: removed.

.gitignore
```diff
@@ -47,3 +47,5 @@ third_party/googletest
 !third_party/googletest/googletest/cmake
 !third_party/googletest/googletest/include
 !third_party/googletest/googletest/src
+/maci64
```
karthikreddy09: I would recommend removing these ... it causes confusion and the naming is user specific anyway.

framework/include/cppmicroservices/Constants.h
```diff
@@ -21,11 +21,11 @@
 =============================================================================*/
 
 #ifndef CPPMICROSERVICES_CONSTANTS_H
-#define CPPMICROSERVICES_CONSTANTS_H
+#  define CPPMICROSERVICES_CONSTANTS_H
 
-#include "cppmicroservices/FrameworkConfig.h"
+#  include "cppmicroservices/FrameworkConfig.h"
```
karthikreddy09: Is this the result of running clang-format ?? This style looks wierd. a space after #

framework/src/util/SharedLibrary.cpp
```diff
@@ -124,6 +123,21 @@ void SharedLibrary::Load(int flags)
 #endif
 }
 
+void SharedLibrary::Load(const std::unordered_map<std::string, Any>& config)
+{
+#ifdef US_PLATFORM_POSIX
+  int options = RTLD_LAZY | RTLD_LOCAL;
+  try {
+    options = any_cast<int>(config.at(Constants::LIBRARY_LOAD_OPTIONS));
+  } catch (...) {
```
karthikreddy09: I think it's better to use the `[]` API instead of the `at` since the option is guaranteed to exist based on what I understand from `InitProperties`.  If that is not true, use `find` and avoid exceptions from `at`

As far as the exceptions from any_cast, we should log the exception instead of swallowing it.

framework/src/bundle/CoreBundleContext.cpp
```diff
@@ -72,6 +76,10 @@ std::unordered_map<std::string, Any> InitProperties(
     std::string(CppMicroServices_VERSION_STR);
   configuration[Constants::FRAMEWORK_VENDOR] = std::string("CppMicroServices");
 
+#ifdef US_PLATFORM_POSIX
+  configuration[Constants::LIBRARY_LOAD_OPTIONS] = Any(RTLD_LAZY | RTLD_LOCAL);
```
karthikreddy09: Isn't insert the right API to use here? Looks like this overrides user provided value

.gitignore
carneyweb: Done.

framework/include/cppmicroservices/Constants.h
carneyweb: Done.

framework/src/util/SharedLibrary.cpp
carneyweb: Done.

framework/src/bundle/CoreBundleContext.cpp
carneyweb: Done.

framework/src/util/SharedLibrary.cpp
carneyweb: I was unable to change this to use the [] API instead of **at** as the config object is const, and the [] API is not, while the **at** API is.

carneyweb:
@saschazelzer Sorry about the churn with all the submissions. I believe this is now in a pull-able state. Do you have any further issues or can these changes now be pulled in?

framework/src/util/SharedLibrary.cpp
```diff
@@ -124,6 +123,21 @@ void SharedLibrary::Load(int flags)
 #endif
 }
 
+void SharedLibrary::Load(const std::unordered_map<std::string, Any>& config)
```
saschazelzer: I suggest to remove this `Load` overload, we can already pass in int options via `Load(int)`. This keeps the SharedLibrary class simpler.

framework/src/util/SharedLibrary.cpp
```diff
@@ -124,6 +123,21 @@ void SharedLibrary::Load(int flags)
...
+{
+  int options = 0;
+#ifdef US_PLATFORM_POSIX
+  try {
+    options = any_cast<int>(config.at(Constants::LIBRARY_LOAD_OPTIONS));
```
saschazelzer: I think we should do this in `CoreBundleContext::Init()` and save the int options in a member variable. And instead of silently using defaults in case of an exception, we should explicitly check the Any type and log (via `DIAG_LOG`) that the expected type is wrong and we fall back to defaults.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -591,7 +591,7 @@ std::exception_ptr BundlePrivate::Start0()
         libHandle = BundleUtils::GetExecutableHandle();
       } else {
         if (!lib.IsLoaded()) {
-          lib.Load();
+          lib.Load(coreCtx->frameworkProperties);
```
saschazelzer: Here, you could use the new member of `coreCtx` which holds the `int` options and use the existing `SharedLibrary::Load(int)` function.

framework/src/util/SharedLibrary.cpp
carneyweb: Done.

framework/src/util/SharedLibrary.cpp
carneyweb: Done.

framework/src/bundle/BundlePrivate.cpp
carneyweb: Done.

Issue #307, opened 2018-10-16
Expose checksum from zip archive.


mw-cwilcox:
I'm requesting an enhancement, to expose a checksum of the bundle data.

### Motivation

My team is using CppMicroServices in deployment where DLLs may be loaded, unloaded, and reloaded through the process lifetime, and would like to utilize a checksum to know if a bundle has changed from a prior load.  There is already a 'crc32' in the zip-data, so this seems straightforward to implement.

### Approach

- ('miniz.c' provides 'struct mz_zip_archive_file_stat', which has-a 'mz_uint32 m_crc32;' and is always set in 'mz_zip_reader_file_stat')
- Add a new field to 'BundleResourceContainer::Stat', perhaps 'uint32_t crc32', or 'unsigned int checksum'; the field would be assigned by 'BundleResourceContainer::GetStat', which already has the value inside 'struct mz_zip_archive_file_stat'.
- Add a new getter to 'BundleResource', such as 'uint32_t GetCrc32() const;' or 'unsigned int GetChecksum() const;'.

### Branch

I intend to prototype exactly what I described here on a fork in a feature branch.

mw-cwilcox:
Here's what I was thinking, using 'uint32_t' for the type, and 'GetCrc32' for the method name.

ksubramz:
Thanks Charles. The motivation and the changes you made in your feature branch look fine to me. I am awaiting feedback from other core developers.

jeffdiclemente:
@mw-cwilcox looks good to me. please go ahead and issue a PR to the CppMicroServices development branch.

If you could also add the contents of this issue as a RFC, using the process in https://github.com/CppMicroServices/rfcs, that would be very helpful.

Thanks!

jeffdiclemente:
Fixed in #308

PR #308, opened 2018-10-17
307 bundle crc32

mw-cwilcox:
Here's the PR for Issue [#307](https://github.com/CppMicroServices/CppMicroServices/issues/307)

### DIFF
--- a/framework/test/driver/BundleResourceTest.cpp
+++ b/framework/test/driver/BundleResourceTest.cpp
@@ -161,6 +161,9 @@ void testInvalidResource(const Bundle& bundle)
 
   US_TEST_CONDITION(res.GetChildren().empty(), "Check empty children")
   US_TEST_CONDITION(res.GetSize() == 0, "Check zero size")
+  US_TEST_CONDITION(res.GetCompressedSize() == 0, "Check zero compressed size")
+  US_TEST_CONDITION(res.GetLastModified() == 0, "Check zero last modification time")
+  US_TEST_CONDITION(res.GetCrc32() == 0, "Check zero CRC-32")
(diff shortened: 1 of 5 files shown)


framework/test/driver/BundleResourceTest.cpp
```diff
@@ -161,6 +161,9 @@ void testInvalidResource(const Bundle& bundle)
 
   US_TEST_CONDITION(res.GetChildren().empty(), "Check empty children")
   US_TEST_CONDITION(res.GetSize() == 0, "Check zero size")
+  US_TEST_CONDITION(res.GetCompressedSize() == 0, "Check zero compressed size")
+  US_TEST_CONDITION(res.GetLastModified() == 0, "Check zero last modification time")
+  US_TEST_CONDITION(res.GetCrc32() == 0, "Check zero CRC-32")
```
ksubramz: It'd be great if we have a positive test for non-zero CRC-32. Maybe a good test point to add that is in `testCompressedResource`. Something like

```cpp
 US_TEST_CONDITION(res.GetCrc32() == XXXX, "Check valid CRC-32")
```
Apart from that I have no flags from this submission. Thanks!

mw-cwilcox: Hi Krish, I didn't add any positive tests, because it looked like everything else used the helper 'checkResourceInfo', I didn't want to have to force *all* the positive tests to now need to provide the CRC-32, and I didn't have a sense for which specific test should have a CRC-32 check.

jeffdiclemente:
@saschazelzer can you review this? thanks!

Issue #320, opened 2019-01-22
AnyMap assignment doesn't copy the type


ksubramz:
The implementation of `any_map` assignment operator doesn't copy the type. So, the following piece of code

```cpp
  AnyMap o1(AnyMap::ORDERED_MAP);
  o1["key"] = 42;
  AnyMap uo1(AnyMap::UNORDERED_MAP);
  uo1 = o1;
  ASSERT_EQ(any_cast<int>(uo1.at("key")), 42);
```
crashes because `uo1`'s type is still `UNORDERED_MAP` and the internal getter gets a nullptr. Performing an `at` dereferences the nullptr

I am fixing it as part of fix for #319

Issue #319, opened 2019-01-21
modernize-pass-by-value change makes installing bundles with deeply nested manifests really slow


ksubramz:
In commit 6cf07ab26ccd3a447217232f020a27803559dd32, the following change was made to `Any::Holder`

```diff
-    Holder(const ValueType& value)
-      : _held(value)
+    Holder(ValueType  value)
+      : _held(std::move(value))
```

This is Clang's modernize-pass-by-value change (https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html). This idiom essentially eliminates the need to write explicit copy and move constructors and is generally beneficial.

However, this is not true for every case. Consider the case for bundles with a deeply nested manifest (an example pathological case is provided below).  When we parse this bundle, (`BundleManifest::ParseJson{Object,Value}`), we build the data-structure to be an `Any` whose underlying object is an `AnyMap` which is basically `map<std::string, Any>`. This is a recursive data-structure which mirrors the JSON hierarchy. 

`Any` stores the underlying `AnyMap` inside the `_held` member variable which is of type `Holder<AnyMap>` (and so on for every level in the hierarchy). The problem here is that we don't provide a move constructor for `AnyMap`. This results in constructor of `Holder` copying the `AnyMap` from the call site to `value` and then copying the `value` to `_held`. This gets into a recursion where copies of `AnyMap` are made unnecessarily in a deeply nested way. 

I counted the number of times the `AnyMap` copy constructor is called during the `InstallBundles()` for the manifest below:
 - before the change : 675 times
 - after the change : 268435398 times

Certainly needs fixing.

**Solution:**
1. There needs to be a fix inside `Any::Holder` to have explicit move and copy constructors for maximum efficiency.
2. `AnyMap` should have a `noexcept` move constructor to optimize cases where temporary `AnyMap` objects are involved.

**The pathological manifest:**
```json
{
   "Test_AtCompoundKey" : {
      "relativelylongkeyname_element" : true,
      "relativelylongkeyname_map" : {
         "relativelylongkeyname_element" : true,
         "relativelylongkeyname_map" : {
            "relativelylongkeyname_element" : true,
            "relativelylongkeyname_map" : {
               "relativelylongkeyname_element" : true,
               "relativelylongkeyname_map" : {
                  "relativelylongkeyname_element" : true,
                  "relativelylongkeyname_map" : {
                     "relativelylongkeyname_element" : true,
[...]

PR #321, opened 2019-01-22
Fix regression caused due to modernize-pass-by-value

ksubramz:
Fix regression caused by #319

Also fix issue in AnyMap assignment #320

Refactor a bit of code in AnyMap


framework/src/util/AnyMap.cpp
```diff
@@ -892,6 +863,56 @@ any_map::unordered_any_cimap& any_map::uoci_m()
...
+      throw std::logic_error("invalid map type");
+  }
+}
+
+void any_map::move_from(any_map&& other) noexcept
+{
+  switch (other.type) {
+    case map_type::ORDERED_MAP:
+      map.o = other.map.o;
+      other.map.o = nullptr;
```
karthikreddy09: This seems safer but why do we care to perform this operation if the object is supposed to be discarded anyway? If the user tries to use the object after it has been moved, it's a logic error in the program. There is no guarantee that the object checks for null before performing an operation on it's member pointer anyway.

Don't pay for what you don't need to.

framework/test/bundles/libWithDeepManifest/resources/manifest.json
```diff
@@ -0,0 +1,78 @@
...
+                     }
+                  }
+               }
+            }
+         }
+      }
+   },
+   "bundle.activator" : true,
+   "bundle.symbolic_name" : "TestBundleWithDeepManifest",
+   "bundle.vendor" : "The MathWorks, Inc."
```
karthikreddy09: remove this

framework/test/driver/BundleManifestTest.cpp
```diff
@@ -59,6 +59,14 @@ void TestUnicodeProperty(BundleContext bc)
 #endif
 }
 
+void TestDeepManifest(BundleContext bc)
+{
+  auto bundle = testing::InstallLib(bc, "TestBundleWithDeepManifest");
+  auto headers = bundle.GetHeaders();
+  US_TEST_CONDITION(headers.at(Constants::BUNDLE_SYMBOLICNAME).ToString() ==
```
karthikreddy09: why check the symbolic name? whats the purpose of this test?

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -189,4 +195,21 @@ TEST(AnyMapTest, AnyMap)
...
+}
+
+TEST(AnyMapTest, Move)
+{
+  AnyMap o(AnyMap::ORDERED_MAP);
+  o["do"] = Any(1);
+  o["re"] = Any(2);
+  AnyMap o_anymap_move_ctor(std::move(o));
+  ASSERT_EQ(any_cast<int>(o_anymap_move_ctor.at("do")), 1);
+  ASSERT_DEATH(o.size(), ".*")
```
karthikreddy09: This behavior is not part of the C++ standard, so why do we need to implement it?

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -189,4 +195,21 @@ TEST(AnyMapTest, AnyMap)
...
+    << "This call should result in a crash because the object has been moved from";
+
+  AnyMap uo_anymap_move_assign(AnyMap::UNORDERED_MAP);
```
karthikreddy09: I would split this into two test points. One test point for move_ctor another for move_assign

framework/src/util/AnyMap.cpp
ksubramz: `o` is a pointer to the map. If this assignment to `nullptr` isn't done, the temporary `AnyMap`'s destructor will delete the map pointed to by `o`. Hence assigning this to `nullptr` so that the deletion doesn't occur.

framework/test/driver/BundleManifestTest.cpp
ksubramz: The purpose of this test is to see `InstallLib` call finishes quickly. Based on our offline discussions, I'll change the test name to illustrate what it's doing

framework/test/gtest/AnyMapTest.cpp
ksubramz: I wanted a test to confirm that we are actually moving the guts of `o` to `o_anymap_move_ctor` - so any operation of `o` must crash because it's underlying pointer is a `nullptr`. The original problem this PR is trying to fix is a copy was performed even when we explicitly say `std::move` (because there was no move ctor). I wanted to verify we actually fix that.

framework/src/util/AnyMap.cpp
ksubramz: Based on offline discussion, we both felt that the underlying pointers should be `unique_ptr`s. That will make the code more compact and idiomatic.

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -189,4 +195,36 @@ TEST(AnyMapTest, AnyMap)
...
+  o["do"] = Any(1);
+  o["re"] = Any(2);
+  AnyMap o_anymap_move_ctor(std::move(o));
+  ASSERT_EQ(any_cast<int>(o_anymap_move_ctor.at("do")), 1);
+  ASSERT_DEATH(o.size(), ".*")
+    << "This call should result in a crash because the object has been moved from";
+
+  AnyMap uo(AnyMap::UNORDERED_MAP);
+  AnyMap uo_move(std::move(uo));
+  ASSERT_EQ(uo_move.size(), 0) << "Size of an empty moved-to AnyMap must be 0";
```
jeffdiclemente: on gcc and clang you'll get a signed/unsigned compiler failure.

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -189,4 +195,36 @@ TEST(AnyMapTest, AnyMap)
...
+
+  AnyMap uoci(AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS);
+  AnyMap uoci_move(std::move(uoci));
+  ASSERT_EQ(uoci_move.size(), 0) << "Size of an empty moved-to AnyMap must be 0";
```
jeffdiclemente: on gcc and clang you'll get a signed/unsigned compiler failure.

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -209,11 +209,11 @@ TEST(AnyMapTest, MoveConstructor)
 
   AnyMap uo(AnyMap::UNORDERED_MAP);
   AnyMap uo_move(std::move(uo));
-  ASSERT_EQ(uo_move.size(), 0) << "Size of an empty moved-to AnyMap must be 0";
+  ASSERT_EQ(uo_move.size(), static_cast<std::size_t>(0)) << "Size of an empty moved-to AnyMap must be 0";
```
jeffdiclemente: use an integer suffix instead of the static_cast.
https://en.cppreference.com/w/cpp/language/integer_literal

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -209,11 +209,11 @@ TEST(AnyMapTest, MoveConstructor)
...
 
   AnyMap uoci(AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS);
   AnyMap uoci_move(std::move(uoci));
-  ASSERT_EQ(uoci_move.size(), 0) << "Size of an empty moved-to AnyMap must be 0";
+  ASSERT_EQ(uoci_move.size(), static_cast<std::size_t>(0)) << "Size of an empty moved-to AnyMap must be 0";
```
jeffdiclemente: use an integer suffix instead of the static_cast.
https://en.cppreference.com/w/cpp/language/integer_literal

PR #323, opened 2019-01-23
322 getheaders change returntype

karthikreddy09:
improve performance of GetHeaders API

### DIFF
--- a/framework/include/cppmicroservices/Bundle.h
+++ b/framework/include/cppmicroservices/Bundle.h
@@ -426,6 +426,9 @@ public:
    * This method continues to return Manifest header information while
    * this bundle is in the \c UNINSTALLED state.
    *
+   * \note The lifetime of the returned reference is bound to the lifetime of 
+   * this \c Bundle object.
+   *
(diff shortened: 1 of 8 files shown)


framework/include/cppmicroservices/Bundle.h
```diff
@@ -430,7 +430,7 @@ class US_Framework_EXPORT Bundle
    *
    * @see Constants#BUNDLE_LOCALIZATION
    */
-  AnyMap GetHeaders() const;
+  const AnyMap& GetHeaders() const;
```
jeffdiclemente: a note describing the lifetime guarantee for the returned ```AnyMap&``` would be helpful to users.

ksubramz:
Discussed with Abhinay, the only flag I had was someone hanging on that returned reference and accessing after the bundle is destroyed (lifetime issue). Abhinay, however, explicitly details the lifetime in the API comments.
An alternative worth thinking about is returning a `shared_ptr<const AnyMap>` that solves the lifetime issues/read-only-ness, but sacrifices a bit of performance.

Issue #327, opened 2019-03-21
Add GoogleMock to the "thirdparty" folder


karthikreddy09:
It would be great if we had GoogleMock along with GoogleTest. we can use mock objects to test classes like the ServiceFactory etc.

karthikreddy09:
Fixed by #330

PR #330, opened 2019-03-25
327 add gmock

karthikreddy09:
Add gmock framework into the "thirdparty" folder so that we can use mock objects in our tests.


karthikreddy09:
Fixes #327

karthikreddy09:
Fixes #327

PR #335, opened 2019-03-27
Add RapidJSON to third_party folder

karthikreddy09:

### DIFF
--- a/third_party/README
+++ b/third_party/README
@@ -138,3 +138,10 @@ Patches
 
  * Suppress CMake warning about policy CMP0048
  * Suppress tr1 namespace deprecation warnings
+
+ RapidJSON
+ ---------
+
+ MIT License
+
+ https://github.com/Tencent/rapidjson/tree/version1.1.0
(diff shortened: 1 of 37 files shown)


third_party/README
```diff
@@ -138,3 +138,10 @@ Patches
 
  * Suppress CMake warning about policy CMP0048
  * Suppress tr1 namespace deprecation warnings
+
+ RapidJSON
+ ---------
+
+ MIT License
+
+ https://github.com/Tencent/rapidjson/tree/version1.1.0
```
saschazelzer: I would prefer the same style as above: Pointing to the release page and state the version number in a separate line below.

karthikreddy09: Updated to match existing style

karthikreddy09: @saschazelzer Could you please take a look at the latest changes and provide your feedback. I have other changes I want to publish, which depend on this PR.

karthikreddy09:
fixes #335

PR #341, opened 2019-04-03
335 add json validator

### LINKED ISSUE #336: Create a tool to validate JSON files using JSON schema files
Users have run into runtime failures when manifest.json was syntactically correct but had missing information such as "symbolic_name". It's better to detect these errors early in the build step rather than during runtime.

karthikreddy09:
fixes #336

### DIFF
--- a/tools/jsonschemavalidator/main.cpp
+++ b/tools/jsonschemavalidator/main.cpp
@@ -26,6 +26,7 @@ limitations under the License.
 #include "rapidjson/schema.h"
 #include "rapidjson/stringbuffer.h"
 #include "rapidjson/prettywriter.h"
+#include <memory>
 #include <iostream>
 
 using namespace rapidjson;
--- a/tools/jsonschemavalidator/main.cpp
+++ b/tools/jsonschemavalidator/main.cpp
@@ -23,6 +23,7 @@ limitations under the License.
 #include <fstream>   // std::ifstream
 #include <iostream>  // std::cerr
 #include <memory>    // std::unique_ptr
+#include <sstream>   // std::sstream
 #include <stdexcept> // std::runtime_error
 #include <string>    // std::to_string
 
@@ -77,14 +78,15 @@ std::pair<bool, std::string> validate(const std::string& jsonfile,
     rapidjson::SchemaValidator validator(sd);
     rapidjson::Document json = ReadJSONDocument(jsonfile);
     if (!json.Accept(validator)) {
-      std::string errorMsg("JSON file " + jsonfile +
-                           " does not match schema file " + schemafile);
       rapidjson::StringBuffer sb;
       rapidjson::PrettyWriter<rapidjson::StringBuffer> w(sb);
       validator.GetError().Accept(w);
-      errorMsg.append("\nError report:\n");
-      errorMsg.append(sb.GetString());
-      throw std::runtime_error(errorMsg);
+      std::stringstream errStream;
+      errStream << "JSON file " << jsonfile << " does not match schema file "
+                << schemafile << std::endl;
+      errStream << "Error report:" << std::endl;
+      errStream << sb.GetString() << std::endl;
+      throw std::runtime_error(errStream.str());
     }
   } catch (const std::exception& e) {
     return std::make_pair(false, e.what());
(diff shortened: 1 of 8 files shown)


tools/jsonschemavalidator/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+ */
+Document ReadJSONDocument(const std::string& filePath)
+{
+  std::unique_ptr<FILE, FILEDeleter> fp(fopen(filePath.c_str(), "r"));
+  if (!fp)
+  {
+    std::string err("File '" + filePath + "' not found");
+    throw std::runtime_error(err);
+  }
+  char buffer[4096];
```
jeffdiclemente: should the json document be limited to 4096 bytes? Can [BasicIStreamWrapper](http://rapidjson.org/classrapidjson_1_1_basic_i_stream_wrapper.html) be used?

tools/jsonschemavalidator/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+  {
+    SchemaDocument sd(ReadJSONDocument(schemafile));
+    SchemaValidator validator(sd);
+    Document json = ReadJSONDocument(jsonfile);
+    if (!json.Accept(validator))
+    {
+      std::string errorMsg("JSON file " + jsonfile + " does not match schema file " + schemafile);
+      StringBuffer sb;
+      PrettyWriter<StringBuffer> w(sb);
+      validator.GetError().Accept(w);
```
jeffdiclemente: The error report output isn't terribly intuitive nor informative to CppMicroServices bundle developers. I wonder if there is a way to translate the error report. Does rapidjson document the error report format?

karthikreddy09: Agree, the report is not intuitive. This is the only info I found on reporting ... http://rapidjson.org/md_doc_schema.html#Reporting. I am not sure if it's worth spending time on parsing the error report and generating a user friendly string given the website states the possibility of changing the report format in the future.

jeffdiclemente: I'm ok with leaving the error report as is for now since there appears to be no reliable way to parse it into a user friendly message.

tools/jsonschemavalidator/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+    "json-file",
+    Custom_Arg::NonEmpty,
+    " --json-file, -j \tJSON file to be validated" },
+  { 0, 0, 0, 0, 0, 0 }
+};
+
+int main(int argc, char *argv[])
+{
+  argc -= (argc > 0);
+  argv += (argc > 0); // skip program name in argv[0]
```
carneyweb: Are these two line correct? It looks like it's possible for argc to decrement without argv incrementing. For instance, what if argc is 1? Then the first line sets it to zero, and the second line makes argv NOT increment over the program name.

tools/jsonschemavalidator/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include "rapidjson/stringbuffer.h"
+#include "rapidjson/prettywriter.h"
+#include <memory>
+#include <iostream>
+#include <string>
+
+using namespace rapidjson;
+
+// functor object for deleting FILE objects
+struct FILEDeleter
```
carneyweb: Should this be called FILECloser? The name implies that the file will be deleted from the file system.

tools/jsonschemavalidator/main.cpp
carneyweb: What Jeff said :-)

tools/jsonschemavalidator/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+std::pair<bool, std::string> validate(const std::string& jsonfile, const std::string& schemafile) noexcept
+{
+  try
+  {
+    SchemaDocument sd(ReadJSONDocument(schemafile));
+    SchemaValidator validator(sd);
+    Document json = ReadJSONDocument(jsonfile);
+    if (!json.Accept(validator))
+    {
+      std::string errorMsg("JSON file " + jsonfile + " does not match schema file " + schemafile);
```
carneyweb: Should errorMsg be a stringstream? Appending is faster.

tools/jsonschemavalidator/main.cpp
karthikreddy09: removed!

tools/jsonschemavalidator/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+      StringBuffer sb;
+      PrettyWriter<StringBuffer> w(sb);
+      validator.GetError().Accept(w);
+      errorMsg.append("\nError report:\n");
```
carneyweb: should the "\n" be std::endl to ensure the display is right across all platforms?

tools/jsonschemavalidator/main.cpp
karthikreddy09: good catch!

karthikreddy09:
fixes #341

Issue #263, opened 2018-03-08
Invalid Bundle causes crash on method invocation


karthikreddy09:
A default constructed `Bundle` object is an invalid object. Calling a method on an invalid Bundle object such as `GetBundleId` causes a crash. I think it should throw and let the user handle the exception instead of a hard crash.

gitabhishekdp:
I think its true for all the getters that can be called on default constructed Bundle.

PR #346, opened 2019-04-16
263 Fix for Invalid Bundle causes crash on method invocation

gitabhishekdp:
Fix for invalid Bundle causes crash on method invocation #263


gitabhishekdp:
> 
> 
> There are more functions where, if called on a default constructed `Bundle` object, it would crash. I think that in any function where `d` is accessed, a check should happen.

There are getters in the Bundle interface which has US_DEPRECATED. I added a check for those too, but during the unit testing, the compilation failed because these functions/getters are marked as deprecated. 

What do you recommend we do for such functions ?

jeffdiclemente:
> > There are more functions where, if called on a default constructed `Bundle` object, it would crash. I think that in any function where `d` is accessed, a check should happen.
> 
> There are getters in the Bundle interface which are marked as US_DEPRECATED. I added a check for those too, but during the unit testing, the compilation failed because these functions/getters are marked as deprecated.
> 
> What should we do for such functions ?

@gitabhishekdp the deprecation warnings as errors are turned off in usFrameworkTestDriver and not for the gtest unit tests. I'd suggest adding the following from ```framework/test/driver/CMakeLists.txt``` into the gtest CMakeLists.txt file:
```
# Disable deprecation warnings.
if (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") OR
    ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang") OR
    ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU"))
  set_source_files_properties(${_srcs} PROPERTIES COMPILE_FLAGS -Wno-error=deprecated-declarations)
endif()
```

PR #350, opened 2019-05-03
Integrate google benchmark

karthikreddy09:
Added google benchmark to third_party folder and created a new target for benchmark tests


framework/test/CMakeLists.txt
```diff
@@ -2,3 +2,4 @@
 add_subdirectory(bundles)
 add_subdirectory(driver)
 add_subdirectory(gtest)
+add_subdirectory(bench)
```
carneyweb: Don't know if the lack of a carriage return will affect things with this file. Sometimes it's a problem.

karthikreddy09: seems to build fine on all platforms

framework/test/bench/BenchTest.cpp
```diff
@@ -0,0 +1,119 @@
+#include "benchmark/benchmark.h"
```
jeffdiclemente: I'd suggest naming this source file something more specific to what its meant to benchmark.

CMakeLists.txt
```diff
@@ -340,6 +340,11 @@ if(US_BUILD_TESTING)
     # static library even if CppMicroServices is a shared library.
     set(BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
     add_subdirectory(third_party/googletest)
+
+    set(BENCHMARK_ENABLE_INSTALL OFF)
+    set(BENCHMARK_ENABLE_TESTING OFF CACHE BOOL "" FORCE)
```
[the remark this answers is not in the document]
karthikreddy09: These are Google's variable. we are setting them to disable test and install actions for the benchmark target

PR #352, opened 2019-05-09
Fix Makefile builds on macOS

karthikreddy09:
Fix the builds using CMake "Unix Makefiles" on macOS

### DIFF
--- a/cmake/CMakeResourceDependencies.cpp
+++ b/cmake/CMakeResourceDependencies.cpp
@@ -20,14 +20,12 @@
 
 =============================================================================*/
 
-#include <iostream>
-
 namespace {
 
 // This is dummy code to silence some linkers warning about
 // empty object files.
 struct CMakeResourceDependencies
 {
-  CMakeResourceDependencies() { std::cout << std::flush; }
+  CMakeResourceDependencies() { }
 };
 }


jeffdiclemente:
After talking offline with Abhinay about why these changes fix the issue, two questions fell out of the conversation:

1. why isn't CMakeResourceDependencies.cpp built with the same compiler arguments as the rest of the code base?
2. Is there a reason why CMakeResourceDependencies.cpp needs to exist at all?

These two questions may undercover more work that is needed. However for the time being, I'm ok with this change to fix the failure.

cmake/CMakeResourceDependencies.cpp
```diff
@@ -20,14 +20,12 @@
...
-#include <iostream>
-
 namespace {
 
 // This is dummy code to silence some linkers warning about
 // empty object files.
 struct CMakeResourceDependencies
 {
-  CMakeResourceDependencies() { std::cout << std::flush; }
+  CMakeResourceDependencies() { }
```
saschazelzer: The `std::cout` code was added to silence `has no symbols` warnings issued by `ranlib` on MacOS, see commit
bc72d205201930ab9baef8bc203395d548e600f7. Looking at the CI build output, those warnings now appear again.

`ranlib` seems to have a command line option now to suppress those warnings. Maybe we can make use of them?

carneyweb: Added an issue to deal with the warning.

PR #359, opened 2019-05-14
Remove unnecessary copy of interface map

karthikreddy09:

### DIFF
--- a/framework/src/bundle/BundleContext.cpp
+++ b/framework/src/bundle/BundleContext.cpp
@@ -34,13 +34,13 @@
 #include "ServiceReferenceBasePrivate.h"
 #include "ServiceRegistry.h"
 
+#include <cstdio>
 #include <memory>
 #include <utility>
-#include <cstdio>
 
 namespace cppmicroservices {
 
-BundleContext::BundleContext(std::shared_ptr<BundleContextPrivate>  ctx)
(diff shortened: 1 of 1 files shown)


framework/src/bundle/BundleContext.cpp
```diff
@@ -297,17 +297,10 @@ InterfaceMapConstPtr BundleContext::GetService(
   // but we ignore it since the time window is small and
   // the result is the same as if the calling thread had
   // won the race condition.
-
-  // Although according to the API contract the returned map should not be modified, there is nothing stopping the consumer from
-  // using a const_pointer_cast and modifying the map. This copy step is to protect the map stored within the framework.
```
saschazelzer: Why is that comment not valid anymore? Are you not concerned about clients anymore const-casting the map? Would be fine with me.

karthikreddy09: It's similar to the scenario where an API returns a const reference and the user const_cast's it and modifies the object. At that point, it's a breach of contract and the user has to deal with the consequences/undefined behavior.

saschazelzer: Well, I agree. I believe it was like that before we introduced the copy. So the concerns in the comment are gone, right?

karthikreddy09: yes

PR #361, opened 2019-05-29
Perf manifest parsing

carneyweb:
Improve normal performance by only initializing the deprecated properties if the deprecated apis are used by doing a deep copy of m_Headers into m_PropertiesDeprecated. Otherwise m_PropertiesDeprecated is never created. These changes will be deleted completely when the deprecated apis are completely removed.


framework/src/bundle/BundleManifest.cpp
```diff
@@ -143,34 +177,41 @@ bool BundleManifest::Contains(const std::string& key) const
...
+  if (m_Headers.cend() != iter)
+  {
     return iter->second;
   }
   return Any();
 }
 
 Any BundleManifest::GetValueDeprecated(const std::string& key) const
 {
+  copy_deprecated_properties(m_Headers, m_PropertiesDeprecated);
```
jeffdiclemente: there is a possible race condition if more than one thread calls a deprecated bundle property API. I'd suggest using ```std::call_once``` for all the calls to ```copy_deprecated_properties```

framework/src/bundle/BundleManifest.cpp
```diff
@@ -143,34 +177,41 @@ bool BundleManifest::Contains(const std::string& key) const
...
+  if (m_PropertiesDeprecated.cend() != iter)
+  {
     return iter->second;
   }
   return Any();
 }
 
 std::vector<std::string> BundleManifest::GetKeysDeprecated() const
 {
+  copy_deprecated_properties(m_Headers, m_PropertiesDeprecated);
```
jeffdiclemente: there is a possible race condition if more than one thread calls a deprecated bundle property API. I'd suggest using ```std::call_once``` for all the calls to ```copy_deprecated_properties```

framework/src/bundle/BundleManifest.cpp
```diff
@@ -143,34 +177,41 @@ bool BundleManifest::Contains(const std::string& key) const
...
+       ++iter)
+  {
     keys.push_back(iter->first);
   }
   return keys;
 }
 
 std::map<std::string, Any> BundleManifest::GetPropertiesDeprecated() const
 {
+  copy_deprecated_properties(m_Headers, m_PropertiesDeprecated);
```
jeffdiclemente: there is a possible race condition if more than one thread calls a deprecated bundle property API. I'd suggest using ```std::call_once``` for all the calls to ```copy_deprecated_properties```

framework/src/bundle/BundleManifest.cpp
```diff
@@ -143,34 +177,41 @@ bool BundleManifest::Contains(const std::string& key) const
...
   std::vector<std::string> keys;
-  for (AnyMap::const_iterator iter = m_PropertiesDeprecated.begin();
-       iter != m_PropertiesDeprecated.end();
-       ++iter) {
+  for (AnyMap::const_iterator iter = m_PropertiesDeprecated.cbegin();
+       iter != m_PropertiesDeprecated.cend();
+       ++iter)
+  {
```
jeffdiclemente: The formatting of the if statement does not follow convention.

framework/src/bundle/BundleManifest.cpp
```diff
@@ -143,34 +177,41 @@ bool BundleManifest::Contains(const std::string& key) const
 Any BundleManifest::GetValue(const std::string& key) const
 {
   auto iter = m_Headers.find(key);
-  if (iter != m_Headers.end()) {
+  if (m_Headers.cend() != iter)
+  {
```
jeffdiclemente: The formatting of the if statement does not follow convention.

cmake/CMakeResourceDependencies.cpp
```diff
@@ -20,14 +20,12 @@
```
jeffdiclemente: was this file accidentally brought in to the PR? This change doesn't appear to be relevant to manifest parsing.

framework/src/bundle/BundleManifest.cpp
carneyweb: Done.

framework/src/bundle/BundleManifest.cpp
carneyweb: Done.

framework/src/bundle/BundleManifest.cpp
carneyweb: Done.

framework/src/bundle/BundleManifest.cpp
carneyweb: Done.

framework/src/bundle/BundleManifest.cpp
carneyweb: Done.

cmake/CMakeResourceDependencies.cpp
carneyweb: I believe this came in as part of a merge of 349_Fix_Makefile_builds_on_macos.

jeffdiclemente: this file should be removed from the PR then.

framework/test/gtest/BundleManifestTest.cpp
```diff
@@ -36,6 +36,59 @@
```
jeffdiclemente: this file should be formatted using the clang.format file in the root source directory

cmake/CMakeResourceDependencies.cpp
carneyweb: Hopefully it's now removed. I restored the state of the file to be as it was before the merge.

PR #362, opened 2019-05-30
Add benchmarks for ServiceTrackers

jeffdiclemente:


framework/test/bench/fooservice.h
```diff
@@ -0,0 +1,14 @@
+
+// A dummy interface and impl to use with service trackers
+namespace googlebenchmark {
```
karthikreddy09: please remove google

PR #363, opened 2019-06-10
Port BaT benchmarks

carneyweb:
Ported the benchmarks from existing MathWorks benchmarks.


framework/test/bench/AnyMapPerfTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+class AnyMapPerfTestFixture : public ::benchmark::Fixture
+{
+public:
+  using benchmark::Fixture::SetUp;
+  using benchmark::Fixture::TearDown;
+
+  void SetUp(const ::benchmark::State&)
+  {
+    using namespace cppmicroservices;
```
jeffdiclemente: the cppmicroservices namespace was already brought in on line 16

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -149,7 +149,7 @@ TEST(AnyMapTest, AnyMap)
   AnyMap uco_anymap(uco);
 
   AnyMap o_anymap1(AnyMap::ORDERED_MAP);
-  o_anymap1 = o_anymap1;
```
jeffdiclemente: why is this commented?

framework/test/bench/AnyMapPerfTest.cpp
carneyweb: Done.

framework/test/gtest/AnyMapTest.cpp
carneyweb: Removed comment out line.

framework/src/util/AnyMap.cpp
```diff
@@ -146,7 +146,7 @@ Any AtCompoundKey(const std::vector<Any>& v,
         ref_any_cast<std::vector<Any>>(h), tail, std::move(defaultval));
     }
   }
-  return defaultval;
+  return std::move(defaultval);
```
jeffdiclemente: what's the purpose of these changes? They don't seem to have any relation to the benchmark tests.

carneyweb: Without that change I get the following warning/error:

      local variable 'defaultVal' will be copied despite being returned by name
      [-Werror,-Wreturn-std-move]

PR #367, opened 2019-06-18
Improve ServiceTracker performance

jeffdiclemente:
Improve performance by using std::unordered_map to track services

Use std::shared_ptr instead of ExplicitlySharedDataPointer

fix typo in type name

### DIFF
--- a/framework/test/driver/ServiceTrackerTest.cpp
+++ b/framework/test/driver/ServiceTrackerTest.cpp
@@ -33,9 +33,9 @@
 #include "TestingConfig.h"
 #include "TestingMacros.h"
 
+#include <chrono>
 #include <future>
 #include <memory>
-#include <chrono>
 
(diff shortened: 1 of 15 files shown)


framework/test/driver/ServiceTrackerTest.cpp
```diff
@@ -33,33 +33,25 @@
...
 
+#include <chrono>
 #include <future>
 #include <memory>
-#include <chrono>
 
 using namespace cppmicroservices;
 
 bool CheckConvertibility(const std::vector<ServiceReferenceU>& refs,
                          std::vector<std::string>::const_iterator idBegin,
```
karthikreddy09: Why take iterators when the copy is going to happen anyway. Why not const std::vector& as a parameter and construct the vector at the call site rather than inside this method

PR #368, opened 2019-06-21
Fix any map cihash

carneyweb:
Updated the AnyMap cihash algorithm to use std::hash to produce hash values.

### DIFF
--- a/framework/include/cppmicroservices/AnyMap.h
+++ b/framework/include/cppmicroservices/AnyMap.h
@@ -34,25 +34,12 @@ namespace detail {
 
 struct any_map_cihash
 {
-  std::size_t operator()(const std::string& key) const
-  {
-    std::string lcase = key;
-    std::transform(lcase.begin(), lcase.end(), lcase.begin(), ::tolower);
-    return std::hash<std::string>{}(lcase);
-  }
+  std::size_t operator()(const std::string& key) const;
 };
 
 struct any_map_ciequal
 {
-  bool operator()(const std::string& l, const std::string& r) const
-  {
-    return (l.size() == r.size()
-            && std::equal(l.begin(), l.end(), r.begin()
-                          , [](char a, char b)
-                            {
-                              return tolower(a) == tolower(b);
-                            }));
-  }
+  bool operator()(const std::string& l, const std::string& r) const;
 };
 
 }
--- a/framework/include/cppmicroservices/AnyMap.h
+++ b/framework/include/cppmicroservices/AnyMap.h
@@ -32,12 +32,12 @@ namespace cppmicroservices {
 
 namespace detail {
 
-struct any_map_cihash
+struct US_Framework_EXPORT any_map_cihash
 {
   std::size_t operator()(const std::string& key) const;
 };
 
-struct any_map_ciequal
+struct US_Framework_EXPORT any_map_ciequal
 {
   bool operator()(const std::string& l, const std::string& r) const;
 };
--- a/framework/include/cppmicroservices/AnyMap.h
+++ b/framework/include/cppmicroservices/AnyMap.h
@@ -34,13 +34,27 @@ namespace detail {
 
 struct any_map_cihash
 {
-  std::size_t operator()(const std::string& key) const;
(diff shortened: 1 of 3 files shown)


framework/src/util/AnyMap.cpp
```diff
@@ -23,27 +23,13 @@
 #include "cppmicroservices/AnyMap.h"
 
 #include <stdexcept>
+#include <algorithm>
+#include <functional>
```
saschazelzer: Just looking at the diff, it is not clear to me why those include directives were added.

framework/include/cppmicroservices/AnyMap.h
```diff
@@ -34,13 +34,27 @@ namespace detail {
 
 struct any_map_cihash
 {
-  std::size_t operator()(const std::string& key) const;
+  std::size_t operator()(const std::string& key) const
```
saschazelzer: Is there a good reason for moving the definitions to the header file?

carneyweb: Moved inline because I added a test that checks a couple of simple things about the hash, and that required that the symbol either be exported from the cpp file, or that the hash algorithm be inline. I found it easier to deal with inline, so I did that. If that's a problem I can make the change to export the symbol instead.

framework/src/util/AnyMap.cpp
carneyweb: Removed. They were there in an earlier iteration that had the implementation for any_map_cihash::operator() in the .cpp file instead of the header.

framework/include/cppmicroservices/AnyMap.h
saschazelzer: I am not a proponent of exporting symbols or putting implementation details (definitions) in headers to support tests (unless there is a really good reason). In this case, I think you could test the hash function implicitly by using a any_map with special test values.

carneyweb: Ordinarily I'd agree. However, I think it's important that we test the hash algorithm (which was not being tested before) itself, and not the use of the algorithm in the code. That is, we should treat the hash algorithm as the unit under test and test it directly. Also, The hash algorithm is in the header file for AnyMap which has implementation details (because of all the templated code). Revealing the particulars of the cihash in comparison to other details that are visible there seems to be of minor concern.

So, in sum, I think it's important that we test the hash algorithm, and I don't think it's worth a lot of code reorganization for the very few lines of implementation code exposed.

saschazelzer: In this particular case, I disagree. There is no code reorganization required (actually, the proposed changes need more reorganization - moving the implementation to the header). The other details which are already exposed are out of necessity (if not, I would advocate to move them as well).

The simple tests for hash equality and non-equality are equivalent to testing if the unordered_map contains the case-variant key or not. This *is* testing the hash algorithm.

If you want to test the hash class directly, why not use `std::unordered_map<...>::hasher` instead of exposing its details?

carneyweb: I just tried using any_map::unordered_any_cimap::hasher in the tests and that gives me an unresolved symbol because the the operator() isn't exported. I can add the US_Framework_EXPORT macro to the any_map_cihash and any_map_ciequal structs and that fixes the symbol problem. Would that be preferable to you? If so, I can submit that change.

As far as testing the map for whether or not it contains the key, while this has the effect of testing the hash algorithm, that's testing it as a side effect. I think it's better in general to actually write unit tests to exercise exactly the code under test rather than trying to do so by manipulating the calling code.

saschazelzer: Thanks for trying `hasher`, adding the export macro is okay for me.

I agree in general with what you say. In our case, it affects the library API surface, potential backwards compatibility, (I know we do not provide strong guarantees for that), compile dependencies, etc. But I think we found a good compromise here.

carneyweb: OK. I just pushed these changes in. Once the ci builds finish we can pull the changes into development. Thanks for taking the time to review :-)

PR #377, opened 2019-07-22
Don't install gmock and google benchmark libs

jeffdiclemente:
Installing gtest/gmock/benchmark libs has side-effects for downstream clients who are using their own gtest/gmock/benchmark libs.


jeffdiclemente:
@CppMicroServices/developers If there are no flags by next Monday, I will merge this into development.

Issue #337, opened 2019-03-29
Make DS code buildable.


karthikreddy09:
The code on branch "54_declarative_services" is currently not buildable. Add CMake functions to make it buildable.

karthikreddy09:
Here's the makefile code snippet that needs to be translated into a cmake function that needs to be called from a bundle's CMakeList file

```
# variable to store the name of the auto-generated cpp file
US_DS_CPP_FILE_NAME := autogen_ServiceComponentGlue.cpp
# variable to indicate if the module uses declarative services
MICROSERVICE_USES_DECLARATIVE_SERVICES ?=
MICROSERVICE_HAS_ACTIVATOR ?= $(if $(MICROSERVICE_USES_DECLARATIVE_SERVICES),false,true)

# SERVICE_COMPONENTS_HEADER_PATH contains path to the header file which is
# an aggregate of all the service implementation headers in the module
# SERVICE_COMPONENTS_HEADER_PATH is a relative path from moduleroot
SERVICE_COMPONENTS_HEADER_PATH ?= ServiceComponents.hpp
SERVICE_COMPONENTS_HEADER := $(notdir $(SERVICE_COMPONENTS_HEADER_PATH))

SCRCODEGEN_TOOL := $(BUILD_ROOT)/bin/$(ARCH)/SCRCodeGen$(EXE_EXT)

# target to create the auto generated file for Declarative Services
$(call GENERATED_SRC_FILE_PATH,$(US_DS_CPP_FILE_NAME)) : $(SCRCODEGEN_TOOL) $(US_MOD_MANIFEST_JSON_FILE_PATH) $(MAKEFILE_LIST) $(SERVICE_COMPONENTS_HEADER_PATH)
	@echo "==== Generating $(BOLD_ON)$@$(BOLD_OFF) ===="
	@$(RM) $@
	$(SCRCODEGEN_TOOL) --manifest $(US_MOD_MANIFEST_JSON_FILE_PATH) --out-file $@ --include-headers $(SERVICE_COMPONENTS_HEADER)
```

jeffdiclemente:
fixed in PR #383

PR #378, opened 2019-08-08
Build Declarative Services

jeffdiclemente:
These are the minimum set of changes needed to build and run all tests for Declarative Services.

There is most likely more work to enable users to generate DS bundles, however this is not the goal of this PR.

Fixes: #337 


jeffdiclemente:
Closing as this code, plus additional DS changes will be tracked in PR #383.

PR #382, opened 2019-09-06
Update per feedback and fix complier warning

Jeeyuen-Rick:
Update tests per Jeff's feedback and fixed complier waning on Mac.

### DIFF
--- a/framework/test/gtest/BundleEventTest.cpp
+++ b/framework/test/gtest/BundleEventTest.cpp
@@ -35,12 +35,18 @@ class BundleEventTest : public ::testing::Test
 protected:
   Bundle            bundle;
   BundleEvent       bEvt;
+  Framework         f;
   FrameworkEvent    fEvt;
   BundleEvent::Type eventType;
 
+public:
+  BundleEventTest()
+    : f(FrameworkFactory().NewFramework()){};
+
+  ~BundleEventTest() override = default;
+
   virtual void SetUp()
   {
-    auto f = FrameworkFactory().NewFramework();
     f.Start();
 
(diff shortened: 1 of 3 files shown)


framework/test/gtest/BundleEventTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+    eventType = bEvt.GetType();
+  }
+
+  virtual void TearDown()
+  {}
+};
+
+TEST_F(BundleEventTest, invalidBundleEvents)
+{
+  const auto eventBundle = bEvt.GetBundle();
```
jeffdiclemente: to create an invalid bundle use the ```Bundle``` default constructor. You don't need ```bEvt``` in this case.

framework/test/gtest/BundleEventTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+  ASSERT_THROW(BundleEvent(eventType, eventBundle),
+               std::invalid_argument);
+}
+
+TEST_F(BundleEventTest, invalidBundleOrigin)
+{
+  const auto eventBundle = fEvt.GetBundle();
+  const auto eventOrigin = bEvt.GetOrigin();
+  ASSERT_THROW(BundleEvent(eventType, eventBundle, eventOrigin),
```
jeffdiclemente: to create an invalid bundle use the Bundle default constructor. You don't need ```bEvt``` and ```fEvt``` in this case.

Jeeyuen-Rick: I need a valid bundle in this case else the invalid bundle exception throws before the invalid origin.

PR #384, opened 2019-09-11
Fixed #217, Updates framework exception severity levels and uses ServiceExceptions

### LINKED ISSUE #217: Use correct framework event severity and exception types for service factory errors

achristoforides:
This PR addresses issue #217 which requested a review of framework exception severity and usage of the already-existing ServiceException class in lieu of std::logic_error as stated by the spec.

By looking through the docs for OSGi, I identified the proper exception types and modified the currently used exceptions in the code. The same was done for std::logic_errors in the correct areas to better match the spec as well.

Two tests (BundleTest and ServiceFactoryTest) had to be updated to reflect the newly introduced state into the software's internal logic. For both tests, it was a matter of updating the type of exceptions or expected exception messages so that they matched the source code changes made.

### DIFF
--- a/framework/src/service/ServiceReferenceBasePrivate.cpp
+++ b/framework/src/service/ServiceReferenceBasePrivate.cpp
@@ -107,7 +107,7 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceFromFactory(
       FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
                      MakeBundle(bundle->shared_from_this()),
                      message,
-                     std::make_exception_ptr(ServiceException(message.c_str(),
+                     std::make_exception_ptr(ServiceException(message,
                        ServiceException::Type::FACTORY_EXCEPTION))));
   }
   return s;
--- a/framework/src/service/ServiceReferenceBasePrivate.cpp
+++ b/framework/src/service/ServiceReferenceBasePrivate.cpp
@@ -76,7 +76,7 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceFromFactory(
           FrameworkEvent::Type::FRAMEWORK_ERROR,
           MakeBundle(bundle->shared_from_this()),
           message,
-          std::make_exception_ptr(ServiceException(message.c_str(), 
+          std::make_exception_ptr(ServiceException(message, 
             ServiceException::Type::FACTORY_ERROR)
           )));
       return smap;
@@ -100,14 +100,14 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceFromFactory(
       }
     }
     s = smap;
-  } catch (...) {
+  } catch (std::exception& ex) {
     s.reset();
     std::string message = "ServiceFactory threw an unknown exception.";
     registration->bundle->coreCtx->listeners.SendFrameworkEvent(
       FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
                      MakeBundle(bundle->shared_from_this()),
                      message,
-                     std::make_exception_ptr(ServiceException(message,
+                     std::make_exception_ptr(ServiceException(ex.what(),
                        ServiceException::Type::FACTORY_EXCEPTION))));
   }
   return s;
@@ -279,13 +279,13 @@ bool ServiceReferenceBasePrivate::UngetPrototypeService(
       try {
         sf->UngetService(
           MakeBundle(bundle), ServiceRegistrationBase(registration), service);
-      } catch (...) {
+      } catch (std::exception& ex) {
         std::string message("ServiceFactory threw an exception");
         registration->bundle->coreCtx->listeners.SendFrameworkEvent(
           FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
                          MakeBundle(bundle->shared_from_this()),
                          message,
-                         std::make_exception_ptr(ServiceException(message,
+                         std::make_exception_ptr(ServiceException(ex.what(),
                            ServiceException::Type::FACTORY_EXCEPTION))));
       }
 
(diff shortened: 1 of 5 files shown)


framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -100,11 +102,13 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceFromFactory(
...
   } catch (...) {
     s.reset();
+    std::string message = "ServiceFactory threw an unknown exception.";
     registration->bundle->coreCtx->listeners.SendFrameworkEvent(
       FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
                      MakeBundle(bundle->shared_from_this()),
-                     std::string("ServiceFactory threw an unknown exception."),
-                     std::current_exception()));
+                     message,
+                     std::make_exception_ptr(ServiceException(message,
```
jeffdiclemente: Since an exception was thrown, we want to preserve the exception message and use that string as the ```message``` in the ```ServiceException``` object. Otherwise we are losing information which may be useful to users.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -278,10 +282,11 @@ bool ServiceReferenceBasePrivate::UngetPrototypeService(
       } catch (...) {
         std::string message("ServiceFactory threw an exception");
         registration->bundle->coreCtx->listeners.SendFrameworkEvent(
-          FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING,
+          FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
                          MakeBundle(bundle->shared_from_this()),
                          message,
-                         std::current_exception()));
+                         std::make_exception_ptr(ServiceException(message,
```
jeffdiclemente: Since an exception was thrown, we want to preserve the exception message and use that string as the ```message``` in the ```ServiceException``` object. Otherwise we are losing information which may be useful to users.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -358,10 +363,11 @@ bool ServiceReferenceBasePrivate::UngetService(
     } catch (...) {
       std::string message("ServiceFactory threw an exception");
       registration->bundle->coreCtx->listeners.SendFrameworkEvent(
-        FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING,
+        FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
                        MakeBundle(bundle->shared_from_this()),
                        message,
-                       std::current_exception()));
+                       std::make_exception_ptr(ServiceException(message,
```
jeffdiclemente: Since an exception was thrown, we want to preserve the exception message and use that string as the ```message``` in the ```ServiceException``` object. Otherwise we are losing information which may be useful to users.

framework/src/service/ServiceRegistrationBase.cpp
```diff
@@ -256,10 +256,11 @@ void ServiceRegistrationBase::Unregister()
           std::string message(
             "ServiceFactory UngetService implementation threw an exception");
           d->bundle->coreCtx->listeners.SendFrameworkEvent(
-            FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING,
+            FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
                            MakeBundle(d->bundle->shared_from_this()),
                            message,
-                           std::current_exception()));
+                           std::make_exception_ptr(ServiceException(message,
```
jeffdiclemente: Since an exception was thrown, we want to preserve the exception message and use that string as the ```message``` in the ```ServiceException``` object. Otherwise we are losing information which may be useful to users.

framework/src/service/ServiceRegistrationBase.cpp
```diff
@@ -273,10 +274,11 @@ void ServiceRegistrationBase::Unregister()
         std::string message(
           "ServiceFactory UngetService implementation threw an exception");
         d->bundle->coreCtx->listeners.SendFrameworkEvent(
-          FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING,
+          FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
                          MakeBundle(d->bundle->shared_from_this()),
                          message,
-                         std::current_exception()));
+                         std::make_exception_ptr(ServiceException(message,
```
jeffdiclemente: Since an exception was thrown, we want to preserve the exception message and use that string as the ```message``` in the ```ServiceException``` object. Otherwise we are losing information which may be useful to users.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -98,13 +100,15 @@ InterfaceMapConstPtr ServiceReferenceBasePrivate::GetServiceFromFactory(
       }
     }
     s = smap;
-  } catch (...) {
+  } catch (std::exception& ex) {
```
jeffdiclemente: ```std::exception&``` should be ```const```.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -275,13 +279,14 @@ bool ServiceReferenceBasePrivate::UngetPrototypeService(
       try {
         sf->UngetService(
           MakeBundle(bundle), ServiceRegistrationBase(registration), service);
-      } catch (...) {
+      } catch (std::exception& ex) {
```
jeffdiclemente: ```std::exception&``` should be ```const```.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -355,13 +360,14 @@ bool ServiceReferenceBasePrivate::UngetService(
     try {
       sf->UngetService(
         MakeBundle(bundle), ServiceRegistrationBase(registration), sfi);
-    } catch (...) {
+    } catch (std::exception& ex) {
```
jeffdiclemente: ```std::exception&``` should be ```const```.

framework/src/service/ServiceRegistrationBase.cpp
```diff
@@ -252,14 +252,15 @@ void ServiceRegistrationBase::Unregister()
         try {
           serviceFactory->UngetService(
             MakeBundle(i.first->shared_from_this()), *this, service);
-        } catch (...) {
+        } catch (std::exception& ex) {
```
jeffdiclemente: ```std::exception&``` should be ```const```.

framework/src/service/ServiceRegistrationBase.cpp
```diff
@@ -269,14 +270,15 @@ void ServiceRegistrationBase::Unregister()
       try {
         serviceFactory->UngetService(
           MakeBundle(i.first->shared_from_this()), *this, i.second);
-      } catch (...) {
+      } catch (std::exception& ex) {
```
jeffdiclemente: ```std::exception&``` should be ```const```.

PR #387, opened 2019-09-19
Add tests for Bundle.cpp to improve code coverage

Jeeyuen-Rick:
Add tests for the stream operator.

### DIFF
--- a/framework/test/gtest/BundleStreamOperatorTest.cpp
+++ b/framework/test/gtest/BundleStreamOperatorTest.cpp
@@ -1,0 +1,57 @@
+/*=============================================================================
+
+  Library: CppMicroServices 
+
+  Copyright (c) The CppMicroServices developers. See the COPYRIGHT
+  file at the top-level directory of this distribution and at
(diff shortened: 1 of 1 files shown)


framework/test/gtest/BundleStreamOperatorTest.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+  std::stringstream buffer;
+  std::streambuf* backup = std::cout.rdbuf(buffer.rdbuf());
+  std::cout << bundleState;
+  std::string actStr = buffer.str();
+  ASSERT_TRUE(actStr.find(expectedStr) != std::string::npos);
+  std::cout.rdbuf(backup);
+}
+
+TEST(BundleStreamOperatorTest, streamStateUninstalled)
```
jeffdiclemente: instead of starting each testpoint with "stream", i'd suggest using "bundle", i.e. ```bundleStateUninstalled```. This more accurately reflects what is being tested.

PR #389, opened 2019-09-23
Optimization of AtCompoundKey and LDAP Queries

achristoforides:
This PR introduces a small subset of Abseil code in to the repo to help optimize the AtCompoundKey function and LDAP queries. Additionally, a few modifications to the CMake files are made to allow the abseil files to build and be correctly included in the project for use.


framework/src/CMakeLists.txt
```diff
@@ -90,4 +106,40 @@ set(_private_headers
...
+# Turn off warnings when compiling Abseil files
+file(GLOB_RECURSE ABSL_FILES 
+  "${CMAKE_SOURCE_DIR}/third_party/absl/*.cc" 
+  "${CMAKE_SOURCE_DIR}/third_party/absl/*.c" 
+  "${CMAKE_SOURCE_DIR}/third_party/absl/*.h")
+
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR 
+    "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" 
+    OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  set(CF "-w")
```
jeffdiclemente: I'd suggest renaming ```CF``` to something more self-documenting.

framework/src/CMakeLists.txt
```diff
@@ -90,4 +106,40 @@ set(_private_headers
...
+  set_source_files_properties(
```
jeffdiclemente: I'd suggest removing the existing -Werror and replacing it with -w only for the abseil source files.

framework/src/util/AnyMap.cpp
```diff
@@ -123,18 +135,18 @@ Any AtCompoundKey(const AnyMap& m,
...
 {
   auto pos = key.find(".");
   const auto head = key.substr(0, pos);
   const auto tail = (pos == AnyMap::key_type::npos) ? "" : key.substr(pos + 1);
+
   int index = 0;
-  try {
-    index = std::stoi(head);
-  } catch (...) {
+  bool res = absl::SimpleAtoi(head, &index);
```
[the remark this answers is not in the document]
achristoforides: Converting the existing code to use std::stoi with string construction is actually just as fast, if not faster. I'll use std::stoi instead here.

CMakeLists.txt
```diff
@@ -363,6 +363,10 @@ set(US_CXX_COVERAGE_FLAGS )
 if(MSVC)
   set(disabled_warnings
     4251 # 'identifier' : class 'type' needs to have dll-interface to be used by clients of class 'type2'
+    4244 # type conversions
+    4996 # unsafe functions
+    4267
```
kevinleeMW: could we add comments briefly describing the last two warnings as well?

framework/src/util/AnyMap.cpp
```diff
@@ -35,74 +38,74 @@ std::size_t any_map_cihash::operator()(const std::string& key) const
...
+const Any& AtCompoundKey(const std::vector<Any>& v,
+                         const absl::string_view& key)
 {
   auto pos = key.find(".");
   if (pos != AnyMap::key_type::npos) {
     auto head = key.substr(0, pos);
     auto tail = key.substr(pos + 1);
 
-    const int index = std::stoi(head);
+    int index = std::stoi(std::string(head));
```
jeffdiclemente: should be ```const```

framework/src/util/AnyMap.cpp
```diff
@@ -35,74 +38,74 @@ std::size_t any_map_cihash::operator()(const std::string& key) const
...
     } else if (h.Type() == typeid(std::vector<Any>)) {
       return AtCompoundKey(ref_any_cast<std::vector<Any>>(h), tail);
     }
-    throw std::invalid_argument("Unsupported Any type at '" + head +
-                                "' for dotted get");
+    throw std::invalid_argument("Unsupported Any type at '" +
+                                std::string(head) + "' for dotted get");
   } else {
-    const int index = std::stoi(key);
+    int index = std::stoi(std::string(key));
```
jeffdiclemente: should be ```const```

jeffdiclemente:
@saschazelzer do you have any flags for this? If not i'm planning on merging this in on Monday. Thanks.

PR #390, opened 2019-09-26
Scope file handles to BundleResource and BundleResourceStream life-times, resolves #309

### LINKED ISSUE #309: Scope file handle to lifetime of BundleResource and BundleResourceStream
This is a follow up enhancement which came from discussions in issue #299.

The idea is to keep the file handle open for the lifetime of the ```BundleResource``` and ```BundleResourceStream``` objects. The file handle will be closed once the last instance is destroyed. 
Both these classes provide access to bundle resources and if no instances exist, there is no reason to maintain an open file handle.

achristoforides:
This PR introduces changes which resolve issue #309 .

Instead of closing BundleResource file handles at the end of the Bundle life-time, they are now cleaned up if there are no opened resources in a given bundle/bundle container. To achieve this, a reference counting method is used to keep track of how many opened resources there are associated with the given bundle. If this number reaches 0, then the container is closed (thus closing the file handles). If a user requests a resource after the bundle has been closed, it is then reopened and their resource is then served all while keeping track of the reference count.


jeffdiclemente (changes requested):
Tests need to be added to ```framework\test\gtest\OpenFileHandleTest.cpp``` for this new behavior.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -878,6 +878,9 @@ BundlePrivate::BundlePrivate(CoreBundleContext* coreCtx,
     state = Bundle::STATE_RESOLVED;
   }
   */
+
+  // Close the resource container after the bundle has been made
+  barchive->GetResourceContainer()->CloseContainer();
```
jeffdiclemente: This should be removed.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

jeffdiclemente (approved):

jeffdiclemente:
@saschazelzer do you have any flags for this? If not i'm planning on merging this in on Monday. Thanks.

framework/src/bundle/BundleArchive.cpp
```diff
@@ -82,9 +83,15 @@ BundleResource BundleArchive::GetResource(const std::string& path) const
   if (!resourceContainer) {
     return BundleResource();
   }
+
+  if (!resourceContainer->IsContainerOpen()) {
+    resourceContainer->OpenContainer();
+  }
```
jeffdiclemente: This isn't necessary since constructing the ```BundleResource``` will call ```OpenContainer()``` if necessary.

framework/src/bundle/BundleArchive.cpp
```diff
@@ -98,6 +105,10 @@ std::vector<BundleResource> BundleArchive::FindResources(
     return result;
   }
 
+  if (!resourceContainer->IsContainerOpen()) {
+    resourceContainer->OpenContainer();
+  }
```
jeffdiclemente: This isn't necessary since constructing the ```BundleResource``` will call ```OpenContainer()``` if necessary.

jeffdiclemente (changes requested):
I looked over the code one more time and noticed a couple things.

framework/src/bundle/BundleResource.cpp
```diff
@@ -61,6 +60,28 @@ class BundleResourcePrivate
   std::atomic<int> ref;
 };
 
+BundleResourcePrivate::BundleResourcePrivate(std::shared_ptr<const BundleArchive> archive)
+  : archive(std::move(archive))
+  , ref(1)
+{
+  if (this->archive) {
+    if (!this->archive->resourceContainer->IsContainerOpen())
+      this->archive->resourceContainer->OpenContainer();
```
jeffdiclemente: This can be removed. Constructing ```BundleResource``` will open the container if necessary.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

jeffdiclemente (changes requested):
The code coverage failure can be solved with a slightly different solution. If the default member initialization of numOpenResources is moved to the constructor and the friend classes are removed, at least locally on my machine, this fixes the code coverage error.

I still don't know why this solves the issue though.

My next step is to isolate the friend class change and see if that causes the same code coverage failure.

jeffdiclemente:
> The gcov failure can be solved with a slightly different solution. If the default member initialization of numOpenResources is moved to the constructor and the friend classes are removed, at least locally on my machine, this fixes the gcov error.
> 
> I still don't know why this solves the issue though.
> 
> My next step is to isolate the friend class change and see if that causes the same gcov failure.

adding back the friend class (and not the default member initialization of numOpenResources) does **not** produce the same gcov file containing ~20 lines containing ```/*EOF*/``` (which ultimately leads to cmake failing).

framework/src/bundle/BundleResource.cpp
```diff
@@ -61,6 +60,25 @@ class BundleResourcePrivate
...
+    this->archive->IncrementNumOpenResources();
+  }
+}
+
+BundleResourcePrivate::~BundleResourcePrivate()
+{
+  if (this->archive) {
+    this->archive->DecrementNumOpenResources();
+
+    if (this->archive->GetNumOpenResources() == 0)
```
jeffdiclemente: single line if statement should still have braces.

jeffdiclemente (changes requested):

framework/src/bundle/BundleArchive.h
```diff
@@ -156,11 +156,29 @@ struct BundleArchive : std::enable_shared_from_this<BundleArchive>
...
+   * @return the number of opened resources.
+   */
+  unsigned int GetNumOpenResources() const;
+
 private:
   BundleStorage* const storage;
   const std::unique_ptr<Data> data;
   const std::shared_ptr<BundleResourceContainer> resourceContainer;
   const std::string resourcePrefix;
+  mutable unsigned int numOpenResources;
```
jeffdiclemente: this should be an ```atomic<uint32_t>``` since a race can occur if more than one thread tries to create ```BundleResource``` object.

framework/src/bundle/BundleResource.cpp
```diff
@@ -61,6 +60,26 @@ class BundleResourcePrivate
...
+}
+
+BundleResourcePrivate::~BundleResourcePrivate()
+{
+  if (this->archive) {
+    this->archive->DecrementNumOpenResources();
+
+    if (this->archive->GetNumOpenResources() == 0) {
+      this->archive->GetResourceContainer()->CloseContainer();
+    }
```
jeffdiclemente: move this logic into the ```DecrementNumOpenResources()``` and either rename the function or document it to reflect the behavior.

jeffdiclemente (changes requested):

framework/src/bundle/BundleArchive.h
```diff
@@ -156,11 +157,32 @@ struct BundleArchive : std::enable_shared_from_this<BundleArchive>
...
+   *       associated with this BundleArchive object is closed.
+   */
+  void UnregisterOpenedResource() const;
+
+  /**
+   * Get the number of opened resources.
+   * 
+   * @return the number of opened resources.
+   */
+  unsigned int GetNumOpenResources() const;
```
jeffdiclemente: ```GetNumOpenResources()``` isn't needed anymore. it can be removed.

framework/src/bundle/BundleArchive.cpp
```diff
@@ -136,9 +141,27 @@ void BundleArchive::SetAutostartSetting(int32_t setting)
...
+
+void BundleArchive::RegisterOpenedResource() const 
+{ 
+  numOpenResources++; 
+}
+
+void BundleArchive::UnregisterOpenedResource() const 
+{ 
+  numOpenResources--;
+  if (numOpenResources == 0) {
```
jeffdiclemente: there is a race between lines 157 and 158 where one thread can decrement ```numOpenResources``` and before it executes line 158 another thread executes line 157. Even within line 158 there is a race between the time it takes to retrieve the value in ```numOpenResources``` and compare it to ```0```. The race can be solved by using https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange or by adding a mutex around the increment, decrement and comparison of ```numOpenResources``` (in which case you don't need to use ```std::atomic```).

jeffdiclemente (changes requested):

jeffdiclemente (commented):
performance should be measured within the google benchmark tests to see what the overhead is now that a mutex has been introduced.

jeffdiclemente (approved):

jeffdiclemente:
> performance should be measured within the google benchmark tests to see what the overhead is now that a mutex has been introduced.

I ran the two bundle install benchmarks against development and this branch. Here are the results:

Benchmark Test  | development Time   | development CPU   | achristoforides:309-scope-file-handle Time | achristoforides:309-scope-file-handle CPU 
------------------- | ---------- | ---------- | ---------- | --------------|
BundleInstallFixture/BundleInstallCppFramework/manual_time_mean  |     274950 ns  |     331298 ns   |      284231 ns   |    342625 ns
BundleInstallFixture/BundleInstallCppFramework/manual_time_median  |   273459 ns   |    328281 ns  |       282698 ns     |  339929 ns
BundleInstallFixture/BundleInstallCppFramework/manual_time_stddev    |  17049 ns   |     19412 ns   |      17422 ns      |  19760 ns
BundleInstallFixture/LargeBundleInstallCppFramework/manual_time_mean   |    271779 ns   |    336457 ns    |      276112 ns     |  333663 ns
BundleInstallFixture/LargeBundleInstallCppFramework/manual_time_median  |   269830 ns  | 335545 ns     |    274588 ns    |   331362 ns
BundleInstallFixture/LargeBundleInstallCppFramework/manual_time_stddev   |   16674 ns   |  19432 ns  |    15395 ns    |    17959 ns

jeffdiclemente:
I'd like to hold off on merging this for now. This change does regress bundle install performance which I believe is important.

achristoforides:
| Benchmark Test                                                         | development Time | development CPU | achristoforides:309-scope-file-handle Time | achristoforides:309-scope-file-handle CPU |
|------------------------------------------------------------------------|------------------|-----------------|--------------------------------------------|-------------------------------------------|
| BundleInstallFixture/BundleInstallCppFramework/manual_time_mean        | 158891 ns        | 175477 ns       | 157877 ns                                  | 175230 ns                                 |
| BundleInstallFixture/BundleInstallCppFramework/manual_time_median      | 158668 ns        | 175203 ns       | 157544 ns                                  | 174824 ns                                 |
| BundleInstallFixture/BundleInstallCppFramework/manual_time_stddev      | 1474 ns          | 1651 ns         | 1646 ns                                    | 1833 ns                                   |
| BundleInstallFixture/LargeBundleInstallCppFramework/manual_time_mean   | 163047 ns        | 180998 ns       | 161200 ns                                  | 178758 ns                                 |
| BundleInstallFixture/LargeBundleInstallCppFramework/manual_time_median | 162577 ns        | 180516 ns       | 160931 ns                                  | 178499 ns                                 |
| BundleInstallFixture/LargeBundleInstallCppFramework/manual_time_stddev | 1524 ns          | 1719 ns         | 1533 ns                                    | 1702 ns                                   |

I updated this PR by removing the mutex in BundleArchive and replacing it with a std::atomic implementation. The results in this table were generated by running the benchmark test with --benchmark_repetitions=256. As you can see, the timings are actually better compared to what they were before this change.

framework/src/bundle/BundleArchive.cpp
```diff
@@ -147,16 +148,16 @@ std::shared_ptr<BundleResourceContainer> BundleArchive::GetResourceContainer()
...
-void BundleArchive::UnregisterOpenedResource() const 
-{ 
-  std::lock_guard<std::mutex> lock(numOpenResourcesLock);
-  numOpenResources--;
+void BundleArchive::UnregisterOpenedResource() const
+{
+  uint32_t expected = 0;
+  while(!numOpenResources.compare_exchange_weak(expected, expected - 1));
   if (numOpenResources == 0) {
     resourceContainer->CloseContainer();
```
jeffdiclemente: unfortunately there's still a race between the time in which ```numOpenResources``` is atomically set and when it is evaluated by the if statement.

jeffdiclemente (changes requested):

achristoforides:
| Benchmark Test                                                         | development Time | development CPU | achristoforides:309-scope-file-handle Time | achristoforides:309-scope-file-handle CPU |
|------------------------------------------------------------------------|------------------|-----------------|--------------------------------------------|-------------------------------------------|
| BundleInstallFixture/BundleInstallCppFramework/manual_time_mean        | 395675 ns        | 450865 ns       | 399481 ns                                  | 451873 ns                                 |
| BundleInstallFixture/BundleInstallCppFramework/manual_time_median      | 374271 ns        | 439257 ns       | 371490 ns                                  | 442217 ns                                 |
| BundleInstallFixture/BundleInstallCppFramework/manual_time_stddev      | 46192 ns         | 56893 ns        | 50370 ns                                   | 64054 ns                                  |
| BundleInstallFixture/LargeBundleInstallCppFramework/manual_time_mean   | 388965 ns        | 454660 ns       | 402515 ns                                  | 453679 ns                                 |
| BundleInstallFixture/LargeBundleInstallCppFramework/manual_time_median | 367462 ns        | 443154 ns       | 378240 ns                                  | 442913 ns                                 |
| BundleInstallFixture/LargeBundleInstallCppFramework/manual_time_stddev | 47679 ns         | 59251 ns        | 51271 ns                                   | 66591 ns                                  |

Newly ran test (release mode, 512 iterations) comparing the mutex implementation vs. the development branch.

PR #394, opened 2019-10-11
Add tests for BundleResource.cpp

Jeeyuen-Rick:


framework/test/gtest/BundleResourceTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+{
+  BundleResource resource = bundleR.GetResource("icons/");
+  ASSERT_EQ(size(resource.GetChildResources()), 0);
+}
+
+TEST_F(BundleResourceTest, iterateChildNode)
+{
+  BundleResource resource = bundleR.GetResource("icons/");
+  resource.GetChildren();
+  ASSERT_EQ(size(resource.GetChildResources()), 3);
```
jeffdiclemente: you can also call the ```size()``` method of the object returned by ```GetChildResources()```.  ```std::size()``` only calls the ```size()``` method of the container.

framework/test/gtest/BundleResourceTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+};
+
+TEST_F(BundleResourceTest, operatorEqualTo)
+{
+  BundleResource a;
+  BundleResource b;
+  ASSERT_TRUE(a == b);
+}
+
+TEST_F(BundleResourceTest, invalidBundleResource)
```
jeffdiclemente: This is a pretty specific test case. Can the test case name be more specific?

framework/test/gtest/BundleResourceTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  virtual void TearDown()
+  {
+    f.Stop();
+    f.WaitForStop(std::chrono::milliseconds::zero());
+  }
+};
+
+TEST_F(BundleResourceTest, operatorEqualTo)
```
jeffdiclemente: its not necessary to start and stop the Framework. no bundles are being installed in this test case.

framework/test/gtest/BundleResourceTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+TEST_F(BundleResourceTest, invalidBundleResource)
+{
+  BundleResource result;
+  ASSERT_EQ(size(result.GetChildResources()), 0);
+}
+
+TEST_F(BundleResourceTest, emptyChildNode)
+{
+  BundleResource resource = bundleR.GetResource("icons/");
+  ASSERT_EQ(size(resource.GetChildResources()), 0);
```
jeffdiclemente: This doesn't seem right. There are 3 files in the ```icons``` folder. I would expect ```GetChildResources()``` to return ``3``. Especially since the call to ```resource.GetChildResources()``` in test case, iterateChildNode, returns 3.

Why does ```resource.GetChildResources()``` return 0 here and 3 in test case iterateChildNode?

PR #401, opened 2019-10-21
on ARM target suppress cast-align error and add libatomic

michkrom:
On ARM (RPi, Raspbian, GCC 8.3)  several reinterpret_cast<T*>(char*) cause cast-align warnings about need to increase alignment.  These warnings seem benign but due to -Werror they break the build. So for ARM target suppress this error but leave the warning (-Wno-error=cast-align).

Also, gcc on arm does not generate intrinsic for atomic operations but uses libatomic. Hence explicitly add libatomic to linker input for ARM.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,11 @@ if (CMAKE_CXX_COMPILER_VERSION AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS ${US_
   message(FATAL_ERROR "${US_COMPILER_${CMAKE_CXX_COMPILER_ID}_MINIMUM_VERSION_PRODUCT} or higher required (you are using ${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VERSION}).")
 endif()
 
+STRING(SUBSTRING "${CMAKE_SYSTEM_PROCESSOR}" 0 3 SYSTEM_PROCESSOR_SUBSTR)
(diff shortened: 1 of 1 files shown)


michkrom:
I confirm the Sign-off. But I have committed the changeset before adding this line to message. Should I  start the PR over?

Signed-off-by: Michal Krombholz krombholz@gmail.com

jeffdiclemente:
I believe you should be able to edit the original commit message and push it to this PR without having to start a new PR.

michkrom:
Commit message amended for proper sign-off.

CMakeLists.txt
```diff
@@ -474,6 +486,10 @@ if(CMAKE_COMPILER_IS_GNUCXX AND NOT APPLE)
       set(US_LINK_FLAGS "${US_LINK_FLAGS} ${_linkflag}")
     endif()
   endforeach()
+  if(ARM)
+    # atomic intrinsics are in libatomic on ARM GCC
+    set(US_LINK_FLAGS "${US_LINK_FLAGS} -latomic")
```
jeffdiclemente: Are there any other ARM compilers other than GCC? If so, do they have support for ```-latomic```?

If the intent is only to support ARM GCC compilers, that's fine with me, however I would prefer to have a stricter check for this on lines 114 - 117 and to reflect the restriction in the variable name (e.g. ```GCC_ARM``` instead of ```ARM```). This way other developers can understand the ARM toolchain supported should they want to add other ARM toolchain support.

PR #408, opened 2019-11-05
Gtest submodule migration

achristoforides:
This PR removes the gtest source from our repository and instead links to it using git submodules. This requires an extra "--recursive" flag when cloning the repository to also clone the submodules.


jeffdiclemente:
appveyor build needs to be updated to clone with ```--recursive```.

There should be instructions for developers in ```README.rst``` on how to clone and update the repo now that git submodules are being used.

PR #409, opened 2019-11-08
Service ctor exception crash

carneyweb:
If for some reason an exception is thrown in the constructor of a service implementation, a later call to ServiceObjectsBase::GetService() would throw an exception because the ServiceInterfaceMap is not populated for this service. Added a null check against the interfacemap to avoid this exception.

### DIFF
--- a/compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
+++ b/compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
@@ -25,7 +25,6 @@
 #include "TestUtils.hpp"
 #include "TestInterfaces/Interfaces.hpp"
 #include "cppmicroservices/ServiceEvent.h"
-#include "cppmicroservices/ServiceObjects.h"
 
 namespace sc  = cppmicroservices::service::component;
 namespace scr = cppmicroservices::service::component::runtime;
@@ -74,11 +73,7 @@ TEST_F(tServiceComponent, testThrowingLifeCycleHooks) //DS_TOI_9
 {
   auto testBundle = StartTestBundle("TestBundleDSTOI9");
   auto ctxt = framework.GetBundleContext();
-  auto const refs = ctxt.GetServiceReferences<test::LifeCycleValidation>();
   auto sRef = ctxt.GetServiceReference<test::LifeCycleValidation>();
-  auto serviceObjects = ctxt.GetServiceObjects<test::LifeCycleValidation>(sRef);
-  (void)serviceObjects.GetService();
-  (void)serviceObjects.GetServiceReference();
   ASSERT_TRUE(static_cast<bool>(sRef));
   auto service = ctxt.GetService<test::LifeCycleValidation>(sRef);
   EXPECT_EQ(service, nullptr) << "Service object must be nullptr since the service component should have thrown an exception when activated";
--- a/compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
+++ b/compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
@@ -25,6 +25,7 @@
 #include "TestUtils.hpp"
 #include "TestInterfaces/Interfaces.hpp"
 #include "cppmicroservices/ServiceEvent.h"
+#include "cppmicroservices/ServiceObjects.h"
 
 namespace sc  = cppmicroservices::service::component;
 namespace scr = cppmicroservices::service::component::runtime;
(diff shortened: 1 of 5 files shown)


compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -73,11 +74,14 @@ TEST_F(tServiceComponent, testThrowingLifeCycleHooks) //DS_TOI_9
 {
   auto testBundle = StartTestBundle("TestBundleDSTOI9");
   auto ctxt = framework.GetBundleContext();
+  auto const refs = ctxt.GetServiceReferences<test::LifeCycleValidation>();
```
jeffdiclemente: is this necessary? i don't see ```refs``` being used anywhere.

compendium/test_bundles/TestBundleDSTOI9/src/ServiceImpl.cpp
```diff
@@ -1,15 +1,26 @@
...
-  {
-    deactivated = true;
-    throw std::runtime_error("Exception thrown from user code");
-  };
-  ServiceComponent9::~ServiceComponent9() {};
+
+ServiceComponent9::ServiceComponent9()
+  : test::LifeCycleValidation()
+{
+  throw std::exception {};
```
jeffdiclemente: you should be able to test for this condition without DS by creating a ```ServiceFactory``` derived class which throws an exception when its constructed.

framework/src/service/ServiceObjects.cpp
```diff
@@ -138,20 +138,17 @@ std::shared_ptr<void> ServiceObjectsBase::GetService() const
     return nullptr;
   }
 
-  // interfaceMap can be null under certain circumstance
-  // e.g. if a constructor is thrown in the service implementation.
-  // In that case, we bail out early.
   auto interfaceMap = d->GetServiceInterfaceMap();
+  // g2113391: interfaceMap can be null under certain circumstance e.g. if a constructor
+  // is thrown in the service implementation.  In that case, we bail out early.
```
jeffdiclemente: you'll want to remove the issue #. its not relevant here.

framework/test/gtest/ServiceFactoryTest.cpp
```diff
@@ -140,15 +159,41 @@ TEST_F(ServiceFactoryTest, TestGetServiceThrows)
...
   context.RemoveListener(std::move(lToken));
 }
 
+// Tests the return type and FrameworkEvent generated when a
+// ServiceFactory instance throws an exception in it's GetService
+// callback
+TEST_F(ServiceFactoryTest, TestGetServiceObjectThrows)
+{
+  auto context = framework.GetBundleContext();
+  auto sf = std::make_shared<MockFactory>();
```
jeffdiclemente: shouldn't this be creating an instance of ```ThrowingFactory```?

framework/test/gtest/ServiceFactoryTest.cpp
```diff
@@ -140,15 +143,42 @@ TEST_F(ServiceFactoryTest, TestGetServiceThrows)
...
+                                                   Constants::SERVICE_SCOPE
+                                                   , Any(Constants::SCOPE_PROTOTYPE)
+                                                 }});
+
+  auto sref = context.GetServiceReference<ITestServiceA>();
+  auto serviceObjects = context.GetServiceObjects<ITestServiceA>(sref);
+
+  // Next line used to crash if a service implementation threw an exception in the
+  // constructor. This test case checks to make sure that the fix is in place.
+  (void)serviceObjects.GetService();
```
jeffdiclemente: probably want to wrap this in a ```EXPECT_NO_THROW``` so gtest can flag it as a failure.

carneyweb: Done.

framework/src/service/ServiceObjects.cpp
```diff
@@ -138,12 +138,18 @@ std::shared_ptr<void> ServiceObjectsBase::GetService() const
...
   }
 
-  std::shared_ptr<UngetHelper> h(
-    new UngetHelper(d->GetServiceInterfaceMap(),
-                    d->m_reference,
-                    d->m_context->bundle->shared_from_this()));
-  return std::shared_ptr<void>(
-    h, (h->interfaceMap->find(d->m_reference.GetInterfaceId()))->second.get());
+  auto interfaceMap = d->GetServiceInterfaceMap();
+  // interfaceMap can be null under certain circumstance e.g. if a constructor is thrown
```
jeffdiclemente: nitpick: a constructor is not thrown, it throws an exception.

PR #411, opened 2019-11-20
Improve BundleRegistry::Install() performance

achristoforides:
This PR introduces a new method for performing bundle installations which is more friendly to concurrency.

An implementation using maps to keep track of initial bundle installs is presented as a method for avoiding the single race condition (two threads trying to install a bundle that isn't already installed) present within the BundleRegistry::Install() operation. This method improves the amount of code that can be ran concurrently, thus improving the performance of installing bundles in a concurrent fashion.

### DIFF
--- a/framework/src/bundle/BundleRegistry.cpp
+++ b/framework/src/bundle/BundleRegistry.cpp
@@ -63,24 +63,6 @@ void BundleRegistry::Clear()
 }
 
 /*
-  A struct which contains the necessary objects to utilize condition
-  variables. A thread will wait on this WaitCondition if the waitFlag
-  is set to true.
-*/
-struct WaitCondition
-{
-  std::unique_ptr<std::mutex> m;
-  std::unique_ptr<std::condition_variable> cv;
-  bool waitFlag;
-
-  WaitCondition()
-    : m(std::make_unique<std::mutex>())
-    , cv(std::make_unique<std::condition_variable>())
-    , waitFlag(true)
-  {}
-};
-
-/*
   This function acquires a lock and decrements the reference count
   for the specified bundle. If this decrement causes the count to be
   0, it is erased from the initialBundleInstallMap map.
--- a/framework/src/bundle/BundleRegistry.cpp
+++ b/framework/src/bundle/BundleRegistry.cpp
@@ -62,41 +62,180 @@ void BundleRegistry::Clear()
   bundles.v.clear();
 }
 
+/*
+  A struct which contains the necessary objects to utilize condition
+  variables. A thread will wait on this WaitCondition if the waitFlag
+  is set to true.
+*/
+struct WaitCondition
+{
+  std::unique_ptr<std::mutex> m;
+  std::unique_ptr<std::condition_variable> cv;
+  bool waitFlag;
+
+  WaitCondition()
+    : m(std::make_unique<std::mutex>())
+    , cv(std::make_unique<std::condition_variable>())
+    , waitFlag(true)
+  {}
+};
+
+/*
+  This function acquires a lock and decrements the reference count
+  for the specified bundle. If this decrement causes the count to be
+  0, it is erased from the initialBundleInstallMap map.
+*/
+void BundleRegistry::DecrementInitialBundleMapRef(
+  cppmicroservices::detail::MutexLockingStrategy<>::UniqueLock& l,
+  const std::string& location)
+{
+  l.Lock();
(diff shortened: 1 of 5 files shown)


framework/test/bench/bundleinstall.cpp
```diff
@@ -45,6 +61,66 @@ class BundleInstallFixture
...
+      for (uint32_t j = 0; j < uint32_t(numOfIterations); j++, currentIndex++) {
+        tempListOfBundles.push_back(str5kBundles[currentIndex]);
+      }
+
+      if (i == numThreads - 1) {
+        tempListOfBundles.push_back(str5kBundles[0]);
+      }
+
+      bundlesToInstallPerThread.push_back(tempListOfBundles);
+    }
```
jeffdiclemente: is there a STL algorithm that can do this? to avoid having to write loops.

framework/src/bundle/BundleRegistry.cpp
```diff
@@ -62,36 +62,170 @@ void BundleRegistry::Clear()
...
+        // Wait for the install thread to notify this thread that it is safe
+        // to install the current bundle
+        std::unique_lock<std::mutex> lock(
+          *(initialBundleInstallMap[location].second.m));
+        initialBundleInstallMap[location].second.cv->wait(
+          lock, [&location, this] {
+            return !initialBundleInstallMap[location].second.waitFlag;
+          });
+
+        lock.unlock();
```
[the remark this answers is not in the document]
achristoforides: A std::unique_lock is required. https://en.cppreference.com/w/cpp/thread/condition_variable/wait

framework/test/bench/bundleinstall.cpp
achristoforides: Unfortunately no. I rewrote it so that it is cleaner and uses the std::vector<T>(iterator, iterator) constructor.

framework/src/bundle/BundleRegistry.cpp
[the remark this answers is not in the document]
jeffdiclemente: If I'm reading https://en.cppreference.com/w/cpp/thread/unique_lock/~unique_lock correctly, ```lock.unlock()``` is redundant since destruction of the ```std::unqiue_lock``` object will unlock the mutex.

framework/test/bench/bundleinstall.cpp
```diff
@@ -45,6 +62,66 @@ class BundleInstallFixture
...
+    auto framework = cppmicroservices::FrameworkFactory().NewFramework();
+    framework.Start();
+
+    // Generate paths to each bundle
+    std::vector<std::string> str5kBundles;
+    for (uint32_t count = 1; count <= 5000; count++) {
+      std::string copiedBundlePath = std::string(bundleBasePath);
+      copiedBundlePath.append(std::to_string(count));
+      str5kBundles.push_back(copiedBundlePath);
+    }
```
jeffdiclemente: I'd suggest using one of the STL algorithms instead of writing a loop.

framework/src/bundle/BundleRegistry.cpp
```diff
@@ -62,36 +62,169 @@ void BundleRegistry::Clear()
...
     res.insert(res.end(), newBundles.begin(), newBundles.end());
     if (res.empty()) {
       throw std::runtime_error("All bundles rejected by a bundle hook");
     } else {
       return res;
     }
+  } else {
+    // Check to see if another thread is currently in the process of installing this
+    // bundle for the first time.
+    auto p = initialBundleInstallMap.find(location);
```
carneyweb: What is "p"? Would a longer, more descriptive name make it more obvious?

framework/src/bundle/BundleRegistry.cpp
```diff
@@ -197,31 +199,30 @@ std::vector<Bundle> BundleRegistry::Install(const std::string& location,
         // to install the current bundle
         std::unique_lock<std::mutex> lock(
           *(initialBundleInstallMap[location].second.m));
-        initialBundleInstallMap[location].second.cv->wait(
-          lock, [&location, this] {
-            return !initialBundleInstallMap[location].second.waitFlag;
-          });
+        while (!initialBundleInstallMap[location].second.cv->wait_for(lock, 0.1ms, [&location, this] {
+          return !initialBundleInstallMap[location].second.waitFlag;
+          }));
```
jeffdiclemente: a comment describing why this fixes the deadlock would be helpful.

PR #420, opened 2019-12-12
Fixes #419, Removes files with incompatible licenses

### LINKED ISSUE #419: How to handle license conflict (Apache-2.0) with the applicable licenses?
Hello Contributors,

CPP-Microservices (3.3.0) comes under Apache-2.0 but we found a license conflict with these LGPL-2.1 files:

[https://github.com/CppMicroServices/CppMicroServices/blob/v3.3.0/third_party/cppmicroservices_elf.h] (LGPL-2.1+)

[https://github.com/CppMicroServices/CppMicroServices/blob/v3.3.0/framework/include/cppmicroservices/SharedData.h] (LGPL-2.1)

[https://github.com/CppMicroServices/CppMicroServices/blob/v3.3.0/third_party/cppmicroservices_mach-o.h] (BSD-4-Clause-UC)

How would you suggest to solve this issue?
Can the files be removed?

Many thanks,

Regards
Raghavendra

achristoforides:
This PR is in reference to issue #419 . It removes the dependencies on the listed files.

Fixes #419 .


saschazelzer:
I welcome moving away from `ExplicitlySharedDataPointer`, replacing it with `std::shared_ptr`.

However, without having reviewed all the details of this PR, replacing `SharedDataPointer` with `std::shared_ptr` is generally not equivalent. The first is implicitly copy-on-write, while the latter is semantically analogous to `ExplicitlySharedDataPointer`.

Removing unused code like the elf / mach header is okay, but please do not replace them with OS specific include directives. The original goal was to make this code compilable on any platform, by intentionally embedding relevant struct definitions. The license is not a problem.

achristoforides:
Thanks for your response.

I understand the point you're making about SharedDataPointer not exactly being functionally similar to std::shared_ptr. I'm curious as to why the code still exists as it does today though for a few reasons:

It appears that the SharedDataPointer class was used as a replacement for std::shared_ptr when it wasn't available in the c++ standard. As an example, SharedLibrary uses the SharedDataPointer to wrap the SharedLibraryPrivate object. In today's world, this usage of the copy-on-write idiom for this class is not necessary. It is my understanding that it was required when CppMicroServices didn't have the bundle layer implemented, but at this point it has been developed enough that the raw use of SharedLibrary is unnecessary. Do you happen to know of any use cases where a user uses the exported header for SharedLibrary? Also, why is the copy-on-write idiom so important for the replacement of SharedDataPointer to have? In my view, especially in the case of SharedLibrary, there is no need for copying on write at all; it actually appears to be more inefficient. If there are multiple callers have a reference to a SharedLibrary, calling setX() will modify the underlying object that all the std::shared_ptr objects are referring to; a copy-propagation operation to the rest of the callers holding a reference is not necessary.

It's because of these primary reasons I believe that a switch to std::shared_ptr would be okay.

As a side note, I believe that there should be a design change for SharedLibrary to make the code more modern, maintainable, and extendable. The following is a list of design changes that I propose:

1) Stop exporting SharedLibrary.h and keeping the class internal only.
2) Use std::shared_ptr instead of SharedDataPointer.
2) Remove "setter" and some getter functions from the API. Since the class would be used internal only, there appears to be no need to have numerous setter functions to change the path/prefixes/suffixes of the shared library. This can be pre-computed and passed into the SharedLibrary constructor to load the library and overall removes the amount of technical debt incurred by having to maintain these unused functions.

Let me know what you think. I'd love to go over this with you to arrive at a holistically better solution.

achristoforides:
Thanks for your response.

I understand the point you're making about SharedDataPointer not exactly being functionally similar to std::shared_ptr. I'm curious as to why the code still exists as it does today though for a few reasons:

It appears that the SharedDataPointer class was used as a replacement for std::shared_ptr when it wasn't available in the c++ standard. As an example, SharedLibrary uses the SharedDataPointer to wrap the SharedLibraryPrivate object. In today's world, this usage of the copy-on-write idiom for this class is not necessary. It is my understanding that it was required when CppMicroServices didn't have the bundle layer implemented, but at this point it has been developed enough that the raw use of SharedLibrary is unnecessary. Do you happen to know of any use cases where a user uses the exported header for SharedLibrary? Also, why is the copy-on-write idiom so important for the replacement of SharedDataPointer to have? In my view, especially in the case of SharedLibrary, there is no need for copying on write at all; it actually appears to be more inefficient. If there are multiple callers have a reference to a SharedLibrary, calling setX() will modify the underlying object that all the std::shared_ptr objects are referring to; a copy-propagation operation to the rest of the callers holding a reference is not necessary.

It's because of these primary reasons I believe that a switch to std::shared_ptr would be okay.

As a side note, I believe that there should be a design change for SharedLibrary to make the code more modern, maintainable, and extendable. The following is a list of design changes that I propose:

1) Stop exporting SharedLibrary.h and keeping the class internal only.
2) Use std::shared_ptr instead of SharedDataPointer.
2) Remove "setter" and some getter functions from the API. Since the class would be used internal only, there appears to be no need to have numerous setter functions to change the path/prefixes/suffixes of the shared library. This can be pre-computed and passed into the SharedLibrary constructor to load the library and overall removes the amount of technical debt incurred by having to maintain these unused functions.

Let me know what you think. I'd love to go over this with you to arrive at a holistically better solution.

httpservice/src/HttpOutputStreamBuffer.h
```diff
@@ -54,7 +55,7 @@ class HttpOutputStreamBuffer : public std::streambuf
 
 private:
   std::vector<char> m_Buffer;
-  HttpServletResponsePrivate* m_Response;
+  std::shared_ptr<HttpServletResponsePrivate> m_Response;
```
saschazelzer: I think this will result in a smart-pointer life-cycle, leaking the HttpServletResponsePrivate object.

framework/include/cppmicroservices/SharedLibrary.h
```diff
@@ -209,7 +209,7 @@ class US_Framework_EXPORT SharedLibrary
   bool IsLoaded() const;
 
 private:
-  ExplicitlySharedDataPointer<SharedLibraryPrivate> d;
+  std::shared_ptr<SharedLibraryPrivate> d;
```
saschazelzer: I think we should make this more straight-forward and use a std::unique_ptr here. Then always deep-copy the private impl in the SharedLibrary copy constructor.

saschazelzer:
@achristoforides Thanks for taking this on. `SharedDataPointer` was not a replacement for the back-then-not-yet-usable C++11 `std::shared_ptr` (which does not have copy-on-write semantics). It's raison d'être is to conveniently create types which expose value semantics but internally share data if possible (until a client calls a non-const member function, which then implicitly clones the internal representation if there is more than one user, before making the modification).

However, as of now, the only class with non-const members (not counting the assignment operator) is `SharedLibrary` which I am totally fine with re-doing. This should not affect the semantics of that class and hence no test changes should be necessary though.

Regarding your other points about `SharedLibrary`, there is [this RFC](https://github.com/CppMicroServices/rfcs/pull/7) which converges towards using `SharedLibrary` in the DS bundle in order to re-use the library loading logic.

achristoforides:
@saschazelzer Regarding SharedDataPointer, I guess I'm not understanding exactly how it is different than using a std:;shared_ptr. 

As a concrete example, we can look at Bundle.h and LDAPFilter.h. Inside Bundle.h, we see a private member which is is the private implementation of Bundle but stored as a std::shared_ptr and in LDAPFilter.h there is a SharedDataPointer to the private implementation of LDAPFilter. How are these fundamentally different? Why use std::shared_ptr in Bundle and not in LDAPFilter? Are there some sort of design decisions/hidden assumptions being made?

And to your point about creating types which expose value semantics, I believe that using std::shared_ptr still accomplishes this. In the case of LDAPFilter, a user still creates it the normal way and interacts with the object using the 'dot' operator. The only aspect which fundamentally changes is the way in which we, as implementers, implement the functionality of the public and private member functions. Instead of internally using value semantics on the pointer to the private data, we use pointer semantics; clients are never affected/see these changes and I believe it adds a layer of unnecessary indirection and meaning to the data being represented.

Am I missing something?

saschazelzer:
@achristoforides `Bundle` was designed to be a reference type. Modifying a `Bundle` instance affects all instances which represent the "same" bundle. Contrary, the `LDAPFilter` type is designed to be a value type. If it would have a non-const member function (a setter), it would access the internal implementation using `d->` or `*d` by `T& operator*()` or `T* operator->()` (note the missing `const`). Look at the implementation of these operators in `SharedDataPointer` and `std::shared_ptr`. The first clones itself if the non-const variant is used and its reference count is > 1, while the latter does no such thing, it just always returns the same raw pointer shared with all instances created via copy constructor or assignment operator (`ExplicitlySharedDataPointer` behaves the same as `std::shared_ptr`). So modifying a `LDAPFilter` instance is local (if it would have non-const member functions) while modifying a `Bundle` instance is not and the type needs to e.g. provide thread-safety guarantees if desired.

Because all of our types which use `SharedDataPointer` except for `SharedLibrary` are essentially read-only, the difference is not visible to a client and we can just use `std::shared_ptr` as well.

achristoforides:
@saschazelzer Ahhhhh I think I understand now. So from what I'm reading, you're saying we should continue to use SharedDataPointer in SharedLibrary due to the desired functionality but modifications to all other classes which use SharedDataPointer can be safely converted to using std::shared_ptr right?

saschazelzer:
Almost. I am arguing the after inspecting the public API of our classes which make use of `SharedDataPointer`, all except `SharedLibrary` can be changed to use `std::shared_ptr` instead because they only expose const methods to users of those classes cannot observe any difference. Should we ever need to introduce non-const methods, the internals will have to change but it looks like this is unlikely.

For `SharedLibrary`, I remarked in my code review to use `std::unique_ptr` instead.

In conclusion, after making those changes, the `SharedDataPointer` header file can be removed.

achristoforides:
@saschazelzer Sounds good to me. I appreciate the explanation!

httpservice/src/HttpOutputStreamBuffer.h
achristoforides: I'm almost certain that this won't occur.

achristoforides:
@saschazelzer I believe I've done what we discussed. I encourage you and others to take a look as well! @jeffdiclemente

framework/src/util/SharedLibrary.cpp
```diff
@@ -69,7 +68,14 @@ SharedLibrary::SharedLibrary()
   : d(new SharedLibraryPrivate)
 {}
 
-SharedLibrary::SharedLibrary(const SharedLibrary&) = default;
+SharedLibrary::SharedLibrary(const SharedLibrary& other) : d(new SharedLibraryPrivate) {
```
saschazelzer: Just use the copy constructor of `SharedLibraryPrivate` instead of doing it manually below.

If the copied raw `m_Handle` pointer is not null, set it to `nullptr` and call `Load()`. Same in the assignment operator. This will ensure `Unload` can be called on a copied `SharedLibrary` instance without messing up the internal library reference counting.

framework/src/util/SharedLibrary.cpp
```diff
@@ -88,35 +94,51 @@ SharedLibrary::SharedLibrary(const std::string& absoluteFilePath)
 
 SharedLibrary::~SharedLibrary() = default;
 
-SharedLibrary& SharedLibrary::operator=(const SharedLibrary&) = default;
+SharedLibrary& SharedLibrary::operator=(const SharedLibrary& other)
+{
+  SharedLibraryPrivate* slp = new SharedLibraryPrivate;
```
saschazelzer: `d.reset(new SharedLibraryPrivate(*other.d))` should be enough, plus the `m_Handle` check and library loading as in the copy constructor.

framework/src/util/SharedLibrary.cpp
```diff
@@ -134,36 +156,43 @@ void SharedLibrary::Load()
...
 }
 
 void SharedLibrary::SetName(const std::string& name)
 {
   if (IsLoaded() || !d->m_FilePath.empty())
     return;
-  d.Detach();
-  d->m_Name = name;
+
+  SharedLibraryPrivate p = *d;
```
saschazelzer: `d` is not shared with anyone, why can't we just do `d->m_Name = name` but need to create another copy here? In fact, `make_unique` invokes the copy constructor again, effectively creating a second copy.

Same for the other changes below.

httpservice/src/HttpOutputStreamBuffer.h
saschazelzer: I am afraid "almost" is not good enough :)

The `HttpServletResponsePrivate` instance can now hold on to a shared_ptr (instead of a raw pointer) to itself via the `m_HttpOutputStreamBuf` member. That is a classic smart pointer cycle and hence a memory leak in my book.

saschazelzer:
@jeffdiclemente I agree that we could discuss the design of `SharedLibrary` in general. For this PR, I think we should move forward with removing `SharedDataPointer` but not change its semantics. Both to keep this PR focused and also to not make backwards-incompatible changes. `SharedLibrary` is released public API after all.

framework/src/util/SharedLibrary.cpp
```diff
@@ -47,16 +47,15 @@
...
 public:
   SharedLibraryPrivate()
-    : 
-     m_Suffix(US_LIB_EXT)
+    : m_Suffix(US_LIB_EXT)
     , m_Prefix(US_LIB_PREFIX)
   {}
 
-  void* m_Handle{nullptr};
+  std::shared_ptr<void *> m_Handle{ nullptr };
```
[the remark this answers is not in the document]
achristoforides: If this is a raw pointer, then calling load would have no affect on another object if it is aliased with the one which is now opened (necessitating a list of all objects which are aliased with the current one). It's like this because an argument was made for using std::unique_ptr.

httpservice/src/HttpOutputStreamBuffer.h
achristoforides: Ahhh I missed this. I can convert it back.

framework/src/util/SharedLibrary.cpp
achristoforides: I believe that the handle should remain a std::shared_ptr. It removes the necessity to call Load() on the newly constructed object if creating another SharedLibrary object through the copy constructor. Not to mention, calling Load() on the same dll twice doesn't appear to be a good idea. If you would really like me to switch the handle back to a raw pointer and unnecessarily call Load() again, I can, but I'd prefer not to for the reasons above.

framework/src/util/SharedLibrary.cpp
achristoforides: For the reason above about keeping the handle as a std::shared_ptr, I'm not implementing the opening part of your comment unless you decide that you truly want it to be a raw pointer.

httpservice/src/HttpOutputStreamBuffer.h
achristoforides: Ahhh I didn't see this. I will revert it back to using a raw pointer.

framework/src/util/SharedLibrary.cpp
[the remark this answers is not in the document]
achristoforides: I agree that the overall design should be discussed, but by making the pointer to the SharedLibraryPrivate objects a unique pointer, I firmly believe that keeping the handle as a std::shared_ptr is okay. More specifically, I think it is alright since it doesn't necessitate the extra call to Load() like your suggestion does. If another SharedLibrary object is created through the copy constructor, all relevant data is copied over and the handles are shared; there is no need to call Load() once in the first object and then again in the second during construction. Additionally, since it is a shared pointer, opening/closing the library from one of the objects will affect the other.

framework/src/util/SharedLibrary.cpp
achristoforides: Yep, but due to my comment above to your other post, unless you absolutely want to keep the handle as a raw pointer, then I won't be doing the extra Load() call in this PR.

framework/src/util/SharedLibrary.cpp
achristoforides: Yes, but due to my comment above to your other post, unless you absolutely want to keep the handle as a raw pointer, then I won't be doing the extra Load() call in this PR.

jeffdiclemente:
> @jeffdiclemente I agree that we could discuss the design of `SharedLibrary` in general. For this PR, I think we should move forward with removing `SharedDataPointer` but not change its semantics. Both to keep this PR focused and also to not make backwards-incompatible changes. `SharedLibrary` is released public API after all.

@saschazelzer @achristoforides 
I agree. I would prefer to make as few changes as possible to ```SharedLibrary``` in this PR.

framework/include/cppmicroservices/SharedLibrary.h
```diff
@@ -209,7 +209,7 @@ class US_Framework_EXPORT SharedLibrary
   bool IsLoaded() const;
 
 private:
-  ExplicitlySharedDataPointer<SharedLibraryPrivate> d;
+  std::unique_ptr<SharedLibraryPrivate> d;
```
jeffdiclemente: Based on @saschazelzer comment that ```ExplicitySharedDataPointer``` is equivalent to ```std::shared_ptr``` and that ```SharedLibrary``` is copy-able, I think this should be ```std::shared_ptr``` and not ```unique_ptr```. Creating a new ```unique_ptr``` from the pointer to an existing memory object already owned by a ```unique_ptr``` in the ```SharedLibrary``` copy ctor is not correct to me. This seems to violate the single ownership model of ```unique_ptr```.

framework/src/util/SharedLibrary.cpp
```diff
@@ -88,7 +88,12 @@ SharedLibrary::SharedLibrary(const std::string& absoluteFilePath)
 
 SharedLibrary::~SharedLibrary() = default;
 
-SharedLibrary& SharedLibrary::operator=(const SharedLibrary&) = default;
+SharedLibrary& SharedLibrary::operator=(const SharedLibrary& other)
+{
+  d = other.d;
+
+  return *this;
+}
```
jeffdiclemente: the copy ctor and assignment operator can be ```= default```

Issue #421, opened 2019-12-16
ServiceComponent include is not exported for the INSTALL target


achristoforides:
When installing CppMicroServices, the ServiceComponent headers are not present.

PR #422, opened 2019-12-17
Install DS headers

jeffdiclemente:


jeffdiclemente:
This will need to be merged into master as well since it stops users from integrating with Declarative Services.

achristoforides:
I believe this closes #421 right?

jeffdiclemente:
Yes, this will fix #421

PR #424, opened 2019-12-17
Migrate BundleActivatorTest to gtest

achristoforides:
This PR migrates BundleActivatorTest from the test driver to gtest.

### DIFF
--- a/framework/test/gtest/BundleActivatorTest.cpp
+++ b/framework/test/gtest/BundleActivatorTest.cpp
@@ -27,39 +27,67 @@
 #include "cppmicroservices/FrameworkEvent.h"
 #include "cppmicroservices/FrameworkFactory.h"
 
+#include "TestUtilBundleListener.h"
 #include "TestUtils.h"
-#include "TestingConfig.h"
-#include "TestingMacros.h"
+
+#include "gtest/gtest.h"
 
 using namespace cppmicroservices;
 
-// bundle.activator property not specified in the manifest and bundle has no activator class
-void TestNoPropertyNoClass(BundleContext& context)
+class BundleActivatorTest : public ::testing::Test
 {
-  auto bundle = testing::InstallLib(context, "TestBundleR");
-  US_TEST_CONDITION_REQUIRED(bundle, "Test for existing bundle TestBundleR");
-  US_TEST_NO_EXCEPTION(bundle.Start());
+protected:
+  static void SetUpTestSuite()
+  {
+    f = std::make_shared<cppmicroservices::Framework>(
+      FrameworkFactory().NewFramework());
+    f->Start();
+
+    EXPECT_NO_THROW(ctx = std::make_shared<cppmicroservices::BundleContext>(
+                      f->GetBundleContext()));
+  }
+
+  static void TearDownTestSuite()
+  {
+    f->Stop();
+    f->WaitForStop(std::chrono::milliseconds::zero());
+    f = nullptr;
+    ctx = nullptr;
+  }
+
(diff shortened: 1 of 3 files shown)


framework/test/gtest/BundleActivatorTest.cpp
```diff
@@ -27,39 +27,67 @@
...
+#include "TestUtilBundleListener.h"
 #include "TestUtils.h"
-#include "TestingConfig.h"
-#include "TestingMacros.h"
+
+#include "gtest/gtest.h"
 
 using namespace cppmicroservices;
 
+class BundleActivatorTest : public ::testing::Test
```
jeffdiclemente: Take a look at https://github.com/google/googletest/blob/master/googletest/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests-same-data-multiple-tests for guidelines on how to define the test fixture.

framework/test/gtest/BundleActivatorTest.cpp
```diff
@@ -27,39 +27,67 @@
...
+  static void TearDownTestSuite()
+  {
+    f->Stop();
+    f->WaitForStop(std::chrono::milliseconds::zero());
+    f = nullptr;
+    ctx = nullptr;
+  }
+
+  static std::shared_ptr<cppmicroservices::Framework> f;
+  static std::shared_ptr<cppmicroservices::BundleContext> ctx;
```
jeffdiclemente: I don't believe these should be static. Its undesired as it re-uses the state of the Framework across multiple tests. I know this is how the old tests were structured however I think that is fundamentally incorrect. Tests shouldn't rely on side-effects from other tests, unless that's the explicit point of the test.

framework/test/gtest/BundleActivatorTest.cpp
```diff
@@ -27,39 +27,67 @@
...
+  {
+    f = std::make_shared<cppmicroservices::Framework>(
+      FrameworkFactory().NewFramework());
+    f->Start();
+
+    EXPECT_NO_THROW(ctx = std::make_shared<cppmicroservices::BundleContext>(
+                      f->GetBundleContext()));
+  }
+
+  static void TearDownTestSuite()
```
jeffdiclemente: these methods shouldn't be ```static```

framework/test/gtest/BundleActivatorTest.cpp
```diff
@@ -27,39 +27,67 @@
...
-#include "TestingMacros.h"
+
+#include "gtest/gtest.h"
 
 using namespace cppmicroservices;
 
+class BundleActivatorTest : public ::testing::Test
+{
+protected:
+  static void SetUpTestSuite()
```
jeffdiclemente: doesn't this method have to be named ```SetUp``` to work?

framework/test/gtest/BundleActivatorTest.cpp
jeffdiclemente: doesn't this method have to be named ```TearDown``` to work?

framework/test/gtest/BundleActivatorTest.cpp
```diff
@@ -27,39 +27,62 @@
...
+  }
+
+  void TearDown() override
+  {
+    f->Stop();
+    f->WaitForStop(std::chrono::milliseconds::zero());
+  }
+
+  std::shared_ptr<cppmicroservices::Framework> f;
+  std::shared_ptr<cppmicroservices::BundleContext> ctx;
```
jeffdiclemente: using ```std::shared_ptr``` isn't necessary.

PR #436, opened 2020-01-21
Support resource linking from the usMacroCreateBundle macro

jeffdiclemente:
Change the Framework and DS to link resources by default, if supported.

### DIFF
--- a/cmake/usMacroCreateBundle.cmake
+++ b/cmake/usMacroCreateBundle.cmake
@@ -5,8 +5,8 @@ macro(usMacroCreateBundle _project_name)
 project(${_project_name})
 
 cmake_parse_arguments(${PROJECT_NAME}
-  "SKIP_EXAMPLES;SKIP_INIT;LINK_RESOURCES;APPEND_RESOURCES"
-  "VERSION;TARGET;SYMBOLIC_NAME"
+  "SKIP_EXAMPLES;SKIP_INIT"
+  "VERSION;TARGET;SYMBOLIC_NAME;EMBED_RESOURCE_METHOD"
   "DEPENDS;PRIVATE_INCLUDE_DIRS;LINK_LIBRARIES;SOURCES;PRIVATE_HEADERS;PUBLIC_HEADERS;RESOURCES;BINARY_RESOURCES"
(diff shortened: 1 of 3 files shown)


cmake/usMacroCreateBundle.cmake
```diff
@@ -43,6 +43,22 @@ if(${PROJECT_NAME}_DEPENDS)
   )
 endif()
 
+# Set the resource embedding method
+set(_resource_embed_type )
+if(${PROJECT_NAME}_LINK_RESOURCES AND ${PROJECT_NAME}_APPEND_RESOURCES)
```
carneyweb: Since there is only two possible states, should there be just one flag... something like "${PROJECT_NAME}_EMBED_TYPE" with values of either "LINK" or "APPEND". Then no error checking for and invalid state of two flags with 4 possible values (when only 2 are valid) needs to be done.

jeffdiclemente: @carneyweb This seems reasonable. If the value of the "embed resource type" is invalid it will be ignored by the macros which this one calls and the default resource embedding method will be used.

cmake/usMacroCreateBundle.cmake
```diff
@@ -43,6 +43,22 @@ if(${PROJECT_NAME}_DEPENDS)
...
+  message(SEND_ERROR "Both APPEND and LINK options were specified. These options are mutually exclusive.")
+endif()
+
+if(${PROJECT_NAME}_LINK_RESOURCES)
+  set(_resource_embed_type LINK)
+  if(NOT ${US_RESOURCE_LINKING_AVAILABLE} OR ${US_RESOURCE_LINKING_AVAILABLE} EQUAL 0)
+    message(WARNING "Resource linking was specified however its not supported on the system. Defaulting to ${US_DEFAULT_RESOURCE_MODE}.")
+    set(_resource_embed_type ${US_DEFAULT_RESOURCE_MODE})
+  endif()
```
jeffdiclemente: this can be removed. It is a redundant check which is already performed in `usFunctionEmbedResources.cmake`

PR #446, opened 2020-02-12
Fix Xcode 11 build failure (TestBundleRA)

kevinleeMW:
TestBundleRA fail to build in Xcode 11 with the following message:
`"main executable failed strict validation"`

This is the behavior described in the following Apple documentation regarding codesign:
https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG309

Appending a resources zip at the end of a dylib is not a supported workflow; and starting with Xcode 11, codesign step fails. 

I added an Apple-only CMake property to disable codesigning TestBundleRA.

### DIFF
--- a/framework/test/bundles/libRWithAppendedResources/CMakeLists.txt
+++ b/framework/test/bundles/libRWithAppendedResources/CMakeLists.txt
@@ -12,6 +12,3 @@ usFunctionCreateTestBundleWithResources(TestBundleRA
   APPEND_RESOURCES
 )
 
-if(APPLE)
-	set_target_properties(TestBundleRA PROPERTIES XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "")
-endif()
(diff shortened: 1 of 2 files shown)


framework/test/bundles/libRWithAppendedResources/CMakeLists.txt
```diff
@@ -12,3 +12,6 @@ usFunctionCreateTestBundleWithResources(TestBundleRA
   APPEND_RESOURCES
 )
 
+if(APPLE)
+	set_target_properties(TestBundleRA PROPERTIES XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "")
+endif()
```
jeffdiclemente: Is it possible to generalize this fix such that any bundle which is built on macOS using the resource compiler's "append" mode will have this xcode attribute set?

kevinleeMW: Added logic to "usFunctionEmbedResources.cmake" file instead, to generically disable code-signing for any bundle appending resources on macOS.

PR #450, opened 2020-02-17
Bundle::GetSymbol() API and Unittests

gitabhishekdp:
Initial Commit


framework/test/driver/BundleGetSymbolTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+    
+    SharedLibrary sh(b.GetLocation());
+    sh.Load();
+    
+    // Test for invalid inputs on valid bundle
+ 
+    US_TEST_FOR_EXCEPTION(std::invalid_argument , b.GetSymbol(sh.GetHandle(),""));
+    US_TEST_FOR_EXCEPTION(std::invalid_argument , b.GetSymbol(nullptr,"_us_create_activator_TestBundleA"));
+    
+    b.Stop();
```
mw-akumar: Is this missing a b.Start() ?

framework/test/driver/BundleGetSymbolTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+    
+    Bundle b;
+    US_TEST_FOR_EXCEPTION(std::invalid_argument , b.GetSymbol(nullptr,""));
+}
+
+void TestGetSymbolInvalidInput()
+{
+    FrameworkFactory factory;
+    auto f = factory.NewFramework();
+    f.Start();
```
mw-akumar: Should RAII be used here to avoid the Start/Stop pattern?

framework/test/driver/BundleGetSymbolTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void TestGetSymbolInvalidBundleInput()
+{
+    // Test for invalid bundle
+    
+    Bundle b;
+    US_TEST_FOR_EXCEPTION(std::invalid_argument , b.GetSymbol(nullptr,""));
+}
+
+void TestGetSymbolInvalidInput()
```
mw-akumar: The names should perhaps reflect the conditions that are being violated. 

e.g invalid symbol name, null shared library.

In the null shared library case, you won't need to even load the Shared Library

framework/test/driver/BundleGetSymbolTest.cpp
gitabhishekdp: Every test has a different requirement. e.g. Invalid Bundle, Valid bundle which is not started/installed and valid bundle with lifetime. Give this, I think SetUp and TearDown might not work we expect do things that are superfluous for some tests. I am keeping the tests as it is for now. I will keep in mind to add SetUp and TearDown henceforth.

framework/src/bundle/Bundle.cpp
```diff
@@ -316,6 +316,21 @@ Bundle::TimeStamp Bundle::GetLastModified() const
...
 }
 
+void* Bundle::GetSymbol(void * handle, const std::string& symname) const
+{
+  if(!d || !handle || symname.empty()) {
+      throw std::invalid_argument("invalid bundle");
+  }
+
+  if(GetState() != STATE_ACTIVE)
+  {
```
jeffdiclemente: make sure to run clang-format on this code (opening brace should be on same line as ```if``` statement).

Also, I suggest putting the constant on the left side of the expression. This ensures that the compiler can find subtle errors like using `=` instead of `==`.

framework/src/bundle/Bundle.cpp
```diff
@@ -316,6 +316,21 @@ Bundle::TimeStamp Bundle::GetLastModified() const
   return d->barchive ? d->barchive->GetLastModified() : d->timeStamp;
 }
 
+void* Bundle::GetSymbol(void * handle, const std::string& symname) const
+{
+  if(!d || !handle || symname.empty()) {
+      throw std::invalid_argument("invalid bundle");
```
jeffdiclemente: this exception can be thrown if more than the bundle is invalid. I'd suggest making the exception message a little more descriptive (to aid debugging).

framework/test/driver/BundleGetSymbolTest.cpp
```diff
@@ -0,0 +1,141 @@
...
+    auto frameworkCtx = f.GetBundleContext();
+    
+    auto b = testing::InstallLib(frameworkCtx, "TestBundleA");
+    b.Start();
+
+    // Test for invalid symbol handle on valid bundle
+ 
+    US_TEST_FOR_EXCEPTION(std::invalid_argument , b.GetSymbol(nullptr,"_us_create_activator_TestBundleA"));
+    
+    b.Stop();
```
jeffdiclemente: just a note: stopping the framework automatically stops all bundles.

framework/include/cppmicroservices/Bundle.h
```diff
@@ -376,6 +376,26 @@ class US_Framework_EXPORT Bundle
    */
   std::string GetLocation() const;
 
+  /**
+    * Retrieves the resolved symbol from bundle shared library and returns a function pointer associated with it
+    * 
+    * @param1 Handle to the Bundle's shared library
+    * @param2 Name of the symbol
```
jeffdiclemente: for consistency with the other doxygen comments I would suggest using the ```@param <name> <description>``` form instead of ```@param1```, ```@param2```, etc...

framework/test/gtest/BundleGetSymbolTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+// Test for invalid symbol name on valid bundle
+TEST_F(BundleGetSymbolTest, TestGetSymbolValidBundleWithInvalidSymbolHandle)
+{
+    SharedLibrary sh(bd.GetLocation());
+    sh.Load();
+ 
+    EXPECT_THROW(bd.GetSymbol(sh.GetHandle(),""),std::invalid_argument);
+}
+
+// Test for bundle which is not installed but with valid inputs
```
jeffdiclemente: ```cppmicroservices::testing::InstallLib``` installs but does not start the bundle. This test isn't testing a bundle which is not installed, its testing a bundle which is not started.

gitabhishekdp:
> I believe there is a missing test; getting a valid symbol from a bundle that has been started (i.e. by calling `Bundle::Start() `) without using `SharedLibrary` to load the bundle's shared library.

@jeffdiclemente  To get the handle to the bundle's shared library (first param to the new API) , I might have to replicate the logic in BundleUtils.cpp. I dont see any other way by which I can achieve this test point.

gitabhishekdp:
> > I believe there is a missing test; getting a valid symbol from a bundle that has been started (i.e. by calling `Bundle::Start() `) without using `SharedLibrary` to load the bundle's shared library.
> 
> @jeffdiclemente To get the handle to the bundle's shared library (first param to the new API) , I might have to replicate the logic in BundleUtils.cpp. I dont see any other way by which I can achieve this test point.

@jeffdiclemente Please let me know if we are ok with getting the handle to bundle's shared library using platform APIs in the test.

jeffdiclemente:
> > > I believe there is a missing test; getting a valid symbol from a bundle that has been started (i.e. by calling `Bundle::Start() `) without using `SharedLibrary` to load the bundle's shared library.
> > 
> > 
> > @jeffdiclemente To get the handle to the bundle's shared library (first param to the new API) , I might have to replicate the logic in BundleUtils.cpp. I dont see any other way by which I can achieve this test point.
> 
> @jeffdiclemente Please let me know if we are ok with getting the handle to bundle's shared library using platform APIs in the test.

yes that's fine with me.

framework/test/gtest/BundleGetSymbolTest.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    sh.Unload();
+}
+
+// Test for valid bundle and valid input without using SharedLibrary interface
+TEST_F(BundleGetSymbolTest, TestGetSymbolValidInputWithoutUsingSharedLibrary)
+{
+    void *pFuncAct = testHelper::dlopen(bd.GetLocation().c_str(),RTLD_LAZY);
+    EXPECT_TRUE(pFuncAct != nullptr) << "Error : null handle returned for bundle shared library !\n";
+    EXPECT_TRUE(bd.GetSymbol(pFuncAct,"_us_create_activator_TestBundleA") != nullptr) << "Error : Empty symbol returned from Bundle::GetSymbol !\n";
```
jeffdiclemente: Sorry, I meant that this test should not use ```SharedLibrary```, nor ```dlopen``` but instead call ```Bundle::Start``` (which will in turn call ```dlopen```) and then test the ```GetSymbol``` API.

gitabhishekdp: @jeffdiclemente Either I am not sure how I can fetch the handle to the bundle's shared library which is needed as a first parameter to new API or I am missing something that's glaringly obvious.

jeffdiclemente: @gitabhishekdp oops, my mistake. Since there is no access to the private classes in the unit tests this isn't feasible. Nevermind. Sorry for the confusion. you can delete this test and the testHelper code since it doesn't add any additional test coverage.

jeffdiclemente:
@gitabhishekdp FYI, there are build errors for the static build configuration. Most likely the new GetSymbol API and its tests don't account for when bundles are built as static libraries.

gitabhishekdp:
@jeffdiclemente Please let me know if there are any more action items left here.

PR #456, opened 2020-03-02
Use Xcode for Travis CI builds on macOS

kevinleeMW:
Switched from "Unix Makefiles" to "Xcode" for Travis CI builds on macOS.
This allows us to catch issues with the generated Xcode projects, which macOS users are more likely to use when debugging / building.


CMakeLists.txt
```diff
@@ -547,6 +547,13 @@ endif()
 
 usFunctionCheckResourceLinking()
 
+if(APPLE AND CMAKE_GENERATOR MATCHES "Xcode")
```
jeffdiclemente: a comment describing why this is necessary will help for maintenance.

PR #462, opened 2020-03-13
Add MSVC 2019 as a supported complier

achristoforides:
When initially compiling, there was a new warning which appeared regarding discarding a return value even though the function is marked as nodiscard. To fix this, US_UNUSED is used in the test.

Additionally, changes were made to the appveyor.yml file to include new tests for MSVC 2019


jeffdiclemente:
@achristoforides can you add Visual Studio 2019 to README.rst, in the table of supported compilers? Thanks.

PR #463, opened 2020-03-13
Fix BundleRegistry deadlock

### LINKED ISSUE #454: potential deadlock in BundleRegistry.cpp
A deadlock can occur if the same physical bundle file is installed multiple times using two different paths (e.g. /path/to/bundle/foo/bar/../../bundleA.dylib and /path/to/bundle/bundleA.dylib).

This happens because of recent changes in BundleRegistry.cpp where an exception coming from ```BundleRegistry::Install0``` is not caught, causing the ```initialBundleInstallMap``` to never be cleaned up. This leaves behind an entry in ```initialBundleInstallMap``` which causes subsequent calls to ```BundleRegistry::Install``` to believe that another thread is already installing the bundle (which its not) and wait indefinitely.

One way to fix this would be to use RAII to clean up ```initialBundleInstallMap```.
Another, less optimal way, would be to wrap calls to ```Install0``` in ```try/catch``` blocks.

To reproduce the issue, the following code can be copy/pasted into ```TestForInstallFailure()``` in ```framework\test\driver\BundleTest.cpp```:

```
  // Test that multiple calls to installing a bundle with the same symbolic name and
  // different paths produce the same result (i.e. an exception).
  const std::string canonicalBundleInstallPath(testing::LIB_PATH
      + util::DIR_SEP
      + US_LIB_PREFIX
      + "TestBundleA"
      + US_LIB_POSTFIX
      + US_LIB_EXT);
  US_TEST_NO_EXCEPTION(frameworkCtx.InstallBundles(canonicalBundleInstallPath));

  const std::string nonCanonicalBundleInstallPath(testing::LIB_PATH
      + util::DIR_SEP
      + "."
      + util::DIR_SEP
      + US_LIB_PREFIX
      + "TestBundleA"
      + US_LIB_POSTFIX
      + US_LIB_EXT);
  US_TEST_FOR_EXCEPTION(std::runtime_error, frameworkCtx.InstallBundles(nonCanonicalBundleInstallPath));
  US_TEST_FOR_EXCEPTION(std::runtime_error, frameworkCtx.InstallBundles(nonCanonicalBundleInstallPath));
```

kevinleeMW:
Fixes #454 

### DIFF
--- a/framework/src/bundle/BundleRegistry.cpp
+++ b/framework/src/bundle/BundleRegistry.cpp
@@ -43,6 +43,17 @@
 
 namespace cppmicroservices {
 
+// Helper class to ensure RAII for InitialBundleMap in case where Install0 throws
+class InitialBundleMapCleanup {
+public:
+  InitialBundleMapCleanup(std::function<void()> cleanupFcn) : _cleanupFcn(cleanupFcn) {}
+  ~InitialBundleMapCleanup() {
+	_cleanupFcn();
+  }
+private:
+  std::function<void()> _cleanupFcn;
+};
+
 BundleRegistry::BundleRegistry(CoreBundleContext* coreCtx)
   : coreCtx(coreCtx)
 {}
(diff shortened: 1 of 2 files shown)


framework/src/bundle/BundleRegistry.cpp
```diff
@@ -43,6 +43,17 @@
 
 namespace cppmicroservices {
 
+// Helper class to ensure RAII for InitialBundleMap in case where Install0 throws
+class InitialBundleMapCleanup {
+public:
+  InitialBundleMapCleanup(std::function<void()> cleanupFcn) : _cleanupFcn(cleanupFcn) {}
```
jeffdiclemente: use ```std::move(cleanupFcn)``` in the constructor's member initializer list since the lambdas you pass in aren't called by anyone else.

kevinleeMW: Added std::move per Jeff's suggestion

PR #466, opened 2020-04-06
SharedLibraryException handling

kevinleeMW:
SharedLibraryException handling per https://github.com/CppMicroServices/rfcs/pull/9


compendium/DeclarativeServices/test/SharedLibraryExceptionTest.cpp
```diff
@@ -0,0 +1,111 @@
...
+  auto context = GetFramework().GetBundleContext();
+  test::InstallLib(context, "TestBundleDSSLE2");
+
+  auto bundles = context.GetBundles();
+  auto bundle = std::find_if( bundles.begin(), bundles.end(), [](const cppmicroservices::Bundle& bundle) {
+    return (bundle.GetSymbolicName() == "TestBundleDSSLE2");
+  });
+
+  ASSERT_NE(bundle, bundles.end()) << "TestBundleDSSLE2 not found";
+  bundle->Start();
```
jeffdiclemente, line 104: In this case `Bundle::Start()` should never throw. I suggest using a ```ASSERT_NO_THROW```

framework/include/cppmicroservices/SharedLibraryException.h
```diff
@@ -0,0 +1,59 @@
...
+// ignore warning c4275 per MS documentation
+// https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4275
+#ifdef _MSC_VER
+#  pragma warning(push)
+#  pragma warning(disable : 4275)
+#endif
+
+namespace cppmicroservices {
+
+class US_Framework_EXPORT SharedLibraryException : public std::system_error
```
jeffdiclemente, line 40: mark `SharedLibraryException` as `final` so no other classes can be derived from it.

compendium/test_bundles/TestBundleDSSLE1/resources/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+    "bundle.symbolic_name" : "TestBundleDSSLE1",
+    "scr" : {
+        "version" : 1,
+        "components" : [{
+            "enabled": true,
+            "immediate" : true,
+            "implementation-class": "sample::ServiceComponentDSSLE1",
+            "service" : {
+                "interfaces" : ["test::InterfaceSLE"]
```
jeffdiclemente, line 10: for both TestBundleDSSLE1 and TestBundleDSSLE2 you can probably use an existing service interface instead of creating a new one since the interface is not relevant to the test.

compendium/DeclarativeServices/test/SharedLibraryExceptionTest.cpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  cppmicroservices::Framework& GetFramework() { return framework; }
+  std::shared_ptr<scr::ServiceComponentRuntime> dsRuntimeService;
+  
+private:
+  cppmicroservices::Framework framework;
+};
+
+TEST_F(SharedLibraryExceptionTest, testDSBundleImmediateTrue)
```
jeffdiclemente, line 79: In addition to the fake bundles which simulate a shared library loading failure, is it possible to write tests which use mocks to test the behavior of internal DS classes when a `SharedLibraryException` is thrown?

compendium/DeclarativeServices/test/SharedLibraryExceptionTest.cpp
kevinleeMW: made the change accordingly

framework/include/cppmicroservices/SharedLibraryException.h
kevinleeMW: made the change accordingly

compendium/test_bundles/TestBundleDSSLE1/resources/manifest.json
kevinleeMW: now using Interface1

compendium/DeclarativeServices/test/SharedLibraryExceptionTest.cpp
kevinleeMW: Will be looking into using mocks

kevinleeMW: The function that throws due to dlopen/loadlibrary seemed a little difficult to mock.
After reviewing current changes, if the needs are still there, we can create a separate task to add more testpoints using mocks.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -520,6 +521,20 @@ std::exception_ptr BundlePrivate::Start0()
...
+        ex.SetBundle(thisBundle);
+      }
+      res = std::make_exception_ptr(ex);
+      
+      // Log SharedLibraryException
+      coreCtx->listeners.SendFrameworkEvent(
+        FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
+                       thisBundle,
+                       std::string(),
+                       res));
```
jeffdiclemente, line 537: sending a framework event isn't required.

compendium/DeclarativeServices/src/manager/states/CMEnabledState.cpp
```diff
@@ -98,7 +103,11 @@ void CMEnabledState::DeleteConfigurations()
   auto fut = GetFuture();
   if(fut.valid())
   {
-    fut.get(); // wait for the configurations to become available
+    try {
+      fut.get(); // wait for the configurations to become available
+    } catch (...) {
+      std::cout << "exception here" << std::endl;
```
jeffdiclemente, line 109: is this leftovers from debugging?

in general, don't write to `std::cout`, log it using the LogService

framework/src/util/SharedLibrary.cpp
```diff
@@ -99,25 +101,31 @@ void SharedLibrary::Load(int flags)
...
     std::string errMsg = "Loading ";
     errMsg.append(libPath)
       .append("failed with error: ")
       .append(util::GetLastWin32ErrorStr());
-
-    throw std::runtime_error(errMsg);
+    
+    // Bundle of origin information is not available here. It will be
+    // set later in BundlePrivate::Start0() in a catch statement.
+    throw cppmicroservices::SharedLibraryException(err_code, errMsg);
```
jeffdiclemente, line 128: catching the `std::runtime_error` in the caller and then re-throwing a `SharedLibraryException` can remove the need to have `SharedLibraryException::SetBundle`

framework/src/util/SharedLibraryException.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+}
+
+Bundle SharedLibraryException::GetBundle() const {
+  return origin;
+}
+
+void SharedLibraryException::SetBundle(const Bundle &bundle) {
+  origin = bundle;
+}
```
jeffdiclemente, line 40: not ideal. There shouldn't be a need to set the bundle after construction. It only complicates the design.

framework/include/cppmicroservices/SharedLibraryException.h
```diff
@@ -0,0 +1,57 @@
...
+#endif
+
+namespace cppmicroservices {
+
+class US_Framework_EXPORT SharedLibraryException final : public std::system_error
+{
+public:
+  explicit SharedLibraryException(const std::error_code ec,
+                                  const std::string& what,
+                                  const cppmicroservices::Bundle& origin = Bundle());
```
jeffdiclemente, line 45: pass by value to allow `std::move`

framework/src/bundle/BundlePrivate.cpp
kevinleeMW: removed sending framework event

compendium/DeclarativeServices/src/manager/states/CMEnabledState.cpp
kevinleeMW: yup, forgot to remove from debugging. removed the unnecessary try/catch

framework/src/util/SharedLibrary.cpp
kevinleeMW: I think it might be a better idea to throw a std::system_error, then rethrow as a SharedLibraryException. Actually that's how I had it implemented first, then thought about implications of this. This would mean that we assume all the code in BundlePrivate::Start0() would be expected to NOT throw a system_error except when calling SharedLibrary::Load(int flags).

I'll make the changes accordingly with std::system_error.

framework/src/util/SharedLibraryException.cpp
kevinleeMW: agreed; removed.

framework/include/cppmicroservices/SharedLibraryException.h
kevinleeMW: Does this mean when a SharedLibraryException is constructed, we must pass in a std::move(bundle)? Any reason why that's desirable over pass-by-reference?

jeffdiclemente: yes. 
we want to make it possible for the compiler to optimize for callers who use `std::move' so that extra memory doesn't have to used.

kevinleeMW: Made changes accordingly

compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
```diff
@@ -37,7 +38,12 @@ std::shared_ptr<ComponentInstance> CCActiveState::Activate(ComponentConfiguratio
   auto logger = mgr.GetLogger();
   if(latch.CountUp())
   {
-    instance = mgr.CreateAndActivateComponentInstance(clientBundle);
+    try {
```
carneyweb, line 41: consider using a scope guard here to CountDown().

framework/test/gtest/SharedLibraryExceptionTest.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "cppmicroservices/FrameworkEvent.h"
+#include "cppmicroservices/FrameworkFactory.h"
+#include "cppmicroservices/SharedLibraryException.h"
+
+#include "TestUtils.h"
+
+#include "gtest/gtest.h"
+
+#if defined (US_BUILD_SHARED_LIBS)
+TEST(SharedLibraryException, FrameworkSharedLibraryException)
```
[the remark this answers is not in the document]
kevinleeMW: Do you mean I should add some assert statements comparing values assigned to the exception (error code, string, and origin bundle) to the expected value? Or do you mean more testing than that?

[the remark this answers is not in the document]
jeffdiclemente: yep, the former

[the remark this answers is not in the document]
kevinleeMW: added testpoint

compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
kevinleeMW: implemented a scope guard class to invoke CountDown()

compendium/DeclarativeServices/test/SharedLibraryExceptionTest.cpp
```diff
@@ -0,0 +1,134 @@
+/*=============================================================================
```
jeffdiclemente, line 1: file name should follow the same naming convention as the other test files, `TestSharedLibraryException.cpp` instead of `SharedLibraryExceptionTest.cpp`

framework/test/gtest/SharedLibraryExceptionTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+TEST_F(SharedLibraryExceptionTest, testConstructSharedLibraryException)
+{
+  auto bundle = GetBundle();
+  std::error_code errco(0, std::system_category());
+  std::string exception_str("Test String");
+  
+  auto ex = cppmicroservices::SharedLibraryException(errco, exception_str, bundle);
+  
+  ASSERT_EQ(ex.code(), errco);
+  ASSERT_EQ(ex.what(), exception_str);
```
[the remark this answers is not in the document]
kevinleeMW: Looks like the behavior is platform dependent. Will be addressing this issue.

compendium/DeclarativeServices/src/manager/BundleLoader.cpp
```diff
@@ -73,21 +76,23 @@ GetComponentCreatorDeletors(const std::string& compName,
     std::wstring bundlePathWstr = UTF8StrToWStr(fromBundle.GetLocation());
     handle = reinterpret_cast<void*>(LoadLibraryW(bundlePathWstr.c_str()));
     if (handle == nullptr) {
+      std::error_code err_code(GetLastError(), std::generic_category());
       std::string errMsg("Unable to load bundle binary ");
       errMsg += fromBundle.GetLocation();
       errMsg += ". Error: ";
       errMsg += std::to_string(GetLastError());
-      throw std::runtime_error(errMsg);
+      throw cppmicroservices::SharedLibraryException(err_code, errMsg, std::move(fromBundle));
```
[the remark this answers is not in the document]
kevinleeMW: done

compendium/DeclarativeServices/src/manager/BundleLoader.cpp
```diff
@@ -73,21 +76,23 @@ GetComponentCreatorDeletors(const std::string& compName,
...
     handle = dlopen(fromBundle.GetLocation().c_str(), RTLD_LAZY | RTLD_LOCAL);
     if (handle == nullptr) {
+      std::error_code err_code(errno, std::generic_category());
       std::string errMsg("Unable to load bundle binary ");
       errMsg += fromBundle.GetLocation();
       errMsg += ". Error: ";
       const char* dlErrMsg = dlerror();
       errMsg += (dlErrMsg) ? dlErrMsg : "none";
-      throw std::runtime_error(errMsg);
+      throw cppmicroservices::SharedLibraryException(err_code, errMsg, std::move(fromBundle));
```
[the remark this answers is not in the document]
kevinleeMW: done

compendium/DeclarativeServices/test/SharedLibraryExceptionTest.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include "cppmicroservices/SharedLibraryException.h"
+
+#include "cppmicroservices/servicecomponent/runtime/ServiceComponentRuntime.hpp"
+
+#include "TestInterfaces/Interfaces.hpp"
+#include "TestUtils.hpp"
+
+namespace scr = cppmicroservices::service::component::runtime;
+
+class SharedLibraryExceptionTest : public ::testing::Test
```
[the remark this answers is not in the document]
kevinleeMW: Using an invalid bundle using Bundle() ctor didn't work, as BundleLoader tries to get the location of the bundle (and fails) before calling dlopen.

Using the framework bundle passes the location validity test, but fails to dlopen; so I'll add a testpoint using the framework bundle as the bundle input parameter.

framework/test/gtest/SharedLibraryExceptionTest.cpp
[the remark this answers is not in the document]
kevinleeMW: Changed the code to do prefix checking, using std::string::rfind

compendium/DeclarativeServices/test/SharedLibraryExceptionTest.cpp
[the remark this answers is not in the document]
kevinleeMW: Also added a testpoint to usFrameworkTests' SharedLibraryExceptionTest.cpp, to create a SharedLibrary instance with the framework, and calling SharedLibrary::Load() to make sure that it throws a std::system_error.

compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
```diff
@@ -23,10 +23,24 @@
...
 namespace scrimpl {
 
+class LatchScopeGuard {
+public:
+  LatchScopeGuard(std::function<void()> cleanupFcn) : _cleanupFcn(std::move(cleanupFcn))
+  {
+    
+  }
+  ~LatchScopeGuard() {
+    _cleanupFcn();
```
jeffdiclemente, line 38: Make sure this destructor cannot throw. `latch.CountDown()` can throw so this destructor can throw. I suggest a `try/catch` and log the exception in the `catch` if its safe to do so (i.e. if `Log` doesn't throw).

compendium/DeclarativeServices/test/SharedLibraryExceptionTest.cpp
kevinleeMW: changed as suggested

compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
kevinleeMW: will try/catch around latch.CountDown(). No mention of any throws from Log (both observing the code and looking at the OSGi spec).

kevinleeMW:
Thanks for your thorough reviews, Jeff and Mike!
If there aren't any additional feedback by tomorrow (Friday 4/24) afternoon, I'll merge the changes into development by EOD Friday!

PR #470, opened 2020-05-14
Convert DS BundleLoader to use core framework APIs

### LINKED ISSUE #469: Convert DS BundleLoader class to use Bundle::GetSymbol and SharedLibrary::Load APIs
Remove duplicate code by using Bundle::GetSymbol and SharedLibrary::Load APIs in DS BundleLoader class.

kevinleeMW:

closes #469

### DIFF
--- a/compendium/DeclarativeServices/src/manager/BundleLoader.cpp
+++ b/compendium/DeclarativeServices/src/manager/BundleLoader.cpp
@@ -21,6 +21,7 @@
   =============================================================================*/
 
 #include "cppmicroservices/Bundle.h"
+#include "cppmicroservices/Constants.h"
 #include "cppmicroservices/SharedLibrary.h"
 #include "cppmicroservices/SharedLibraryException.h"
 
@@ -76,7 +77,13 @@ GetComponentCreatorDeletors(const std::string& compName,
   } else {
     SharedLibrary sh(bundleLoc);
     try {
-      sh.Load();
+      auto ctx = fromBundle.GetBundleContext();
+      auto opts = ctx.GetProperty(Constants::LIBRARY_LOAD_OPTIONS);
+      if (!opts.Empty()) {
+        sh.Load(any_cast<int>(opts));
+      } else {
+        sh.Load();
+      }
     } catch (const std::system_error& ex) {
       // SharedLibrary::Load() will throw a std::system_error when a shared library
       // fails to load. Creating a SharedLibraryException here to throw with fromBundle information.
--- a/compendium/DeclarativeServices/src/manager/BundleLoader.cpp
+++ b/compendium/DeclarativeServices/src/manager/BundleLoader.cpp
@@ -77,6 +77,7 @@ GetComponentCreatorDeletors(const std::string& compName,
   } else {
     SharedLibrary sh(bundleLoc);
     try {
+#ifdef US_PLATFORM_POSIX
       auto ctx = fromBundle.GetBundleContext();
       auto opts = ctx.GetProperty(Constants::LIBRARY_LOAD_OPTIONS);
       if (!opts.Empty()) {
@@ -84,6 +85,9 @@ GetComponentCreatorDeletors(const std::string& compName,
       } else {
         sh.Load();
       }
+#else
+      sh.Load();
+#endif
     } catch (const std::system_error& ex) {
       // SharedLibrary::Load() will throw a std::system_error when a shared library
       // fails to load. Creating a SharedLibraryException here to throw with fromBundle information.
--- a/compendium/DeclarativeServices/src/manager/BundleLoader.cpp
+++ b/compendium/DeclarativeServices/src/manager/BundleLoader.cpp
@@ -77,17 +77,9 @@ GetComponentCreatorDeletors(const std::string& compName,
   } else {
     SharedLibrary sh(bundleLoc);
     try {
-#ifdef US_PLATFORM_POSIX
       auto ctx = fromBundle.GetBundleContext();
       auto opts = ctx.GetProperty(Constants::LIBRARY_LOAD_OPTIONS);
-      if (!opts.Empty()) {
-        sh.Load(any_cast<int>(opts));
-      } else {
-        sh.Load();
-      }
-#else
-      sh.Load();
-#endif
+      sh.Load(any_cast<int>(opts));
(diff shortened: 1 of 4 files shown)


compendium/DeclarativeServices/src/manager/BundleLoader.cpp
```diff
@@ -72,62 +74,39 @@ GetComponentCreatorDeletors(const std::string& compName,
...
-      std::string errMsg("Unable to load bundle binary ");
-      errMsg += fromBundle.GetLocation();
-      errMsg += ". Error: ";
-      const char* dlErrMsg = dlerror();
-      errMsg += (dlErrMsg) ? dlErrMsg : "none";
-      throw cppmicroservices::SharedLibraryException(
-        err_code, std::move(errMsg), std::move(fromBundle));
+    SharedLibrary sh(bundleLoc);
+    try {
+      sh.Load();
```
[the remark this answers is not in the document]
kevinleeMW: I'm having a hard time finding a way to retrieve a framework context in this code.
Is there a public API that would return the framework in this code?
(I only have access to a Bundle object here)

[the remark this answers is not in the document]
jeffdiclemente: `BundleContext::GetProperties()` can be used to retrieve the framework confguration properties. You can use any `Bundle Context` to get the framework properties.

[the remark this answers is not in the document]
kevinleeMW: Added checking for framework property for library load options for posix only

compendium/DeclarativeServices/src/manager/BundleLoader.cpp
```diff
@@ -72,62 +75,49 @@ GetComponentCreatorDeletors(const std::string& compName,
...
     std::regex_replace(compName, std::regex("::"), "_");
   const std::string newInstanceFuncName("NewInstance_" + symbolName);
   const std::string deleteInstanceFuncName("DeleteInstance_" + symbolName);
-#if defined(_WIN32)
-  void* sym = reinterpret_cast<void*>(GetProcAddress(
-    reinterpret_cast<HMODULE>(handle), newInstanceFuncName.c_str()));
-  void* delsym = reinterpret_cast<void*>(GetProcAddress(
-    reinterpret_cast<HMODULE>(handle), deleteInstanceFuncName.c_str()));
+  
+  void* sym = fromBundle.GetSymbol(handle, newInstanceFuncName);
```
achristoforides, line 105: Consider using a more descriptive file name (e.g., newsym or newInstSym)

compendium/DeclarativeServices/src/manager/BundleLoader.cpp
```diff
@@ -72,62 +75,49 @@ GetComponentCreatorDeletors(const std::string& compName,
...
-      std::string errMsg("Unable to load bundle binary ");
-      errMsg += fromBundle.GetLocation();
-      errMsg += ". Error: ";
-      errMsg += std::to_string(GetLastError());
-      throw cppmicroservices::SharedLibraryException(
-        err_code, std::move(errMsg), std::move(fromBundle));
-    }
+    SharedLibrary sh(bundleLoc);
+    try {
+#ifdef US_PLATFORM_POSIX
```
carneyweb, line 80: If we make a guarantee that ctx.GetProperty(Constants::LIBRARY_LOAD_OPTIONS) is always defined to its default values or possibly overwritten by customer settings, then this code collapses to this:

    try {
        auto ctx = fromBundle.GetBundleContext();
        auto opts = ctx.GetProperty(Constants::LIBRARY_LOAD_OPTIONS);
        sh.Load(opts);
    }...

compendium/DeclarativeServices/src/manager/BundleLoader.cpp
kevinleeMW: changed to newsym

compendium/DeclarativeServices/src/manager/BundleLoader.cpp
kevinleeMW: CoreBundleContext.cpp sets the libraryLoadOptions member variable in the following way:

In the constructor: 
`libraryLoadOptions(0)`

In Init():
```
#ifdef US_PLATFORM_POSIX
  try {
      libraryLoadOptions = any_cast<int>(frameworkProperties[Constants::LIBRARY_LOAD_OPTIONS]);
  } catch (...) {
      DIAG_LOG(*sink) << "Unable to read default library load options from config.";
      libraryLoadOptions = RTLD_LAZY | RTLD_LOCAL;
  }
  DIAG_LOG(*sink) << "Library Load Options = " << libraryLoadOptions;
#endif
```
It's guarding against an any_cast failure, in case the user defines a framework property value that is not an int. It does NOT guard against any integer value not defined by dlopen (i.e. -999).

Knowing this, I wanna move the try/catch block in Init() to InitProperties(...). This will parse the framework property, then if the value is an INT, it'll place the default value for POSIX. Later, Init() will assign the "correct(ed) value" to libraryLoadOptions member variable.

The question is: IF the user accidentally set framework property value to a non-integer value, is it okay for us to "correct" the value to a known working default? Or is it better to leave the framework property as the value assigned by user (although wrong), and do the checking when trying to obtain that property later (i.e. when assigning libraryLoadOptions member variable, and when trying to parse the value in BundleLoader class)?

@jeffdiclemente

jeffdiclemente: @kevinleeMW the problem (i.e. an incorrect library load option value) should be caught as early as possible and logged. In general catching and reporting the problem when it happens and not later, when its harder to reason about where the error came from, is preferable. So I suggest using a correct default value in the error case.

kevinleeMW: POSIX:
if user defines a non-integer load option -> override with default value RTLD_LAZY | RTLD_LOCAL
if user defines an integer load option -> use the user-defined value

Windows:
load option framework property is always overriden to be 0 (whether the user defines it or not)

kevinleeMW:
@jeffdiclemente @carneyweb @achristoforides 
I made changes based on all your feedback.

PR #474, opened 2020-06-02
Error if duplicate service component reference names are used

### LINKED ISSUE #473: [Declarative Services] duplicate service reference names do not cause an error
If a service component has more than one service reference with the same name, no error is produced. This can lead to unexpected behavior which is hard to diagnose at run-time.

The reference name must be unique among all other references in the service component (see [RFC](https://github.com/CppMicroServices/rfcs/blob/master/text/0003-declarative-services.md#service-component-description-json-specification)).

This type of error should be caught as early as possible. Ideally the SCRCodeGen tool should catch this and return an error which can fail a build.

test case:

```
{
    "scr" : { "version" : 1,
              "components": [{
                       "implementation-class": "DSSpellCheck::SpellCheckImpl",
                       "references": [{
                         "name": "foo",
                         "interface": "Bar"
                       },
                       {
                         "name": "foo",
                         "interface": "Bar"
                       }]
                       }]
           }
  }
```

jeffdiclemente:
Fixes #473

### DIFF
--- a/compendium/tools/SCRCodeGen/ComponentCallbackGenerator.hpp
+++ b/compendium/tools/SCRCodeGen/ComponentCallbackGenerator.hpp
@@ -103,10 +103,10 @@ private:
       for(const auto& ref : componentInfo.references)
       {
         if ((false == componentInfo.injectReferences)
-            || (ref.policy == "dynamic"))
+            || (ref.second.policy == "dynamic"))
         {
           mStrStream << "  binders.push_back("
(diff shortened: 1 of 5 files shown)


jeffdiclemente:
using an unordered_map changed the auto-generated code which broke the guarantees about the order of injected service dependencies.

jeffdiclemente:
yes, they are unrelated.

jeffdiclemente:
@carneyweb please review.

PR #478, opened 2020-06-19
Fix codecov

apatil19:
Issue: Code coverage was not getting generated for macOS when built with xcode.

Cause: gcov was unable to find any object file as Xcode generates those files in a different path than what gcov expects. Due to this mismatch in paths codecov reports were not generated. 

Solution: Stitching the path was not feasible as it would disturb the relative paths in code coverage report. Therefore, to generate code coverage, added a new osX configuration with cmake generator = 'Unix Makefiles'. This generates the code coverage report and also makes it easy to conserve the relative path.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,22 +89,28 @@ matrix:
       - os: osx
         osx_image: xcode10.3
         compiler: clang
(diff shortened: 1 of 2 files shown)


jeffdiclemente:
@kevinleeMW @carneyweb please review.

PR #482, opened 2020-07-01
Integrate dynamic ds

jeffdiclemente:
support dynamic rebinding of service references for 0..1 and 1..1 cardinality (mandatory unary and optional unary)


jeffdiclemente:
@CppMicroServices/developers see https://github.com/CppMicroServices/rfcs/blob/master/text/0003-declarative-services.md#-bind--unbind-callbacks- for a description of the bind/unbind callbacks and https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#d0e37698 for a table describing the rules for re-binding to a better target service

compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
```diff
@@ -0,0 +1,133 @@
...
+  // OSGi Compendium Release 7 section 112.5.12 Bound Service Replacement
+  //  If an active component configuration has a dynamic reference with unary
+  //  cardinality and the bound service is modified or unregistered and ceases
+  //  to be a target service, or the policy-option is greedy and a better
+  //  target service becomes available then SCR must attempt to replace the
+  //  bound service with a new bound service.
+  auto removeBoundRef = false;
+  std::vector<RefChangeNotification> notifications;
+
+  { // acquire lock on boundRefs
```
pelliott-mathworks, line 89: Just wondering why you are not using ShouldClearBoundRefs here

compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+void ReferenceManagerBaseImpl::BindingPolicy::ClearBoundRefs()
+{
+  auto boundRefsHandle = mgr.boundRefs.lock();
+  boundRefsHandle->clear();
+}
+
+void ReferenceManagerBaseImpl::BindingPolicy::StaticRemoveService(
+  const ServiceReferenceBase& reference)
```
pelliott-mathworks, line 56: Is it possible for this to be called with an invalid reference? Should it be checked? logged? Same question for DynamicRemoveService

compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
```diff
@@ -0,0 +1,133 @@
...
+    RefChangeNotification notification{ mgr.metadata.name,
+                                        RefEvent::BECAME_UNSATISFIED };
+    notifications.push_back(std::move(notification));
+
+    ClearBoundRefs();
+    if (mgr.UpdateBoundRefs()) {
+      Log("Notify SATISFIED for reference " + mgr.metadata.name);
+      RefChangeNotification notification{ mgr.metadata.name,
+                                          RefEvent::BECAME_SATISFIED };
+      notifications.push_back(std::move(notification));
```
kevinleeMW, line 70: Is there a reason why push_back with a std::move is used here instead of emplace_back?

compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
```diff
@@ -0,0 +1,133 @@
...
+      {
+        auto boundRefsHandle =
+          mgr.boundRefs.lock(); // acquires lock on boundRefs
+        if (!boundRefsHandle->empty()) {
+          svcRefToBind = *(boundRefsHandle->begin());
+          Log("Notify BIND for reference " + mgr.metadata.name);
+        }
+      }
+
+      Log("Notify UNBIND for reference " + mgr.metadata.name);
```
kevinleeMW, line 114: Should it log "Notify REBIND"?

compendium/DeclarativeServices/src/manager/ReferenceManagerImpl.cpp
```diff
@@ -384,5 +267,35 @@ void ReferenceManagerImpl::BatchNotifyAllListeners(const std::vector<RefChangeNo
...
+      return std::make_unique<BindingPolicyDynamicReluctant>(ref);
+    }
+    else { // greedy
+      return std::make_unique<BindingPolicyDynamicGreedy>(ref);
+    } 
+  }
+}
+
 }
 }
```
kevinleeMW, line 301: minor detail: are you planning to change the file name to ReferenceManagerBaseImpl?

compendium/DeclarativeServices/test/TestReferenceManagerImpl.cpp
```diff
@@ -170,10 +201,15 @@ TEST_P(ReferenceManagerImplTest, TestIsSatisfied)
...
     ReferenceManagerImpl refManager(fakeMetadata,
                                     GetFramework().GetBundleContext(),
-                                    fakeLogger, FakeComponentConfigName);
-    EXPECT_EQ(refManager.IsSatisfied(), (refManager.IsOptional() ? true : false)) << "Initial state is SATISFIED only if cardinality is optional";
-    auto reg = bc.RegisterService<dummy::Reference1>(std::make_shared<dummy::Reference1>());
-    EXPECT_EQ(refManager.IsSatisfied(), true) << "State expected to be SATISFIED after service registration";
+                                    fakeLogger,
+                                    FakeComponentConfigName);
+    EXPECT_EQ(refManager.IsSatisfied(),
+              (refManager.IsOptional() ? true : false))
```
kevinleeMW, line 207: Can this be simplified to:
EXPECT_EQ(refManager.IsSatisfied(), refManager.IsOptional())
?

compendium/test_bundles/TBDynamicRefPolicy/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+usFunctionCreateDSTestBundle(TBDynamicRefPolicy)
```
kevinleeMW, line 1: I didn't see TBDynamicRefPolicy sub-directory added to .. level CMakeList.txt.
Just wanted to make sure that TBDynamicRefPolicy wasn't added to CMakeLists.txt on purpose.

compendium/DeclarativeServices/src/manager/BindingPolicyStaticGreedy.cpp
```diff
@@ -0,0 +1,104 @@
...
+        replacementNeeded = mgr.IsOptional();
+      }
+    }
+  }
+
+  auto notifySatisfied = ShouldNotifySatisfied();
+  std::vector<RefChangeNotification> notifications;
+  if (replacementNeeded) {
+    Log("Notify UNSATISFIED for reference " + mgr.metadata.name);
+    RefChangeNotification notification{ mgr.metadata.name,
```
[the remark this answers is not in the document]
jeffdiclemente: correct. under a static reference policy, the bind and unbind callbacks are not used (which is what the REBIND change notification is for). Instead the entire service component is destroyed which is handled by the `BECAME_UNSATISFIED` change notification.

compendium/DeclarativeServices/src/manager/BindingPolicyStaticGreedy.cpp
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  // If no service is bound, reactivate the component to bind to the better target service.
+  // Otherwise, check if the service is a higher rank and if so, reactivate the component to bind
+  // to the better target service
+
+  auto replacementNeeded = false;
+  ServiceReference<void> serviceToUnbind;
+  if (mgr.IsSatisfied()) {
+    // means that a service is bound because the refmgr is satisfied.
```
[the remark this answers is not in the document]
jeffdiclemente: you're right. In the case where minCardinality is 0 a service doesn't have to be bound for the ref manager to be satisfied. this comment is not accurate.

compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
```diff
@@ -0,0 +1,133 @@
...
+  const ServiceReferenceBase& reference)
+{
+  auto boundRefsHandle = mgr.boundRefs.lock();
+  auto itr = boundRefsHandle->find(reference);
+  return (itr != boundRefsHandle->end());
+}
+
+bool ReferenceManagerBaseImpl::BindingPolicy::ShouldNotifySatisfied()
+{
+  return (mgr.IsSatisfied() ? false : mgr.UpdateBoundRefs());
```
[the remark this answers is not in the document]
jeffdiclemente: yes

compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
jeffdiclemente: I chose to log each bind and unbind separately so that any issues with the ordering of the bind and unbind callbacks can be spotted. Logging a "rebind" doesn't offer the same granularity of knowing whether a bind/unbind happened. There are cases where only an unbind will happen and others where a bind and unbind will happen.

compendium/DeclarativeServices/src/manager/BindingPolicyDynamicReluctant.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  if (!mgr.IsSatisfied()) {
+    notifySatisfied =
+      mgr.UpdateBoundRefs(); // becomes satisfied if return value is true
+  } else {
+    // previously satisfied
+    // If the service was previously satisfied then either there is
+    // nothing to do or a rebind needs to happen if the cardinality
+    // is optional and there are no bound refs.
+    if (mgr.IsOptional() && 0 == mgr.GetBoundReferences().size()) {
```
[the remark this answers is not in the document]
jeffdiclemente: calling `IsOptional` doesn't appear to be necessary. If `GetBoundReferences` returns 0 then that implies that the service reference is optional (since we're in the `else` because the service reference was previously satisfied).

compendium/test_bundles/TBDynamicRefPolicy/CMakeLists.txt
jeffdiclemente: good catch. this is an old test bundle that I forgot to delete.

compendium/DeclarativeServices/src/manager/ReferenceManagerImpl.cpp
jeffdiclemente: nope, at least not in this PR

compendium/DeclarativeServices/test/TestReferenceManagerImpl.cpp
jeffdiclemente: it can.

compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
jeffdiclemente: the reason is with how the `RefChangeNotification` struct was originally defined. It didn't allow the use of `emplace_back` without adding an appropriate constructor to the struct definition. I did add a constructor to `RefChangeNotification` (for other reasons) so I believe one could use `emplace_back` now.

compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
jeffdiclemente: no good reason, I overlooked it while refactoring. I'll use `ShouldClearBoundRefs`

compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
jeffdiclemente: it should not be possible for an invalid reference to be passed into this function. The service reference is guaranteed valid since its lifetime is scoped to the lifetime of the `ServiceRegistration` object which is responsible for un-registering services. While a service is being unregistered, service listeners are sent notifications (of which `DynamicRemoveService` and `StaticRemoveService` are listeners) before service references are destroyed.

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -191,14 +197,53 @@ void ComponentContextImpl::DisableComponent(const std::string& name)
...
+
+void ComponentContextImpl::AddToBoundServicesCache(
+  const std::string& refName,
+  const cppmicroservices::ServiceReferenceBase& sRef)
+{
+  auto bc = GetBundleContext();
+  cppmicroservices::ServiceObjects<void> sObjs =
+    bc.GetServiceObjects(ServiceReferenceU(sRef));
+  auto boundServicesCacheHandle = boundServicesCache.lock();
+  boundServicesCacheHandle->operator[](refName).emplace_back(
```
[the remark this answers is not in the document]
jeffdiclemente: this was already addressed. Take a look at the latest set of changes

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
```diff
@@ -185,17 +193,19 @@ class SatisfiedFunctor
   std::string skipKey;
 };
 
-void ComponentConfigurationImpl::RefSatisfied(const std::string& refName)
+void ComponentConfigurationImpl::RefSatisfied(const RefChangeNotification& notification)
```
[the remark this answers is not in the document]
jeffdiclemente: This is leftover from a time when I was using the RefSatisfied/RefUnsatisfied methods to perform bind/unbind. Looks like I never reverted this. I'll revert it.

compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
```diff
@@ -0,0 +1,133 @@
...
+  =============================================================================*/
+
+#include "ReferenceManagerImpl.hpp"
+#include "cppmicroservices/logservice/LogService.hpp"
+
+namespace cppmicroservices {
+namespace scrimpl {
+
+void ReferenceManagerBaseImpl::BindingPolicy::Log(
+  std::string&& logStr,
```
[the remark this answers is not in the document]
jeffdiclemente: A const ref should be just fine. I'll change it.

PR #485, opened 2020-07-14
Fix codecov uploader issue on appveyor

apatil19:
Issue: Codecov bash uploader was throwing an error on appveyor causing the build to fail.

Solution:
Appending codecov flags to fix codecov uploader issue on appveyor.
Also, wrapping the code in condition to make it run only when code coverage is needed.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -250,9 +250,11 @@ after_test:
 - if [%WITH_COVERAGE%]==[1] python %APPVEYOR_BUILD_FOLDER%/fixcoveragefilepaths.py -i coverage.xml -s %APPVEYOR_BUILD_FOLDER% -o coverage_pathsfixed.xml
 # use codecov bash script. Requires powershell and bash to be installed, which both are on the AppVeyor machine image
 - ps: |
-    $env:PATH = 'C:\msys64\usr\bin;' + $env:PATH
-    Invoke-WebRequest -Uri 'https://codecov.io/bash' -OutFile codecov.sh
-    bash codecov.sh -f "./coverage_pathsfixed.xml"
+    if($env:WITH_COVERAGE -eq 1){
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -250,9 +250,11 @@ after_test:
 - if [%WITH_COVERAGE%]==[1] python %APPVEYOR_BUILD_FOLDER%/fixcoveragefilepaths.py -i coverage.xml -s %APPVEYOR_BUILD_FOLDER% -o coverage_pathsfixed.xml
 # use codecov bash script. Requires powershell and bash to be installed, which both are on the AppVeyor machine image
 - ps: |
-    $env:PATH = 'C:\msys64\usr\bin;' + $env:PATH
-    Invoke-WebRequest -Uri 'https://codecov.io/bash' -OutFile codecov.sh
-    bash codecov.sh -f "./coverage_pathsfixed.xml"
+    if($env:WITH_COVERAGE -eq 1){
+      $env:PATH = 'C:\msys64\usr\bin;' + $env:PATH
+      Invoke-WebRequest -Uri 'https://codecov.io/bash' -OutFile codecov.sh
+      bash codecov.sh -f "./coverage_pathsfixed.xml" -U "-s" -A "-s"
```
jeffdiclemente, line 256: can you elaborate a little on the root cause of the failure and how adding -U "-s" -A "-s" solves the problem? Thanks.

apatil19: This is an issue with the way the bash script has been updated and with Powershell.
These additional parameters deal with the way Powershell handles curl. Apparently, Powershell doesn’t like the curl-like output, so the -s arguments silence them.

Please refer to following thread from codecov community:
https://community.codecov.io/t/bash-uploader-returns-error-code-on-appveyor/1682

PR #487, opened 2020-07-27
ConfigurationAdmin

achristoforides:
This PR migrates ConfigurationAdmin to open source.


compendium/CMakeLists.txt
```diff
@@ -4,6 +4,6 @@ endif()
 add_subdirectory(LogService)
 add_subdirectory(ServiceComponent)
 add_subdirectory(DeclarativeServices)
+add_subdirectory(ConfigurationAdmin)
+add_subdirectory(CM)
```
carneyweb, line 8: Why "CM" for the subdirectory? That doesn't seem to be in the spirit of the other subdirectories in compendium.

achristoforides: "CM" is used here since the osgi spec uses the 'cm' namespace when describing the features introduced by this PR (https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.cm.html#org.osgi.service.cm). The project itself always has the requirement of following the spec as it makes sense for c++ and using "CM" in our implementation is a part of the spec we can reasonably follow in c++ when introducing these changes.

carneyweb: Gotcha. Makes sense. Thanks.

compendium/CM/include/cppmicroservices/cm/ConfigurationAdmin.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+
+ =============================================================================*/
+
+#ifndef ConfigurationAdmin_hpp
```
pelliott-mathworks, line 23: Just wondering if ConfigurationAdmin_hpp is enough of a guard. Should include the component name to be sure?
CppMicroServices_ConfigurationAdmin.

achristoforides: In this case, it is fine. This file is part of the public interface of ConfigAdmin and none of the other interface header files use this header guard as they utilize their file names as the include guard. Additionally, none of the private headers in the implementation of ConfigAdmin use the same include guard so in this case we are okay. I believe the convention for this project is that the include guard should be the file name; there may be some special cases that need to be considered with other files but this isn't one of them from what I can tell.

compendium/CM/include/cppmicroservices/cm/Configuration.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+
+ =============================================================================*/
+
+#ifndef Configuration_hpp
```
achristoforides, line 23: Go through and update include guards to prefix with (CPPMS_)CM_ for public headers

compendium/CM/include/cppmicroservices/cm/ConfigurationException.hpp
```diff
@@ -0,0 +1,96 @@
...
+          , reason(std::move(rsn))
+          , property(std::move(prop))
+          {}
+
+        /**
+         * Returns the reason for this exception.
+         *
+         * @return The reason for this exception.
+         */
+        inline std::string GetReason() const
```
achristoforides, line 70: remove inline

compendium/CM/include/cppmicroservices/cm/ConfigurationException.hpp
```diff
@@ -0,0 +1,96 @@
...
+          return reason;
+        }
+
+        /**
+         * Returns the property which was resonsible for this exception being throws,
+         * if applicable. Could be empty.
+         *
+         * @return The property which caused this exception.
+         */
+        inline std::string GetProperty() const
```
achristoforides, line 81: remove inline

compendium/CM/include/cppmicroservices/cm/ConfigurationException.hpp
```diff
@@ -0,0 +1,96 @@
...
+        {
+          return property;
+        }
+
+        virtual ~ConfigurationException() noexcept {}
+
+      private:
+        std::string reason;
+        std::string property;
```
achristoforides, line 90: mark these as const

compendium/ConfigurationAdmin/test/Mocks.hpp
```diff
@@ -0,0 +1,99 @@
...
+     * This class is used in tests where the logger is required but the test
+     * does not care if anything is actually sent to the logger
+     */
+    class FakeLogger : public cppmicroservices::logservice::LogService
+    {
+    public:
+      void Log(cppmicroservices::logservice::SeverityLevel /* level */, const std::string& /* message */) override {}
+      void Log(cppmicroservices::logservice::SeverityLevel /* level */, const std::string& /* message */, const std::exception_ptr /* ex */) override {}
+      void Log(const ServiceReferenceBase& /* sr */, cppmicroservices::logservice::SeverityLevel /* level */, const std::string& /* message */) override {}
+      void Log(const ServiceReferenceBase& /* sr */, cppmicroservices::logservice::SeverityLevel /* level */, const std::string& /* message */, const std::exception_ptr /* ex */) override {}
```
achristoforides, line 60: Delete commented parameter names

PR #494, opened 2020-08-05
Support parentheses in LDAP query attribute value

### LINKED ISSUE #496: Support parentheses in LDAP query attribute value
LDAP queries currently fail when attribute value contains parentheses

A filter that has no errors: (prop=foobar)
A filter that throws an exception: (prop=foo(bar))

kevinleeMW:
- changed LDAPExpr implementation to account for parentheses in attribute value
- added test cases

Signed-off-by: Kevin Lee <kevinlee@mathworks.com>

fixes #496

### DIFF
--- a/framework/src/util/LDAPExpr.cpp
+++ b/framework/src/util/LDAPExpr.cpp
@@ -709,14 +709,23 @@ std::string LDAPExpr::ParseState::getAttributeName()
 
 std::string LDAPExpr::ParseState::getAttributeValue()
 {
+  int num_parens = 0;
   std::string sb;
   bool exit = false;
   while (!exit) {
     Byte c = peek();
     switch (c) {
       case '(':
+        num_parens+=1;
+        sb.append(1, c);
+        break;
       case ')':
-        exit = true;
+        if (num_parens > 0) {
+          num_parens-=1;
+          sb.append(1, c);
+        } else {
+          exit = true;
+        }
         break;
       case '*':
         sb.append(1, LDAPExprConstants::WILDCARD());
--- a/framework/src/util/LDAPExpr.cpp
+++ b/framework/src/util/LDAPExpr.cpp
@@ -716,12 +716,12 @@ std::string LDAPExpr::ParseState::getAttributeValue()
     Byte c = peek();
     switch (c) {
       case '(':
-        num_parens+=1;
+        ++num_parens;
         sb.append(1, c);
         break;
       case ')':
         if (num_parens > 0) {
-          num_parens-=1;
+          --num_parens;
           sb.append(1, c);
         } else {
           exit = true;
--- a/framework/test/driver/LDAPFilterTest.cpp
+++ b/framework/test/driver/LDAPFilterTest.cpp
@@ -46,6 +46,10 @@ int TestParsing()
                       "=Person)(|(sn=Jensen)(cn=Babs J*)))");
     US_TEST_OUTPUT(<< "Parsing (o=univ*of*mich*)")
     ldap = LDAPFilter("(o=univ*of*mich*)");
+    US_TEST_OUTPUT(<< "Parsing (prop=foo(bar))")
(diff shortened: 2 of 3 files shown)


framework/test/driver/LDAPFilterTest.cpp
```diff
@@ -46,6 +46,10 @@ int TestParsing()
                       "=Person)(|(sn=Jensen)(cn=Babs J*)))");
     US_TEST_OUTPUT(<< "Parsing (o=univ*of*mich*)")
     ldap = LDAPFilter("(o=univ*of*mich*)");
+    US_TEST_OUTPUT(<< "Parsing (prop=foo(bar))")
```
jeffdiclemente, line 49: there should also be at least one test case that tests invalid (i.e. non-matching parenthesis) input

framework/src/util/LDAPExpr.cpp
```diff
@@ -709,14 +709,23 @@ std::string LDAPExpr::ParseState::getAttributeName()
 
 std::string LDAPExpr::ParseState::getAttributeValue()
 {
+  int num_parens = 0;
   std::string sb;
   bool exit = false;
   while (!exit) {
     Byte c = peek();
```
pelliott-mathworks, line 716: peek looks at the next character in the string and returns that character or it will throw an out of range exception if m_pos is greater than the end of the string. It seems like the only way this while loop will exit is if a mismatched set of parentheses is detected or if an exception is thrown. That doesn't seem right. Am I missing something?

kevinleeMW: peek() will throw, if it goes out of bounds (i.e. didn't find the last ")", where exit is set to true)
the code was originally designed to check for any ")" to exit; now it only exits when the last matching ")" is found.

framework/test/driver/LDAPFilterTest.cpp
kevinleeMW: adding more test cases

framework/src/util/LDAPExpr.cpp
```diff
@@ -709,14 +709,23 @@ std::string LDAPExpr::ParseState::getAttributeName()
...
     switch (c) {
       case '(':
+        num_parens+=1;
```
carneyweb, line 719: ++num_parens; ?

framework/src/util/LDAPExpr.cpp
```diff
@@ -709,14 +709,23 @@ std::string LDAPExpr::ParseState::getAttributeName()
...
+        sb.append(1, c);
+        break;
       case ')':
-        exit = true;
+        if (num_parens > 0) {
+          num_parens-=1;
```
carneyweb, line 724: --num_parens; ?

framework/src/util/LDAPExpr.cpp
```diff
@@ -709,14 +709,23 @@ std::string LDAPExpr::ParseState::getAttributeName()
 
 std::string LDAPExpr::ParseState::getAttributeValue()
```
carneyweb, line 710: should getAttributeValue and getAttributeName be implemented with std::regex? I know they can be a pain, but they might be simpler and more expressive than doing character by character parsing.

kevinleeMW: I agree that we can look into re-implementing some parts of LDAPExpr/LDAPFilter.
Jeff and I briefly talked about this too, perhaps even using a third-party tool.

I think that work would be beyond the scope of this particular issue with parentheses, so I'll create a new issue to consider re-implementing LDAPFilter/LDAPExpr.

framework/src/util/LDAPExpr.cpp
kevinleeMW: Made changes accordingly

framework/src/util/LDAPExpr.cpp
kevinleeMW: Made changes accordingly

PR #497, opened 2020-08-06
ConfigAdmin Extra Tests

achristoforides:
This PR introduces another ConfigAdmin test and makes it open source.

Additionally, it fixes a Valgrind error which was present in TestMetadataParserImplV1.cpp

### DIFF
--- a/compendium/test_bundles/ManagedServiceAndFactoryBundle/src/ManagedServiceFactoryImpl.cpp
+++ b/compendium/test_bundles/ManagedServiceAndFactoryBundle/src/ManagedServiceFactoryImpl.cpp
@@ -12,19 +12,15 @@ TestManagedServiceFactoryImpl::~TestManagedServiceFactoryImpl() = default;
 void TestManagedServiceFactoryImpl::Updated(std::string const& pid, AnyMap const& properties) {
     std::lock_guard<std::mutex> lk(m_updatedMtx);
     if (properties.empty()) {
-        std::cout << "    Updated called for pid \"" << pid << "\" with empty properties" << std::endl;
         m_updatedCallCount[pid] -= 1;
     } else {
         auto const incrementBy = cppmicroservices::any_cast<int>(properties.AtCompoundKey("anInt"));
-        std::cout << "    Updated called for pid \"" << pid
-                  << "\" with value: " << std::to_string(incrementBy) << std::endl;
         m_updatedCallCount[pid] += incrementBy;
     }
 }
 
 void TestManagedServiceFactoryImpl::Removed(std::string const& pid) {
     std::lock_guard<std::mutex> lk(m_removedMtx);
-    std::cout << "    Removed called for pid \"" << pid << "\" with empty properties" << std::endl;
     ++m_removedCallCount[pid];
 }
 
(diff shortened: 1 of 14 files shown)


compendium/test_bundles/ManagedServiceAndFactoryBundle/src/ManagedServiceImpl.cpp
```diff
@@ -0,0 +1,39 @@
...
+    : m_counter{0} {
+}
+
+TestManagedServiceImpl::~TestManagedServiceImpl() = default;
+
+void TestManagedServiceImpl::Updated(AnyMap const& properties) {
+    std::lock_guard<std::mutex> lk(m_counterMtx);
+    if (properties.empty()) {
+        // Usually corresponds to the configuration being removed
+        std::cout << "    Updated called with empty properties" << std::endl;
```
carneyweb, line 21: Should we be outputting things during tests? Doesn't this clutter the logs?

compendium/test_bundles/TestInterfaces/include/TestInterfaces/Interfaces.hpp
```diff
@@ -112,6 +126,18 @@ namespace test
...
+      public:
+        TestManagedServiceFactory() = default;
+        virtual ~TestManagedServiceFactory() = default;
+
+        virtual int getUpdatedCounter(std::string const& pid) = 0;
+
+        virtual int getRemovedCounter(std::string const& pid) = 0;
+
+        virtual std::shared_ptr<TestManagedServiceFactoryServiceInterface> create(std::string const& config) = 0;
+    };
```
kevinleeMW, line 140: I think it'd be nice to have the "TestManaged..." interfaces all close to each other by moving this interface up to where the other two added interfaces are (around line 65)

compendium/test_bundles/ManagedServiceAndFactoryBundle/src/ManagedServiceFactoryImpl.cpp
```diff
@@ -0,0 +1,55 @@
...
+        auto const incrementBy = cppmicroservices::any_cast<int>(properties.AtCompoundKey("anInt"));
+        std::cout << "    Updated called for pid \"" << pid
+                  << "\" with value: " << std::to_string(incrementBy) << std::endl;
+        m_updatedCallCount[pid] += incrementBy;
+    }
+}
+
+void TestManagedServiceFactoryImpl::Removed(std::string const& pid) {
+    std::lock_guard<std::mutex> lk(m_removedMtx);
+    std::cout << "    Removed called for pid \"" << pid << "\" with empty properties" << std::endl;
```
pelliott-mathworks, line 27: Just wondering if it's a good idea to hold the lock during the cout operation?Maybe move this statement down and release the lock before it?

compendium/test_bundles/ManagedServiceAndFactoryBundle/src/ManagedServiceImpl.cpp
achristoforides: These aren't necessary and should be removed. Thanks!

compendium/test_bundles/ManagedServiceAndFactoryBundle/src/ManagedServiceFactoryImpl.cpp
achristoforides: These couts will be removed. The guard is necessary though (even if the cout were to stay) so that we don't have a race if multiple threads call Removed concurrently. If the guard was after the couts, a race condition would exist with respect to the cout. Thanks!

PR #499, opened 2020-08-12
LogService Implementation

achristoforides:
This PR introduces an implementation for the previously unimplemented LogService. To accomplish this, a new third-party library is added to the project to facilitate much of the logging facilities. 

The library is spdlog and all house-keeping tasks regarding adding a new library have been done. Additionally, it is important to note that this PR does not add spdlog as a git submodule; instead it just adds a copy of the include folder into our project since it can be used as a header only library in addition to the fact that we aren't statically linking or dynamically linking to the library.

A comprehensive test suite was added as well and confirms the desired behavior of the logger's internals.


compendium/LogServiceImpl/test/TestLogService.cpp
```diff
@@ -0,0 +1,162 @@
...
+  EXPECT_FALSE(ContainsRegex(log_preamble + "Test invalid maximum severity level"));
+
+  ASSERT_NO_THROW(logger->Log(ls::SeverityLevel::LOG_INFO, "Test invalid exception_ptr", nullptr));
+  EXPECT_TRUE(ContainsRegex(log_preamble + "Test invalid exception_ptr(\\n)" + exception_preamble + "nullptr"));
+
+  ASSERT_NO_THROW(logger->Log(cppmicroservices::ServiceReferenceU{}, ls::SeverityLevel::LOG_INFO, "Test invalid ServiceReferenceBase object", nullptr));
+  EXPECT_TRUE(ContainsRegex(log_preamble + "Test invalid ServiceReferenceBase object(\\n)" + svcRef_preamble + "Invalid service reference(\\n)" + exception_preamble + "nullptr"));
+}
+
+TEST_F(LogServiceImplTests, ThreadSaftey) {
```
jeffdiclemente, line 139: typo: ThreadSafety

compendium/LogServiceImpl/test/CMakeLists.txt
```diff
@@ -0,0 +1,155 @@
...
+endif()
+
+target_link_libraries(${us_logservice_test_exe_name}
+  PRIVATE
+  ${GTEST_BOTH_LIBRARIES}
+  ${GMOCK_BOTH_LIBRARIES}
+  ${${PROJECT_NAME}_LINK_LIBRARIES}
+  CppMicroServices
+  usLogService
+  usServiceComponent
```
jeffdiclemente, line 85: I don't think DS is required

compendium/LogServiceImpl/src/LogServiceImpl.cpp
```diff
@@ -0,0 +1,100 @@
...
+namespace cppmicroservices {
+namespace logservice {
+std::string GetExceptionMessage(const std::exception_ptr& ex)
+{
+  std::string message = "\nException logged: ";
+  if (ex) {
+    std::ostringstream stream;
+    try {
+      std::rethrow_exception(ex);
+    } catch (const std::exception& e) {
```
carneyweb, line 19: Not sure a stringstream buys us anything here. I think just doing “message += typeid(e).name() + e.what()” would be fine.

compendium/LogServiceImpl/src/LogServiceImpl.cpp
```diff
@@ -0,0 +1,100 @@
...
+    message += "none";
+  }
+
+  return message;
+}
+
+std::string GetServiceReferenceInfo(const ServiceReferenceBase& sr)
+{
+  std::ostringstream stream;
+  stream << "\nServiceReference: " << sr;
```
carneyweb, line 33: Same thing here...

std::string rval = “\nServiceReference: “ + sr;
return rval;

achristoforides: We use the operator<< for the ServiceReferenceBase object to fill out the string. If we were to try and concatenate sr as you mentioned, this wouldn't compile.

PR #502, opened 2020-08-19
AnyMapTest migration to GTest

achristoforides:
This PR migrates the AnyMapTest (originally found in the test driver) to the usFrameworkTests gtest suite. An existing test for the AnyMap already existed so the tests from the test driver were added to this existing suite.


framework/test/driver/AnyMapTest.cpp
```diff
@@ -1,145 +0,0 @@
...
-   *       {
-   *         hi : "hi",
-   *         there : "there"
-   *       }
-   *     ]
-   *   },
-   *   dot.key : 5
-   * }
-   *
-   */
```
jeffdiclemente, line 71: it would be helpful to also include this comment in the moved test.

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -250,6 +250,83 @@ TEST(AnyMapTest, CIHashUnique)
...
+
+  auto it =
+    om.insert(std::make_pair(std::string("key1"), Any(std::string("val1"))));
+  ASSERT_TRUE(it.second);
+  ASSERT_EQ("key1", it.first->first);
+  ASSERT_EQ(std::string("val1"), any_cast<std::string>(it.first->second));
+  ASSERT_EQ(std::string("val1"), any_cast<std::string>(om["key1"]));
+
+  AnyMap uoci(AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS);
+  uoci["FIRST"] = 1;
```
jeffdiclemente, line 274: in the original test, the "I" is lower case. I think this was intentional since this is a case insensitive match and there is a test checking for "FIRST" instead of "FiRST". I'd suggest maintaining this in ported gtest.

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -250,6 +250,83 @@ TEST(AnyMapTest, CIHashUnique)
...
+            any_cast<std::string>(om.AtCompoundKey("uoci.Vec.2.there")));
+
+  std::set<std::string> keys;
+  for (auto p : uoci) {
+    keys.insert(p.first);
+  }
+
+  auto key = keys.begin();
+  ASSERT_EQ(3, keys.size());
+  ASSERT_EQ("FIRST", *key++);
```
jeffdiclemente, line 313: should be "FiRST"

PR #509, opened 2020-09-10
Add ds thread pool

jeffdiclemente:
The introduction of a thread pool is meant to offer better performance in cases where declarative services is running on many core machines and the creation of a thread per std::async call is undesirable. 

boost asio was chosen because it is a robust and stable libraries with a compatible software license which provides a thread pool implementation.


compendium/DeclarativeServices/src/SCRActivator.cpp
```diff
@@ -82,13 +86,18 @@ void SCRActivator::Stop(cppmicroservices::BundleContext context)
...
     // clear bundle registry
     {
       std::lock_guard<std::mutex> l(bundleRegMutex);
       bundleRegistry.clear();
     }
     // clear component registry
     componentRegistry->Clear();
+
+    // There should be no more DS operations which involve a thread pool at this point.
+    threadpool->join();
```
jeffdiclemente, line 99: need to guarantee threadpool->join()is called if something above this throws.

jeffdiclemente:
For some reason the travis CI build status is not showing up. It passed with known, unrelated code coverage failure and can be viewed here: https://travis-ci.org/github/CppMicroServices/CppMicroServices/builds/729397726

compendium/DeclarativeServices/src/manager/states/CMDisabledState.cpp
```diff
@@ -41,44 +41,7 @@ CMDisabledState::CMDisabledState()
...
-                   if (ptr) {
-                     std::rethrow_exception(ptr);
-                   }
-                 })
-        .share();
-    return futObj;
-  }
-  // return the stored future in the current enabled state object
-  return currentState->GetFuture();
+  return cm.PostAsyncDisabledToEnabled(currentState);
```
kevinleeMW, line 44: love this!

kevinleeMW (approved):
looked through all the files except the ones under third_party/boost
looks good! love the ScopeGuard changes.

pelliott-mathworks (approved):
Line 48 of SCRBundleExtension.cpp Just wondering if the threadpool argument to the constructor should be checked too? Same comment/question line 51 of ComponentManagerImpl.cpp

jeffdiclemente:
> Line 48 of SCRBundleExtension.cpp Just wondering if the threadpool argument to the constructor should be checked too? Same comment/question line 51 of ComponentManagerImpl.cpp

sure, i'll add checks

PR #510, opened 2020-09-23
Fix manifest bundle version

### LINKED ISSUE #498: Bundle version not properly populating in generated manifests for DS and ConfigAdmin
When introspecting the build area after building the project, the generated manifest.json files for both Declarative Services and Configuration Admin do not contain the specified version found in the makefiles which generate those bundles. Instead of the version string being inserted by CMake, and empty string is in its place.

jeffdiclemente:
Fixes #498

Set the bundle version in manifest.json automatically for any bundle using the usMacroCreateBundle macro.


kevinleeMW (approved):

pelliott-mathworks:
Looks good!

PR #512, opened 2020-10-27
Configadmin ds integration

pelliott-mathworks:
This is prototype code to add ConfigurationListener functionality to DS and ConfigAdmin (Patty) and ListConfigurations code to ConfigAdmin(Kevin) in preparation for integrating ConfigAdmin into DS


compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -300,9 +325,29 @@ namespace cppmicroservices {
...
 
-    std::vector<std::shared_ptr<cppmicroservices::service::cm::Configuration>> ConfigurationAdminImpl::ListConfigurations(const std::string& /* filter */)
+    std::vector<std::shared_ptr<cppmicroservices::service::cm::Configuration>> ConfigurationAdminImpl::ListConfigurations(const std::string& filter)
     {
-      throw std::invalid_argument("Method not currently implemented");
+      std::vector<std::shared_ptr<cppmicroservices::service::cm::Configuration>> result;
+      {
+        std::lock_guard<std::mutex> lk {configurationsMutex};
+
+        if (filter.empty()) {
```
pelliott-mathworks, line 334: Can the filter string be invalid? Should we test for that? The OSGI spec says that ListConfigurations is supposed to throw a InvalidSyntaxException if the filter string is invalid.

kevinleeMW: Creating an LDAPFilter instance using "filter" string that is invalid throws InvalidSyntaxException.
It also throws when an empty string is passed in. That's why I check for an empty string first, to see if the user didn't pass in an argument or passed in an empty string (no filter).

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -300,9 +325,29 @@ namespace cppmicroservices {
...
+          result.reserve(configurations.size());
+          for (auto it : configurations) {
+            result.push_back(it.second);
+          }
+        } else {
+          LDAPFilter ldap(filter);
```
kevinleeMW, line 340: Mike: Use {} instead of ()

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
jeffdiclemente: The LDAPFilter ctor throws `std::invalid_argument`. `std::invalid_argument` can be thrown instead of `InvalidSyntaxException`.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.hpp
```diff
@@ -119,7 +120,7 @@ namespace cppmicroservices {
        *
        * See {@code ConfigurationAdmin#ListConfigurations}
        */
-      std::vector<std::shared_ptr<cppmicroservices::service::cm::Configuration>> ListConfigurations(const std::string& filter) override;
+      std::vector<std::shared_ptr<cppmicroservices::service::cm::Configuration>> ListConfigurations(const std::string& filter = std::string()) override;
```
kevinleeMW, line 123: make sure to modify doxygen documentation to reflect change

compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
```diff
@@ -172,7 +172,24 @@ namespace cppmicroservices {
...
+      EXPECT_NO_THROW(conf2->Update(props2));
+
+      const auto res1 = configAdmin.ListConfigurations();
+      const auto res2 = configAdmin.ListConfigurations("(foo=bar)");
+      const auto res3 = configAdmin.ListConfigurations("(foobar=baz)");
+
+      EXPECT_EQ(res1.size(), 2);
+      EXPECT_EQ(res2.size(), 1);
+      EXPECT_EQ(res2[0]->GetPid(), pid2);
+      EXPECT_TRUE(res3.empty());
```
kevinleeMW, line 192: add a testpoint to make sure invalid_argument exception is thrown properly

compendium/CM/include/cppmicroservices/cm/ConfigurationListener.hpp
```diff
@@ -0,0 +1,109 @@
...
+   */
+  const std::string getPid() const { return pid; }
+  /**
+   * Get the Factory PID which is responsible for this Configuration.
+   */
+  const std::string getFactoryPid() const { return factoryPid; }
+  /**
+   * Get the type of this Configuration.
+   */
+  const ConfigurationEventType getType() const { return type; }
```
pelliott-mathworks, line 83: no need for const on return parameter

compendium/CM/include/cppmicroservices/cm/ConfigurationListener.hpp
```diff
@@ -0,0 +1,109 @@
...
+class ConfigurationListener
+{
+public:
+  /**
+   * Called whenever the Configuration for any service is updated or removed from ConfigurationAdmin,
+   * and when the ConfigurationListener is first registered with the Framework, to provide the initial Configuration.
+   */
+  virtual void configurationEvent(ConfigurationEvent& event) = 0;
+};
+}
```
pelliott-mathworks, line 106: add comments to indicate which namespace is being closed.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -80,6 +80,24 @@ namespace {
     }
   }
 
+  void notifyListener(
```
pelliott-mathworks, line 83: notifyListener is only called in one place. Move code into calling function.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -80,6 +80,24 @@ namespace {
...
+    cppmicroservices::service::cm::ConfigurationListener& configListener)
+  {
+    try {
+      auto configEvent =
+        std::make_unique<cppmicroservices::service::cm::ConfigurationEvent>(
+          configAdmin, type, factoryPid, pid);
+
+      configListener.configurationEvent((*configEvent));
+    } catch (...) {
+      //todo
```
pelliott-mathworks, line 98: todo functionality should log an event and include the exception on the logged event

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -199,6 +221,9 @@ namespace cppmicroservices {
 
       managedServiceFactoryTracker.Close();
       managedServiceTracker.Close();
+      if (configListenerTracker) {
```
pelliott-mathworks, line 224: trackers can throw an exception on Close. Surround these 3 Close statements with try/catch so the destructor does not throw an exception and log the exception using the logger

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -462,6 +507,22 @@ namespace cppmicroservices {
...
+          removed
+            ? cppmicroservices::service::cm::ConfigurationEventType::CM_REMOVED
+            : cppmicroservices::service::cm::ConfigurationEventType::CM_UPDATED;
+
+        for (auto it : configurationListeners) {
+          notifyListener(
+            pid,
+            "",
+            type,
+            cmContext.GetServiceReference("ConfigurationAdminImpl"),
```
pelliott-mathworks, line 523: Use templatised version of GetServiceReference for ConfigurationAdminImpl instead of string version. Compiler more able to find errors.

compendium/DeclarativeServices/src/ConfigurationListenerImpl.cpp
```diff
@@ -0,0 +1,30 @@
...
+  , registry(std::move(componentRegistry))
+  , logger(std::move(logger))
+{
+  if (!bundleContext || !registry || !(this->logger)) {
+    throw std::invalid_argument("ConfigurationListenerImpl Constructor "
+                                "provided with invalid arguments");
+  }
+}
+
+void ConfigurationListenerImpl::configurationEvent(ConfigurationEvent& event) {
```
pelliott-mathworks, line 22: add const to event parameter

pelliott-mathworks (commented):
Comments from code review from team

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -300,9 +314,29 @@ namespace cppmicroservices {
...
+    std::vector<std::shared_ptr<cppmicroservices::service::cm::Configuration>> ConfigurationAdminImpl::ListConfigurations(const std::string& filter)
     {
-      throw std::invalid_argument("Method not currently implemented");
+      std::vector<std::shared_ptr<cppmicroservices::service::cm::Configuration>> result;
+      {
+        std::lock_guard<std::mutex> lk {configurationsMutex};
+
+        if (filter.empty()) {
+          result.reserve(configurations.size());
+          for (auto it : configurations) {
```
carneyweb, line 325: Is this right? if "configurations" is a vector, what's used by the range-based for isn't an iterator.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -300,9 +314,29 @@ namespace cppmicroservices {
...
+            result.push_back(it.second);
+          }
+        } else {
+          LDAPFilter ldap(filter);
+          for (auto it : configurations) {
+            auto props = it.second->GetProperties();
```
carneyweb, line 331: Is this right? if "configurations" is a vector, what's used by the range-based for isn't an iterator.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.hpp
carneyweb, line 123: Prefer {} to () for initializer lists.

carneyweb (changes requested):

compendium/DeclarativeServices/src/ConfigurationListenerImpl.cpp
```diff
@@ -5,61 +5,122 @@ namespace cppmicroservices {
...
                                 "provided with invalid arguments");
   }
 }
 
-void ConfigurationListenerImpl::configurationEvent(const ConfigurationEvent& event) {
+void ConfigurationListenerImpl::configurationEvent(const ConfigurationEvent& event)
+{
+  
+  {
+    if (listenersMap.lock()->empty())
```
carneyweb, line 33: Why is this code scoped?

carneyweb (approved):

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.hpp
```diff
@@ -177,6 +177,12 @@ class ComponentConfigurationImpl
    */
   virtual std::shared_ptr<ComponentInstance> CreateAndActivateComponentInstance(const cppmicroservices::Bundle& bundle) = 0;
 
+  /**
+   * Method called to modify the configuration properties for this component configuration. Subclasses
+   * must implement this method. Returns false if the component instance has not provided a Modified method.
+   */
+  virtual bool ModifyComponentInstanceProperties() = 0;
```
carneyweb, line 184: The comment should use the @return javadoc tag to describe the return values.

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.hpp
```diff
@@ -226,9 +232,19 @@ class ComponentConfigurationImpl
 
   /**
    * Method used to update the properties of a component instance when a configuration
-   * object on which it is dependent changes. 
+   * object on which it is dependent changes. Returns
```
carneyweb, line 235: The comment should use the @return javadoc tag to describe the return values.

compendium/DeclarativeServices/src/manager/BundleOrPrototypeComponentConfiguration.hpp
```diff
@@ -64,6 +64,11 @@ class BundleOrPrototypeComponentConfigurationImpl final
    */
   std::shared_ptr<ComponentInstance> CreateAndActivateComponentInstance(const cppmicroservices::Bundle& bundle) override;
 
+   /**
+   * Method called to modify the configuration properties for this component configuration. 
+   * Returns false if the component instance has not provided a Modified method.
```
carneyweb, line 69: The comment should use the @return javadoc tag to describe the return values.

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.hpp
```diff
@@ -177,6 +177,12 @@ class ComponentConfigurationImpl
    */
   virtual std::shared_ptr<ComponentInstance> CreateAndActivateComponentInstance(const cppmicroservices::Bundle& bundle) = 0;
 
+  /**
+   * Method called to modify the configuration properties for this component configuration. Subclasses
+   * must implement this method. Returns false if the component instance has not provided a Modified method.
```
carneyweb, line 182: The comment should use the @return javadoc tag to describe the return values.

compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.hpp
```diff
@@ -66,6 +66,12 @@ class SingletonComponentConfigurationImpl final
    */
   std::shared_ptr<ComponentInstance> CreateAndActivateComponentInstance(const cppmicroservices::Bundle& bundle) /* noexcept */ override;
 
+  /**
+   * Method called to modify the configuration properties for this component configuration. 
+   * Returns false if the component instance has not provided a Modified method.
+   */
+  bool ModifyComponentInstanceProperties() override;
```
carneyweb, line 73: The comment should use the @return javadoc tag to describe the return values.

compendium/DeclarativeServices/src/manager/states/CCActiveState.hpp
```diff
@@ -53,11 +53,16 @@ class CCActiveState final
   std::shared_ptr<ComponentInstance> Activate(ComponentConfigurationImpl& mgr,
                                               const cppmicroservices::Bundle& clientBundle) override;
 
+  void Deactivate(ComponentConfigurationImpl& mgr) override;
+
   /**
    * Modifies the properties of the component instance when a configuration object on 
-   * which it is dependent changes. No state change. 
+   * which it is dependent changes. No state change. Returns
```
carneyweb, line 60: The comment should use the @return javadoc tag to describe the return values.

compendium/DeclarativeServices/src/manager/states/CCActiveState.hpp
```diff
@@ -78,12 +83,12 @@ class CCActiveState final
    */
   ComponentState GetValue() const override { return ComponentState::ACTIVE; }
 
-  void WaitForTransitionTask() override
-  {
-    latch.Wait();
-  }
+ 
 private:
-  CounterLatch latch;
```
carneyweb, line 80: Just a question: what was "latch" used for, and what's not happening now that it was removed?

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
```diff
@@ -263,17 +305,17 @@ class ComponentInstanceImpl final
 
   // this method is used when injection is false and default constructor is provided by the implementation class
   template <class C = T, class I = Injection,
-            class X = typename std::enable_if<I::value == false>::type,
-            class Y = typename std::enable_if<std::is_default_constructible<C>::value == true>::type>
+            class InjectionFalse = typename std::enable_if<I::value == false>::type,
+            class THasDefaultConstructor = typename std::enable_if<std::is_default_constructible<C>::value == true>::type>
```
carneyweb, line 309: Perhaps "TIsDefaultConstructible" to match std::is_default_constrictible?

carneyweb (commented):
All of my comments below are cosmetic. Don't hold up the submission to fix them, but if you'll be making other changes it might be worth it.

compendium/DeclarativeServices/src/SCRBundleExtension.cpp
```diff
@@ -76,8 +86,55 @@ SCRBundleExtension::SCRBundleExtension(const cppmicroservices::BundleContext& bu
...
+    auto configurationMetadata =
+        metadataParser->ParseAndGetConfigurationMetadata(cmMetadata);
+
+    auto sr = this->bundleContext.GetServiceReference<
+        cppmicroservices::service::cm::ConfigurationAdmin>();
+    auto configAdmin =
+        this->bundleContext
+        .GetService<cppmicroservices::service::cm::ConfigurationAdmin>(sr);
+
+    for (auto& oneCmMetadata : configurationMetadata) {
```
jeffdiclemente, line 119: make this `const auto& oneCmMetadata` if possible.

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
```diff
@@ -0,0 +1,172 @@
...
+        if (it != configProperties.end()) {
+          configProperties.erase(it);
+        }
+        configProperties.emplace(pid, properties);
+        for (const auto item : properties) {
+          mergedProperties[item.first] = item.second;
+        }   
+      }      
+    } 
+  }
```
jeffdiclemente, line 79: minor: formatting

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
```diff
@@ -0,0 +1,172 @@
...
+  } catch (...) {
+    // No ConfigAdmin available
+               
+    return;
```
jeffdiclemente, line 83: probably want to send a message to the log service for diagnostics.

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+  std::lock_guard<std::mutex> lock(propertiesMutex);
+
+  return mergedProperties;
+}
+
+void ConfigurationManager::Initialize() {
+  if ((metadata->configurationPids.empty()) ||
+      (metadata->configurationPolicy == metadata::ComponentMetadata::CONFIG_POLICY_IGNORE))
+    return;
```
jeffdiclemente, line 52: always use brackets even for one line if statements.

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
```diff
@@ -0,0 +1,172 @@
...
+cppmicroservices::AnyMap ConfigurationManager::GetProperties() const noexcept
+{
+  std::lock_guard<std::mutex> lock(propertiesMutex);
+
+  return mergedProperties;
+}
+
+void ConfigurationManager::Initialize() {
+  if ((metadata->configurationPids.empty()) ||
+      (metadata->configurationPolicy == metadata::ComponentMetadata::CONFIG_POLICY_IGNORE))
```
jeffdiclemente, line 51: what happens if the pids are empty and the config policy is set to "require"? Is this possible at runtime? or does the tooling error out at build time (i.e. via the DS code gen tool)?

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/ConfigurationListenerImpl.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include "manager/ConfigurationNotifier.hpp"
+
+namespace cppmicroservices {
+namespace service {
+namespace cm {
+
+  /** ConfigChangeNotification
+     * This class is used by ConfigurationListener to notify ComponentConfigurationImpl
+     * about changes to Configuration Objects.
+     */
```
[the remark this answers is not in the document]
pelliott-mathworks: Fixed

compendium/DeclarativeServices/src/SCRBundleExtension.cpp
pelliott-mathworks: Fixed

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.hpp
pelliott-mathworks: Fixed

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
pelliott-mathworks: This combination (configuration-policy = "require", configuration-pid = [""],) will thrown an exception that will be logged by MetadataParserImplV1::ParseAndGetComponentsMetadata The component won't be loaded.

If just configuration-policy is present and configuration-pid is missing two warning messages are generated. One by the source code generator during the build  and one by MetadataParserImplV1::CreateComponentMetadata during runtime (which goes in the log). The component will be created with configurationPolicy = "ignore" in this case.  

Let me know if you want me to change the empty configuration-pid behavior to be consistent with the missing configuration-pid behavior.

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
pelliott-mathworks: Fixed

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
pelliott-mathworks: Fixed

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
```diff
@@ -113,9 +103,107 @@ TEST_F(tServiceComponent, testUpdateConfig_Modified) //DS_CAI_FTC_1
...
+      return false;
+    });
+
+  ASSERT_TRUE(result) << "Timed out waiting for Update Configuration"
+                         "to complete.";
+
+  compConfigs = GetComponentConfigs(testBundle, componentName, compDescDTO);
+  EXPECT_EQ(compConfigs.size(), 1ul) << "One default config expected";
+  EXPECT_EQ(compConfigs.at(0).state, scr::dto::ComponentState::SATISFIED)
+    << "Component state should be SATISFIED";
```
pelliott-mathworks, line 138: DS is supposed to log the exception thrown by the Modified method. Is there a way to confirm that in a test?

pelliott-mathworks (commented):
Looks good

compendium/DeclarativeServices/test/TestManifestWithInvalidConfiguration.cpp
```diff
@@ -0,0 +1,129 @@
...
+    GetInstance<cppmicroservices::service::cm::ConfigurationAdmin>();
+  ASSERT_TRUE(configAdminService)
+    << "GetService failed for ConfigurationAdmin.";
+
+  // Create configuration object, try to update property
+  // and property update should be ignored.
+  auto configuration = configAdminService->GetConfiguration(componentName);
+  auto configInstance = configuration->GetPid();
+
+  // Confirm configuratino object presented and check component state.
```
pelliott-mathworks, line 59: Can delete this code.

compendium/DeclarativeServices/test/TestManifestWithInvalidConfiguration.cpp
```diff
@@ -0,0 +1,129 @@
...
+  EXPECT_EQ(compConfigs.size(), 1ul) << "One default config expected.";
+  EXPECT_EQ(compConfigs.at(0).state, scr::dto::ComponentState::ACTIVE)
+    << "Ignore policy component state should be ACTIVE";
+
+  auto configAdminService =
+    GetInstance<cppmicroservices::service::cm::ConfigurationAdmin>();
+  ASSERT_TRUE(configAdminService)
+    << "GetService failed for ConfigurationAdmin.";
+
+  // Create configuration object, try to update property
```
pelliott-mathworks, line 54: This comment doesn't match code

compendium/DeclarativeServices/test/TestManifestWithInvalidConfiguration.cpp
```diff
@@ -0,0 +1,129 @@
...
```
pelliott-mathworks, line 104: Should move all of these lines to after the GetInstance on line 122. What we want this test to confirm is that DS does not call the component instance Modified method when configuration-policy is "ignore". To confirm that you have to get the component instance first and then update the configuration object.

compendium/DeclarativeServices/test/TestManifestWithInvalidConfiguration.cpp
```diff
@@ -0,0 +1,129 @@
...
+  auto configuration = configAdminService->GetConfiguration(componentName);
+  auto configInstance = configuration->GetPid();
+
+  cppmicroservices::AnyMap props(
+    cppmicroservices::AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS);
+  const std::string instanceId{ "instance1" };
+  props["uniqueProp"] = instanceId;
+  configuration->Update(props);
+
+  // Confirm configuratino object presented and check component state.
```
pelliott-mathworks, line 115: Can delete these lines. Not necessary for this test.

framework/src/bundle/BundleRegistry.cpp
```diff
@@ -52,9 +52,7 @@ class InitialBundleMapCleanup
   InitialBundleMapCleanup(std::function<void()> cleanupFcn)
```
pelliott-mathworks, line 52: Just checking to make sure you meant to include this file with your submission. Looks like it's part of your other change?

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
```diff
@@ -0,0 +1,517 @@
...
+#include "cppmicroservices/ServiceObjects.h"
+
+#include "TestInterfaces/Interfaces.hpp"
+
+namespace test {
+/**
+   * Verify that a service's configuration can be updated after the service is activated
+   * without deactivating and reactivating the service.
+   */
+TEST_F(tServiceComponent, testUpdateConfig_Modified) //DS_CAI_FTC_1
```
jeffdiclemente, line 36: I don't see how this test guarantees that the service component wasn't deactivated and activated again when updating the configuration.

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
jeffdiclemente: you could implement a LogService for the test and use that to verify that an exception was thrown.

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
```diff
@@ -0,0 +1,517 @@
...
+  props["uniqueProp"] = instanceId2;
+  fut = configuration->Update(props);
+  fut.get();
+  compConfigs = GetComponentConfigs(testBundle, componentName, compDescDTO);
+  EXPECT_EQ(compConfigs.at(0).state, scr::dto::ComponentState::SATISFIED)
+    << "Component state should be SATISFIED";
+
+  // This will cause DS to construct the instance with the updated properties.
+  auto newInstance = GetInstance<test::CAInterface>();
+  ASSERT_TRUE(newInstance) << "GetService failed for CAInterface";
```
jeffdiclemente, line 293: if testing for activation, I would think that the component should be checked that its in the scr::dto::ComponentState::ACTIVE state.

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/runtime/dto/ComponentConfigurationDTO.hpp
```diff
@@ -45,7 +45,7 @@ namespace dto {
  * \addtogroup gr_componentconfigurationdto
  * @{
  */
-enum class ComponentState : uint8_t
+enum US_ServiceComponent_EXPORT ComponentState
```
jeffdiclemente, line 48: this change should be reverted.   scoped enumerations should be used.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/TestCounterLatch.cpp
```diff
@@ -0,0 +1,111 @@
+/*=============================================================================
```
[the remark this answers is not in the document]
pelliott-mathworks: removed the file from compendium/DeclarativeServices/test.

compendium/DeclarativeServices/test/TestFixture.hpp
```diff
@@ -32,20 +32,23 @@
...
+#include "cppmicroservices/cm/ConfigurationAdmin.hpp"
 #include "cppmicroservices/servicecomponent/ComponentConstants.hpp"
 #include "cppmicroservices/servicecomponent/runtime/ServiceComponentRuntime.hpp"
-
-#include "TestUtils.hpp"
+#include <chrono>
 
 namespace test {
 
+auto const DEFAULT_POLL_PERIOD = std::chrono::milliseconds(1000);
```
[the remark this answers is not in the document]
pelliott-mathworks: It's used in TestConfigAdmin.cpp to wait for update operations to complete that were initiated by configuration objects being defined in the manifest.json file in "cm" sections.

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
pelliott-mathworks: Line 89 checks that following the Update operation the component is still in the ACTIVE state. Since immediate=false in the manifest.json file, if this component had been deactivated it would be in the SATISFIED state, waiting for a user to do a GetService to request an instance.

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
pelliott-mathworks: Lines 295 - 301 use the GetProperties method of the component instance to verify the property values. It wouldn't be able to execute this method if the component wasn't ACTIVE. We thought this was enough of a confirmation that the component is active but we could add a state check to confirm that the state is scr::dto::ComponentState::ACTIVE if you think we should.

jeffdiclemente: I think it's better not to rely on an indirect behavior (i.e. the behavior of `GetProperties`) to verify a test point and instead query the state directly.

pelliott-mathworks: Done

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
```diff
@@ -302,216 +308,203 @@ TEST_F(tServiceComponent,
...
-  std::string testBundleName = "TestBundleDSCA26";
-  scopeValidation(componentName,
-                  testBundleName,
-                  true,
-                  cppmicroservices::Constants::SCOPE_BUNDLE,
-                  2);
-}
-
-TEST_F(tServiceComponent,
-       testUpdateConfig_BundleScope_WithoutModified) //DS_CAI_FTC_27
```
jeffdiclemente: This test was re-factored heavily and I don't see these two test points (testUpdateConfig_BundleScope_Modified and testUpdateConfig_BundleScope_WithoutModified)  in this file.  are these tests still needed?

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.hpp
```diff
@@ -35,7 +35,7 @@ namespace scrimpl {
  * singleton instance of the {@link ComponentInstance}. It also implements the ServiceFactory pattern
  * to handle {@link ServiceFactory#GetService} calls from the user.
  */
-class SingletonComponentConfigurationImpl final
+class SingletonComponentConfigurationImpl
```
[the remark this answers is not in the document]
pelliott-mathworks: One of the other ways I looked at to test the Modified Exception Logging was deriving a class from SingletonComponentConfigurationImpl so I had to remove the final keyword. I ended up solving the problem differently and forgot to put it back. Fixing it now...

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
pelliott-mathworks: The code being tested in DS is in BundleOrPrototypeComponentConfiguration.cpp in the construction of the component instance and the modification of a component instance. Bundle and Prototype both share all of this code so only one set of tests is needed. The BundleScope tests weren't testing any additional lines of code. The other problem with the BundleScope tests was that the developer created a second framework to test them. Not sure if there is another way to test a service with Bundle Scope but testing it by creating a second framework was problematic.

jeffdiclemente (approved):

compendium/DeclarativeServices/test/TestSingletonComponentConfiguration.cpp
```diff
@@ -205,6 +227,51 @@ TEST_F(SingletonComponentConfigurationTest, TestGetService)
...
+  // When the mock DoesModifiedMethodExist method is called it will return true;
+  EXPECT_CALL(*mockCompInstance, DoesModifiedMethodExist())
+    .Times(1)
+    .WillRepeatedly(ModifiedMethodExists());
+
+  // Deactivate and Unbindreference will also be called. EXPECT_CALL added
+  // to avoid GMOCK Warning. 
+   // When the mock Modified method is called it will throw the ModifiedMethodException
+  EXPECT_CALL(*mockCompInstance, Deactivate()).Times(1);
+  EXPECT_CALL(*mockCompInstance, UnbindReferences()).Times(1);
```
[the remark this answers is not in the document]
pelliott-mathworks: GMOCK_WARNING:
Uninteresting mock function call - returning directly. Function call Deactivate()
Note: You can safely ignore the above warning unless this call should not happen. Do not suppress it by blindly adding an EXPECT_CALL() if  you don't mean to enforce the call. See https://github.com/google/googletest/blob/master/googlemock/docs/cook_book.md#knowing-when-to-expect for details. 

The second warning looks just like that except the Function call is UnbindReferences. 

The warnings are coming up because Deactivate and UnbindReferences were called but we didn't have any EXPECT_CALL statements to indicate that we expected those to be called. It is valid that they are called during this test so I added the EXPECT_CALL statements to indicate that.

kevinleeMW (approved):

compendium/DeclarativeServices/test/TestSingletonComponentConfiguration.cpp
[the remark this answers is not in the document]
jeffdiclemente: Ok. Unless you want to test that Deactivate adn UnbindReferences are called in this testpoint, adding the EXPECT_CALL is unnecessary since, as the warning says: `You can safely ignore the above warning unless this call should not happen. Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.`. 

I'm fine with leaving the `EXPECT_CALL`s in however this warning is meant to flag the user about possible unenforced expectations on the mock. We have this warning in other testpoints. 

Another solution to get rid of the warning is outlined here: https://google.github.io/googletest/gmock_cook_book.html#NiceStrictNaggy.

PR #518, opened 2021-01-14
Fix ServiceTracker race

jeffdiclemente:
A race was observed involving closing a servicetracker while it is still processing a registered/unregistered service event.

This was observed as a couple different failure modes; either a crash would occur within:
```
bool ServiceReferenceBase::operator==(
  const ServiceReferenceBase& reference) const
{
  return d.load()->registration == reference.d.load()->registration;
}
````
whereby `this->d` was an invalid memory location or as an abort from [`VCRUNTIME140.dll!_purecall()`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/purecall?view=msvc-160) coming from `cppmicroservices::detail::BundleAbstractTracked<cppmicroservices::ServiceReference<T>,cppmicroservices::ServiceTrackerCustomizer<>::TypeTraits,cppmicroservices::ServiceEvent>::TrackAdding` which indicates that a virtual function call failed.

The testpoint `ServiceTrackerCloseRace` fails with a different failure mode.

### CODE REVIEW

**Review by pelliott-mathworks — APPROVED** [review]

**Comment by nonet-dev** [comment]
I think I did spot one problem with closing a ServiceTracker and registering a service concurrently, which the proposed change makes less likely to run into but does not fix.

In `ServiceTracker<S,T>::Open()` we register a service listener by binding a raw pointer of the `_TrackedService` shared_ptr member to its member function `_TrackedService::ServiceChanged`.

In `ServiceTracker<S,T>::Close()` we used to destroy the `_TrackedService` member when the function returns (with the code change in the pull request, this gets delayed to the point where the whole `ServiceTracker` is destroyed). This could destroy the object while its `_TrackedService::ServiceChanged` listener is being called concurrently.

I think instead of delaying the destruction of `_TrackedService` as proposed here, it is enough to bind the full shared_ptr to the `_TrackedService::ServiceChanged` listener when registering it, e.g. doing:

```
d->listenerToken = d->context.AddServiceListener(std::bind(&_TrackedService::ServiceChanged,
                                                                 t, std::placeholders::_1),
                                                       d->listenerFilter);
```

at ServiceTracker.tpp:107.

It might also be worthwhile to check other places in the codebase, if they bind raw pointers to callbacks and if this could pose a problem in concurrent scenarios.

**Comment by saschazelzer** [comment]
I think I did spot one problem with closing a ServiceTracker and registering a service concurrently, which the proposed change makes less likely to run into but does not fix.

In `ServiceTracker<S,T>::Open()` we register a service listener by binding a raw pointer of the `_TrackedService` shared_ptr member to its member function `_TrackedService::ServiceChanged`.

In `ServiceTracker<S,T>::Close()` we used to destroy the `_TrackedService` member when the function returns (with the code change in the pull request, this gets delayed to the point where the whole `ServiceTracker` is destroyed). This could destroy the object while its `_TrackedService::ServiceChanged` listener is being called concurrently.

I think instead of delaying the destruction of `_TrackedService` as proposed here, it is enough to bind the full shared_ptr to the `_TrackedService::ServiceChanged` listener when registering it, e.g. doing:

```
d->listenerToken = d->context.AddServiceListener(std::bind(&_TrackedService::ServiceChanged,
                                                                 t, std::placeholders::_1),
                                                       d->listenerFilter);
```

at ServiceTracker.tpp:107.

It might also be worthwhile to check other places in the codebase, if they bind raw pointers to callbacks and if this could pose a problem in concurrent scenarios.

**Comment by jeffdiclemente** [comment]
@saschazelzer good catch. We should be binding to the `shared_ptr`. I'll look for other places that need an adjustment. 

We need to guarantee that the `TrackedService` object is valid while `ServiceTracker` callbacks are executing (i.e. `AddingService`, `RemovedService`). Threads could be executing `AddingService/RemovedService` while a separate thread calls `ServiceTracker::Close`. Depending on where each thread is, the `TrackedService` would be invalidated for all other references once `std::shared_ptr<_TrackedService> outgoing = d->trackedService.Exchange(std::shared_ptr<_TrackedService>());` returned.

I don't think we can place locks in the code which calls into callbacks. We would have to release the lock while calling user code and then that would allow an opportunity for the `TrackerService` object to be destroyed. An alternative is to make the `ServiceTracker::Open` and `ServiceTracker::Close` methods thread safe and scope the `TrackedService` objects to the `ServiceTracker` objects lifetime. I think this allow any existing `TrackedService` objects to remain alive until callbacks have finished. This method does require a way to remove the `TrackedService` object's service references and not call the `RemovedService` callback.

**Review by pelliott-mathworks — APPROVED** [review]

File: framework/include/cppmicroservices/detail/ServiceTracker.tpp [hunk 1]
```diff
@@ -143,36 +143,26 @@ template<class S, class T>
...
+  } catch (const std::runtime_error& /*e*/) {
     /* In case the context was stopped or invalid. */
   }
 
   d->Modified(); /* clear the cache */
   outgoing->NotifyAll(); /* wake up any waiters */
-  for(auto& ref : references)
-  {
-    outgoing->Untrack(ref, ServiceEvent());
-  }
```
Burgch, line 169: Does this mean that `ServiceTrackerCustomizer`s will no longer get a call to `ServiceRemoved` for every tracked service when the `ServiceTracker` is closed?

PR #519, opened 2021-01-22
Test Lazy Loading of a Delayed Component

apatil19:
Updated to test lazy loading of a delayed component.

Validations:
- Validate the bundle is not loaded into the process before GetService is called.
- Validate the bundle is loaded into the process once GetService is called.


compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -190,7 +203,53 @@ TEST_F(tServiceComponent, testDelayedComponent_LifeCycle) //DS_TOI_52
...
+    HANDLE hProcess = GetCurrentProcess();
+    auto res = EnumProcessModules(hProcess, hMods, sizeof(hMods), &cbNeeded);
+    EXPECT_EQ(res, 1)<<"Enumeration failed";
+
+    char szModName[MAX_PATH];
+    std::string result;
+
+    for (i = 0; i < (cbNeeded / sizeof(HMODULE)); i++) 
+    {
+      GetModuleFileNameEx(hProcess, hMods[i], szModName, sizeof(szModName) / sizeof(TCHAR));
```
pelliott-mathworks, line 222: I think GetModuleFileNameEx is meant to be used to get module information for a different process than the current process. GetModuleFileName is more reliable when getting information from the current process. In either case these functions can fail and will return an error. Should probably be checked.

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -226,6 +285,42 @@ TEST_F(tServiceComponent, testDelayedComponent_LifeCycle) //DS_TOI_52
   compConfigDTOs = dsRuntimeService->GetComponentConfigurationDTOs(compDescDTO);
   EXPECT_EQ(compConfigDTOs.at(0).state, scr::dto::ComponentState::ACTIVE) << "State must be ACTIVE after call to GetService";
   EXPECT_NO_THROW(service->ExtendedDescription()) << "Throws if the dependency could not be found";
+  
+  #if defined(_WIN32)
+    hProcess = GetCurrentProcess();
+    res = EnumProcessModules(hProcess, hMods, sizeof(hMods), &cbNeeded);
```
pelliott-mathworks, line 291: Same comment on EnumProcessModules and GetModuleFileNameEx as above

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -226,6 +285,42 @@ TEST_F(tServiceComponent, testDelayedComponent_LifeCycle) //DS_TOI_52
   compConfigDTOs = dsRuntimeService->GetComponentConfigurationDTOs(compDescDTO);
   EXPECT_EQ(compConfigDTOs.at(0).state, scr::dto::ComponentState::ACTIVE) << "State must be ACTIVE after call to GetService";
   EXPECT_NO_THROW(service->ExtendedDescription()) << "Throws if the dependency could not be found";
+
```
pelliott-mathworks, line 288: Line 280 tests to see if the component state is SATISFIED. For lazy loading a component should not be loaded at this point. Should there be another test like the one above to make sure the module is not loaded?

pelliott-mathworks: Also, there's some duplicate code where the modules are retrieved and checked  and if you add one more test here that will increase the amount of duplicate code. Consider adding a function?

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
apatil19: Replaced GetModuleFileNameEx with GetModuleFileNameA

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
apatil19: Added another test to make sure that the bundle is not loaded even if the component state is SATISFIED.

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -175,13 +187,82 @@ TEST_F(tServiceComponent, testImmediateComponent_LifeCycle_Dynamic) // DS_TOI_51
...
+    std::string result;
+
+    for (i = 0; i < (cbNeeded / sizeof(HMODULE)); i++)
+    {
+        auto file = GetModuleFileNameA(hMods[i], szModName, sizeof(szModName) / sizeof(TCHAR));
+        EXPECT_NE(file, 0) << "Failed to retrive file path";
+        result += szModName;
+    }
+
+    std::size_t found = result.find("TestBundleDSTOI6");
```
pelliott-mathworks, line 214: some duplicate code here. Also appears in the else. Might be worth pulling it into a function?

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -175,13 +187,82 @@ TEST_F(tServiceComponent, testImmediateComponent_LifeCycle_Dynamic) // DS_TOI_51
...
+    }
+
+    CloseHandle(hProcess);
+    result.clear();
+
+  #else
+    auto pid_t = getpid();
+    std::string command("lsof -p " + std::to_string(pid_t));
+    FILE* fd = popen(command.c_str(), "r");
+    EXPECT_NE(fd, nullptr) << "popen failed";
```
pelliott-mathworks, line 231: Should this be an assert? Shouldn't really continue in this code if the popen failed?

pelliott-mathworks (approved):

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
apatil19: Yes. Changed it to Assert.

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -175,13 +187,65 @@ TEST_F(tServiceComponent, testImmediateComponent_LifeCycle_Dynamic) // DS_TOI_51
...
+{
+  #if defined(_WIN32)
+
+    HMODULE hMods[1024];
+    DWORD cbNeeded;
+    unsigned int i;
+
+    HANDLE hProcess = GetCurrentProcess();
+    auto res = EnumProcessModules(hProcess, hMods, sizeof(hMods), &cbNeeded);
+    EXPECT_GT(sizeof(hMods), cbNeeded) << "Size of array is too small to hold all module handles";
```
jeffdiclemente, line 201: `EnumProcessModules` returns the number of bytes required to store all module handles in `cbNeeded`. There shouldn't be a need to test for the size. Instead if the size is too small, call `EnumProcessModules` with a sufficiently large enough buffer.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -175,13 +187,82 @@ TEST_F(tServiceComponent, testImmediateComponent_LifeCycle_Dynamic) // DS_TOI_51
...
+        return false;
+    }
+
+    TCHAR szModName[MAX_PATH];
+    std::string result;
+
+    for (unsigned int i = 0; i < (cbNeeded / sizeof(HMODULE)); i++)
+    {
+        auto file = GetModuleFileNameA(hMods[i], szModName, sizeof(szModName) / sizeof(TCHAR));
+        EXPECT_NE(file, 0) << "Failed to retrive file path";
```
jeffdiclemente, line 214: nit-pick: `file` is not a great name for a variable. What is returned from `GetModuleFileNameA` is not a file, its the length of the string that is copied to the buffer, in characters.

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -175,13 +187,82 @@ TEST_F(tServiceComponent, testImmediateComponent_LifeCycle_Dynamic) // DS_TOI_51
...
+        result += szModName;
```
jeffdiclemente, line 215: Its not necessary to loop over every module name; once the module name is found, the work is done.

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -175,13 +187,82 @@ TEST_F(tServiceComponent, testImmediateComponent_LifeCycle_Dynamic) // DS_TOI_51
...
+  #if defined(US_PLATFORM_WINDOWS)
+
+    HMODULE hMods[1024];
+    DWORD cbNeeded;
+
+    HANDLE hProcess = GetCurrentProcess();
+
+    auto res = EnumProcessModules(hProcess, hMods, sizeof(hMods), &cbNeeded);
+    EXPECT_NE(res, 0) << "Enumeration failed";
+    EXPECT_GT(sizeof(hMods), cbNeeded) << "Size of array is too small to hold all module handles";
```
jeffdiclemente, line 202: you can condense these two `EXPECT_` macros into one that uses `[GetLastError](https://docs.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror)` to provide detailed error information.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -252,4 +278,101 @@ TEST_F(tServiceComponent, testDelayedComponent_LifeCycle) //DS_TOI_52
...
+    HANDLE hProcess = GetCurrentProcess();
+
+    EnumProcessModules(hProcess, hMods, sizeof(hMods), &cbNeeded);
+    EXPECT_FALSE(isErrored("EnumProcessModules"));
+
+    EXPECT_GT(sizeof(hMods), cbNeeded) << "Size of array is too small to hold all module handles";
+    if ((sizeof(hMods) < cbNeeded))
+    {
+        return false;
+    }
```
jeffdiclemente, line 331: This is unnecessary if you are checking the return code from `EnumProcessModules`. If `EnumProcessModules` fails there is no in doing anything else.

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -252,4 +278,101 @@ TEST_F(tServiceComponent, testDelayedComponent_LifeCycle) //DS_TOI_52
...
+    std::cerr << "\nERROR:\n" << functionName << " failed with error " << dw << ": " << lpDisplayBuf << std::endl;
+
+    SetLastError(0);
+    LocalFree(lpDisplayBuf);
+    return true;
+#else
+    std::cerr<<"isErrored("<<functionName<<") will only work for windows";
+    return true;
+#endif
+}
```
jeffdiclemente, line 312: There is already a function which does this in the `util` project, which this executable links against. Look at `Error.cpp`. no need to re-invent the wheel.

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -252,4 +278,101 @@ TEST_F(tServiceComponent, testDelayedComponent_LifeCycle) //DS_TOI_52
...
+        return false;
+    }
+
+    TCHAR szModName[MAX_PATH];
+    std::size_t found;
+
+    for (unsigned int i = 0; i < (cbNeeded / sizeof(HMODULE)); i++)
+    {
+        GetModuleFileNameA(hMods[i], szModName, sizeof(szModName) / sizeof(TCHAR));
+        EXPECT_FALSE(isErrored("GetModuleFileNameA"));
```
jeffdiclemente, line 339: I don't see the point of having `EXPECT_FALSE` here. returning false from this function means the module couldn't be found.  is it worth it to run this code at most `(cbNeeded / sizeof(HMODULE)) -1` times? seems unnecessary.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/TestComponentLifecycle.cpp
```diff
@@ -252,4 +278,75 @@ TEST_F(tServiceComponent, testDelayedComponent_LifeCycle) //DS_TOI_52
   auto sRef2 = ctxt.GetServiceReference<test::Interface2>();
   EXPECT_FALSE(static_cast<bool>(sRef2)) << "Service must not be available after it's dependency is removed";
 }
+
+//Function to validate lazy loading of delayed component
+bool isBundleLoaded(const std::string bundleName)
```
jeffdiclemente, line 283: `bundleName` should be passed in as `const std::string&` or `std::string`, if you want to allow move semantics.

this entire function has general use and could be used by other tests. I suggest moving it into the TestUtils.cpp file and renaming the function to a more descriptive name, something like `isBundleLoadedInThisProcess`

jeffdiclemente (changes requested):

jeffdiclemente (approved):
@apatil19 you'll need to resolve the merge conflicts before I can merge. Thanks.

PR #521, opened 2021-02-04
Updated API to move manifest into the BundleRegistry

carneyweb:
Did I get the details of moving correct?


pelliott-mathworks (approved):
Looks great!

PR #526, opened 2021-02-24
changed "auto headers = " to "auto const& headers = "

carneyweb:
The original version caused an unnecessary copy of the bundle's headers. We only read them, and
copying is time consuming.


jeffdiclemente (approved):

kevinleeMW (approved):

pelliott-mathworks (commented):
Looks good!

PR #527, opened 2021-02-26
Made ToJSON for strings do proper escaping of special characters.

carneyweb:
While working on something that required converting an AnyMap to JSON I found that our implementation didn't properly escape special characters in strings when doing the conversion. I updated the test case with an example of the problem that I found, showed the failure, and then updated the code to show a pass.

### DIFF
--- a/framework/src/util/Any.cpp
+++ b/framework/src/util/Any.cpp
@@ -24,6 +24,7 @@
 #include "Utils.h"
 
 #include <stdexcept>
+#include <iomanip>
 
 namespace cppmicroservices {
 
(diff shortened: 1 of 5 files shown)


framework/src/util/Any.cpp
```diff
@@ -42,26 +43,58 @@ void ThrowBadAnyCastException(const std::string& funcName,
...
+#if NEVER
+  // The original code for this function fails to properly escape the characters in the string to
+  // make it a proper JSON string. The new code below does so.
   return os << '"' << val << '"';
+#endif
+  o << '"';
+  for (auto c = s.cbegin(); c != s.cend(); c++) {
+    switch (*c) {
+      case '"' : o << "\\\""; break;
+      case '\\': o << "\\\\"; break;
```
pelliott-mathworks, line 77: cppreference says that the following are also escape sequences. \\? \\a \\v. Should we support all of them for completeness? Of those 3 that are missing the \\? is the only one that wouldn't be picked up by your default case. It's hex value is 0x3F.

pelliott-mathworks (approved):

framework/src/util/Any.cpp
carneyweb: Hmm... I'm not sure. Although those characters can be escaped in c++, I don't see them as needing special treatment for JSON strings. Here's what needs to happen with them:

https://www.crockford.com/mckeeman.html

jeffdiclemente: since \?, \a, and \v are not in the JSON grammar as needing escaping we shouldn't need to handle them.

jeffdiclemente (approved):

PR #531, opened 2021-03-17
Fix issues with line endings in BundleResourceTest.cpp test

achristoforides:
This change introduces changes to the BundleResourceTest.cpp file which addresses issues with the test dependent upon what git settings a developer has for automatically converting line endings in files.

A .gitattributes file is introduced to the codebase which forces certain file extensions (.ptxt) in this PR to always use LF line endings instead of being different depending on what machine you are using.

The test was updated to reflect these changes.


framework/test/driver/BundleResourceTest.cpp
```diff
@@ -365,24 +364,23 @@ void testResourceTree(const Bundle& bundle)
   // find all .txt files
   std::vector<BundleResource> nodes = bundle.FindResources("", "*.txt", false);
   std::sort(nodes.begin(), nodes.end(), resourceComparator);
-  US_TEST_CONDITION_REQUIRED(nodes.size() == 3, "Found child count")
-  US_TEST_CONDITION(nodes[0].GetResourcePath() == "/foo.txt",
-                    "Check child name")
-  US_TEST_CONDITION(nodes[1].GetResourcePath() == "/foo2.txt",
-                    "Check child name")
-  US_TEST_CONDITION(nodes[2].GetResourcePath() == "/special_chars.dummy.txt",
-                    "Check child name")
```
jeffdiclemente, line 374: why was this removed? shouldn't `std::vector<BundleResource> nodes = bundle.FindResources("", "*.txt", false);` be modified to look for all `ptxt` files instead?

achristoforides: Not all the files in the test bundle were changed to ".ptxt". For example, README.txt still has a .txt extension since the contents of the file are never checked against; it is only used to verify that a recursive search for resources within a bundle works.

achristoforides: The code below this section preserves the check for .txt resources and it still checks recursively as well, thus maintaining the same level of coverage and testing that the previous version had.

jeffdiclemente: I see now. that makes sense.

jeffdiclemente (approved):

PR #533, opened 2021-03-22
Remove Bundle Threads

achristoforides:
This PR removes the BundleThread functionality from CppMicroServices.

BundleThread.h, BundleThread.cpp, and ConcurrencyTest.cpp are removed from the repo.


framework/test/driver/ConcurrencyTest.cpp
```diff
@@ -20,28 +20,29 @@ limitations under the License.
```
jeffdiclemente, line 20: This whole file can be deleted if it was only used to test bundlethreads.

jeffdiclemente (changes requested):

framework/test/driver/BundleTest.cpp
```diff
@@ -1031,116 +1022,6 @@ void TestIllegalBundleStateChange()
   f.WaitForStop(std::chrono::seconds::zero());
 }
 
-// test failure on waiting for bundle operations.
-void TestWaitOnBundleOperation()
```
jeffdiclemente, line 1035: delete the TestBundleWaitOnOperation bundle since it no longer has a purpose once this test is removed.

jeffdiclemente (changes requested):

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -64,16 +65,13 @@ void BundlePrivate::Stop(uint32_t options)
...
 
     // 1:
     if (state == Bundle::STATE_UNINSTALLED) {
       throw std::logic_error("Bundle is uninstalled");
     }
 
-    // 2: If an operation is in progress, wait a little
-    WaitOnOperation(coreCtx->resolver, l, "Bundle::Stop", false);
-
     // 3:
```
kevinleeMW, line 75: update comments so we don't skip // 2

pelliott-mathworks (approved):

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -64,15 +64,15 @@ void BundlePrivate::Stop(uint32_t options)
...
+    auto l = this->Lock();
 
     // 1:
     if (state == Bundle::STATE_UNINSTALLED) {
       throw std::logic_error("Bundle is uninstalled");
     }
 
     // 2: If an operation is in progress, wait a little
-    WaitOnOperation(coreCtx->resolver, l, "Bundle::Stop", false);
+    // Note: This step hass been removed with removal of BundleThread
```
jeffdiclemente, line 75: if BundleThreads is being removed, it doesn't make sense to add comments that reference it. I'd suggest updating the comments as if BundleThreads doesn't exist.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -251,7 +205,8 @@ Bundle::State BundlePrivate::GetUpdatedState(LockType& l)
   return static_cast<Bundle::State>(state.load());
 }
 
-void BundlePrivate::SetStateInstalled(bool sendEvent, UniqueLock& resolveLock)
+void BundlePrivate::SetStateInstalled(bool sendEvent,
```
jeffdiclemente, line 208: I suggest removing the lock parameter completely from all the functions that used to need it because of BundleThreads.

jeffdiclemente (changes requested):

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -65,14 +65,14 @@ void BundlePrivate::Stop(uint32_t options)
...
 
     // 1:
     if (state == Bundle::STATE_UNINSTALLED) {
       throw std::logic_error("Bundle is uninstalled");
     }
 
     // 2: If an operation is in progress, wait a little
-    // Note: This step hass been removed with removal of BundleThread
 
     // 3:
```
jeffdiclemente: you still want to remove this comment then re-number the remainder of the comments.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #537, opened 2021-03-25
Fix #301 so TSAN builds work on Linux

### LINKED ISSUE #301: ResourceCompiler crashes on address-sanitized builds.
Hi guys!

I have ran into an issue with the ResourceCompiler on address-sanitized builds.
It crashes when trying to access the JSON maniffest in an archive (in validateManifestInArchive()).
The call to mz_zip_reader_extract_file_to_heap() does not include a null terminator for the string in
its allocated buffer, although the memory beyond seems zeroed-out. In address-sanitized builds
this space is guarded, causing the program to abort during construction of the istringstream a few
lines below.

I have included a patch that fixes this issue.

```
diff --git a/tools/rc/ResourceCompiler.cpp b/tools/rc/ResourceCompiler.cpp
index ef15376..d2fecdc 100644
--- a/tools/rc/ResourceCompiler.cpp
+++ b/tools/rc/ResourceCompiler.cpp
@@ -179,8 +179,9 @@ void validateManifestInArchive(mz_zip_archive* zipArchive,
                                const std::string& archiveFile,
                                const std::string& archiveEntry)
 {
+  size_t length = 0;
   void* data = mz_zip_reader_extract_file_to_heap(
-    zipArchive, archiveEntry.c_str(), nullptr, 0);
+    zipArchive, archiveEntry.c_str(), &length, 0);
   std::unique_ptr<void, void (*)(void*)> manifestFileContents(data, ::free);
 
   if (!manifestFileContents) {
@@ -191,7 +192,7 @@ void validateManifestInArchive(mz_zip_archive* zipArchive,
   try {
     Json::Value root;
     std::istringstream json(
-      reinterpret_cast<const char*>(manifestFileContents.get()));
+      std::string(reinterpret_cast<const char*>(manifestFileContents.get()), length));
     parseAndValidateJson(json, root);
   } catch (const InvalidManifest& e) {
     std::string exceptionMsg(archiveFile);
```

achristoforides:
Fixes #301 .

This PR makes changes to ResourceCompiler.cpp to remove the ASAN and TSAN errors that occur when building.

Testing was done locally on Linux with US_ENABLE_TSAN turned on since the build would always fail previously if this was turned on.

### DIFF
--- a/tools/rc/ResourceCompiler.cpp
+++ b/tools/rc/ResourceCompiler.cpp
@@ -179,8 +179,9 @@ void validateManifestInArchive(mz_zip_archive* zipArchive,
                                const std::string& archiveFile,
                                const std::string& archiveEntry)
 {
+  size_t length = 0;
   void* data = mz_zip_reader_extract_file_to_heap(
-    zipArchive, archiveEntry.c_str(), nullptr, 0);
+    zipArchive, archiveEntry.c_str(), &length, 0);
   std::unique_ptr<void, void (*)(void*)> manifestFileContents(data, ::free);
 
   if (!manifestFileContents) {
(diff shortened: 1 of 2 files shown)


jeffdiclemente:
the suppression in tsan_suppressions.txt can also be removed now.

jeffdiclemente (approved):

PR #540, opened 2021-03-31
Anymap erase and compare

carneyweb:
Added implementations of AnyMap::erase(), AnyMap::operator==(), and Any::operator==(). The implementation of Any::operator==() behaves slightly differently than the standard interface: If one of the operator==() operands contains a value whose type does not have an operator==() implementation available, the comparison fails instead of failing to compile. This is because of the implementation of Any::Holder::operator==() which fails to compile if ValueType does not support operator==(), which would disallow storing anything in an Any container that does not provide the equality operator.

### DIFF
--- a/framework/include/cppmicroservices/Any.h
+++ b/framework/include/cppmicroservices/Any.h
@@ -56,7 +56,7 @@ template <class T>
 struct has_op_eq
 {
   template <class U>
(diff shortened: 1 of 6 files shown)


pelliott-mathworks (approved):
Well crafted!

achristoforides (approved):

PR #544, opened 2021-04-06
Migrate tests to gtest

apatil19:

- FrameworkEventTest 
- BundleResourceTest


framework/test/gtest/FrameworkFactoryTest.cpp
```diff
@@ -27,7 +27,7 @@ limitations under the License.
 
 using namespace cppmicroservices;
 
-TEST(FrameworkFactoryTest, FrameworkInstantiation)
+TEST(FrameworkFactoryTest, testFrameworkInstantiation)
```
carneyweb, line 30: Just curious... why change the name?

carneyweb (approved):
Ship it!

framework/test/gtest/FrameworkFactoryTest.cpp
apatil19: We have test points starting with 'Test', 'test' or just the name.
I am thinking, may be, we can start all the test-point names with 'test' and make it a standard.

jeffdiclemente (approved):

PR #546, opened 2021-04-12
Enable ASAN for local testing

achristoforides:
This PR introduces changes which enables ASAN when building locally on Windows, Mac, and Linux.

Mac and Linux only require the developer to turn on US_ENABLE_ASAN while Windows additionally requires the developer to specify the path to the runtime DLL for ASAN provided by Microsoft (typically "clang_rt.asan_dbg_dynamic-x86_x64.dll" or "clang_rt.asan_dynamic-x86_x64.dll").


pelliott-mathworks (approved):

jeffdiclemente (changes requested):
Please update the build documentation in doc/src/build_instructions.rst to reflect the TSan and Asan build options. If there is anything special one needs to do on Windows, this would be the place to document it. Thanks.

CMakeLists.txt
```diff
@@ -220,6 +220,8 @@ us_cache_var(CMAKE_DEBUG_POSTFIX d STRING "Executable and library debug name pos
 
 us_cache_var(US_ENABLE_THREADING_SUPPORT ON BOOL "Enable threading support")
 us_cache_var(US_ENABLE_TSAN OFF BOOL "Enable tsan (thread sanitizer)" ADVANCED)
+us_cache_var(US_ENABLE_ASAN OFF BOOL "Enable asan (address sanitizer)" ADVANCED)
+us_cache_var(US_ASAN_USER_DLL "" STRING "Path to asan dll" ADVANCED)
```
jeffdiclemente, line 224: if this variable is only for ASan support on Windows, I suggest making that explicit in the variable description. So something like "Path to Windows Asan DLL"

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #548, opened 2021-04-16
Add from json

carneyweb:
Added new Any::FromJSON static method to create an Any object from a JSON string using.

### DIFF
--- a/framework/include/cppmicroservices/Any.h
+++ b/framework/include/cppmicroservices/Any.h
@@ -424,13 +424,10 @@ public:
    *
    * @throws std::runtime_error with invalid JSON input
    * @param  json a string containing JSON Data
-   * @param  use_ci_map_keys a bool which indicates whether or not the keys in any object maps should
-   *         be case insensitive. We use case insensitive keys in CppMicroServices, but there are
-   *         other uses of this static method that may not need them to be case insensitive.
    * @return an Any object containing the parsed data. 
    */
-  static Any FromJSON(std::string const& json, bool use_ci_map_keys = true);
-  static Any FromJSON(std::istream& json, bool use_ci_map_keys = true);
+  static Any FromJSON(const std::string& json);
+  static Any FromJSON(std::istream& json);
 private:
   class Placeholder
   {
--- a/framework/include/cppmicroservices/Any.h
+++ b/framework/include/cppmicroservices/Any.h
@@ -365,6 +365,8 @@ public:
     return _content ? _content->Type() : typeid(void);
   }
 
+  static Any FromJSON(std::string const& json);
+  static Any FromJSON(std::istream& json);
 private:
   class Placeholder
   {
--- a/framework/include/cppmicroservices/Any.h
+++ b/framework/include/cppmicroservices/Any.h
@@ -418,8 +418,19 @@ public:
     return _content ? _content->Type() : typeid(void);
   }
 
-  static Any FromJSON(std::string const& json);
-  static Any FromJSON(std::istream& json);
+  /**
+   * Parse a JSON string and return the results in an Any object. Hierarchical data is stored in
+   * maps, and lists are stored in vectors.
+   *
+   * @throws std::runtime_error with invalid JSON input
+   * @param  json a string containing JSON Data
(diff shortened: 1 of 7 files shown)


pelliott-mathworks (approved):

jeffdiclemente:
there should be an RFC posted to https://github.com/CppMicroServices/rfcs since this appears to add new public APIs.

framework/include/cppmicroservices/Any.h
```diff
@@ -418,6 +418,19 @@ class US_Framework_EXPORT Any
...
+   *
+   * @throws std::runtime_error with invalid JSON input
+   * @param  json a string containing JSON Data
+   * @param  use_ci_map_keys a bool which indicates whether or not the keys in any object maps should
+   *         be case insensitive. We use case insensitive keys in CppMicroServices, but there are
+   *         other uses of this static method that may not need them to be case insensitive.
+   * @return an Any object containing the parsed data. 
+   */
+  static Any FromJSON(std::string const& json, bool use_ci_map_keys = true);
+  static Any FromJSON(std::istream& json, bool use_ci_map_keys = true);
```
jeffdiclemente, line 433: as discussed offline; remove the optional `use_ci_map_keys` and default to case insensitive unordered map. if users need case sensitive and/or ordered maps, that can be exposed later.

framework/test/gtest/AnyTest.cpp
```diff
@@ -297,3 +297,42 @@ TEST(AnyTest, AnyEquality) {
   EXPECT_NE(lhs, rhs); // and finally, with the "int" element erased, they should not be equal
                        // anymore. 
 }
+TEST(AnyTest, FromJSON) {
```
jeffdiclemente, line 300: need a test to lock down the case insensitive behavior of `FromJSON`

jeffdiclemente (approved):

kevinleeMW (approved):

jeffdiclemente (approved):

framework/src/util/AnyMap.cpp
```diff
@@ -1027,7 +1027,48 @@ bool any_map::operator==(const any_map& rhs) const
...
+            return false;
+          }
+        }
+        for (auto const& i_rhs : *rhs.map.uoci) {
+          // now make sure that the keys from the rhs have data in the lhs also. If the key is found
+          // in the lhs, we've already compared the keys and values themselves so we don't need to
+          // do so again.
+          auto const& i_lhs = map.uoci->find(i_rhs.first);
+          if (i_lhs == std::end(*map.uoci)) {
+            return false;
```
achristoforides, line 1067: Switch to using west-coast const for consistency

achristoforides (changes requested):

framework/src/util/AnyMap.cpp
```diff
@@ -1027,7 +1027,48 @@ bool any_map::operator==(const any_map& rhs) const
...
+      {
+        // In the case of our case insensitive keys map, there's an "oversight" in the
+        // std::unordered_map implementation such that if the 4th template argument is passed in
+        // (that is, the key-equality argument), it is only used when doing lookups in the map, not
+        // when std::unordered_map::operator== is used to check for equality between two
+        // maps. Inside the operator==, keys are compared with their own operator== rather than
+        // using the key equality template argument. In our case, this means that if I have a "key"
+        // in map1, and a "KEY" in map2, the maps will compare as inequal even though we have
+        // specified a case-insensitive comparator. So, we have to roll our own to make sure we get
+        // case-insensitive key comparisons.
```
achristoforides, line 1039: I'm not sure if this is an "oversight". Map equality, and specifically std::unordered_map equality, between two std::unordered_maps is solely dependent on whether or not the key-value pairs in one map are identical to another (albeit allowing for various permutations since the map is unordered).

The comparator we specify is used when finding keys and their associated values. Even in regular maps, it is my understanding that the comparator isn't used when checking equality; ordered maps similarly checks for equality of key-value pairs.

jeffdiclemente:
inactive pull request. closing for now. please submit a new pull request for this functionality.

PR #557, opened 2021-04-29
Doc updates

jeffdiclemente:
There are three parts to this pull request:

1. Making the threading guarantees for ServiceTracker and ServiceTrackerCustomizer more explicit/visible
2. Adding DS, ConfigAdmin and LogService API docs
3. Fixing the `ComponentState` enum to be a `class enum`. This subsequently fixed a doxygen issue, which is why this change was made.


compendium/DeclarativeServices/src/manager/states/CCSatisfiedState.cpp
```diff
@@ -24,7 +24,7 @@
 #include "CCUnsatisfiedReferenceState.hpp"
 #include "../ComponentConfigurationImpl.hpp"
 
-using cppmicroservices::service::component::runtime::dto::UNSATISFIED_REFERENCE;
+using cppmicroservices::service::component::runtime::dto::ComponentState;
```
pelliott-mathworks, line 27: Don't need this line since the one place that uses ComponentState is fully qualified?

pelliott-mathworks (approved):
After these changes are merged into development we should pull them into the configadmin_ds_integration branch (rebase)? Also we have created a bunch of new files in the configadmin_ds_integration branch so we should probably make similar changes in those new files?

jeffdiclemente:
> After these changes are merged into development we should pull them into the configadmin_ds_integration branch (rebase)? Also we have created a bunch of new files in the configadmin_ds_integration branch so we should probably make similar changes in those new files?

yes and yes. only the public facing APIs need to be documented, if they aren't already.

achristoforides (approved):

PR #558, opened 2021-04-30
Fix ServiceTracker race

jeffdiclemente:
The race occurs when a customizer is still running and the ServiceTracker object is in the process of being destroyed.

The fix is to have `ServiceTracker::Close` unregister its service listener and then wait for any existing service events to finish processing before cleaning up and calling untrack.


framework/test/util/ConcurrencyTestUtil.hpp
```diff
@@ -0,0 +1,64 @@
...
+  std::shared_future<void> ready(go.get_future());
+  size_t numCalls = std::max(64u, std::thread::hardware_concurrency());
+  std::vector<std::promise<void>> readies(numCalls);
+
+  // TODO: use std::barrier when available
+  std::vector<std::future<R>> enable_async_futs(numCalls);
+  for (size_t i = 0; i < numCalls; ++i) {
+    enable_async_futs[i] =
+      std::async(std::launch::async, [&readies, i, &ready, &func]() {
+        readies[i].set_value();
```
pelliott-mathworks, line 42: func() is executed before the readies[i].set_value call. I think it should be executed after the func() call. Otherwise the readies[i].get_future().wait() in the next for loop could potentially be satisfied and move on before the last function has finished executing.

pelliott-mathworks (changes requested):

framework/include/cppmicroservices/detail/CounterLatch.h
```diff
@@ -0,0 +1,133 @@
...
+#ifndef CPPMICROSERVICES_DETAIL_COUNTERLATCH_H
+#  define CPPMICROSERVICES_DETAIL_COUNTERLATCH_H
+
+#include <mutex>
+
+namespace cppmicroservices {
+
+namespace detail {
+
+/**
```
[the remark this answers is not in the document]
jeffdiclemente: I decided against doing this since `CounterLatch` is not supposed to be used by anyone other than cppms framework developers. I don't want to advertise this class anymore than it currently is.

framework/test/util/ConcurrencyTestUtil.hpp
jeffdiclemente: This is functioning as designed from what I can tell. The `readies` vector is meant to synchronize all the threads to right before `func()` will be called. Then once `go.set_value()` executes, all the threads race to execute `func()`. `ready.wait()` stops `func()` from being executed prematurely.

achristoforides (approved):

framework/include/cppmicroservices/detail/TrackedService.tpp
```diff
@@ -53,6 +61,15 @@ void TrackedService<S,TTT>::ServiceChanged(const ServiceEvent& event)
...
 
+  (void)latch.CountUp();
+  ScopeGuard sg([this]() {
+    // By using try/catch here, we ensure that this lambda function doesn't
+    // throw inside ScopeGuard's dtor.
+    try {
+      latch.CountDown();
+    } catch (...) { }
+  });
+
```
jeffdiclemente, line 72: there's still a race here; a thread can enter `void TrackedService<S,TTT>::ServiceChanged(const ServiceEvent& event)` after `void ServiceTracker<S,T>::Close()` has finished waiting on all customizers to finish.

pelliott-mathworks (approved):

carneyweb (approved):

PR #562, opened 2021-05-05
Fix TestConcurrentBindUnbind error

pelliott-mathworks:
An exception was being thrown when running the TestConcurrentBindUnbind method because DS was in the middle of binding to a service reference while another thread was unregistering the service. The sObjs.GetService call in ComponentContextImpl::AddToBoundServicesCache was throwing the exception because the ServiceReferenceBasePrivate object had been cleared out by the unregister operation and was returning an empty InterfaceMap. Changed ServiceObjects.cpp to check for the return of an empty map and handle the condition. Signed-off-by: The MathWorks, Inc. <pelliott@mathworks.com>

### DIFF
--- a/compendium/DeclarativeServices/src/ComponentContextImpl.cpp
+++ b/compendium/DeclarativeServices/src/ComponentContextImpl.cpp
@@ -82,10 +82,14 @@ void ComponentContextImpl::InitializeServicesCache()
                 .ToStringNoExcept();
             cppmicroservices::ServiceObjects<void> sObjs =
               bc.GetServiceObjects(sRefU);
-            serviceMap.push_back(sObjs.GetService());
+            auto interfaceMap = sObjs.GetService();
+            if (interfaceMap) {
+                serviceMap.push_back(interfaceMap);
+            }
           }
         }
       });
+ 
   }
 }
 
@@ -208,7 +212,7 @@ void ComponentContextImpl::Invalidate()
   boundServicesCacheHandle->clear();
 }
 
-void ComponentContextImpl::AddToBoundServicesCache(
+bool ComponentContextImpl::AddToBoundServicesCache(
   const std::string& refName,
   const cppmicroservices::ServiceReferenceBase& sRef)
 {
@@ -216,11 +220,16 @@ void ComponentContextImpl::AddToBoundServicesCache(
   cppmicroservices::ServiceObjects<void> sObjs =
     bc.GetServiceObjects(ServiceReferenceU(sRef));
   auto boundServicesCacheHandle = boundServicesCache.lock();
+  auto interfaceMap = sObjs.GetService();
+  if (!interfaceMap) {
+      return false;
+  }
   (*boundServicesCacheHandle)[refName].emplace_back(
-    sObjs.GetService());
+    interfaceMap);
+  return true;
 }
 
-void ComponentContextImpl::RemoveFromBoundServicesCache(
+bool ComponentContextImpl::RemoveFromBoundServicesCache(
   const std::string& refName,
   const cppmicroservices::ServiceReferenceBase& sRef)
 {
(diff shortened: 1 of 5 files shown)


compendium/DeclarativeServices/src/manager/BundleOrPrototypeComponentConfiguration.cpp
```diff
@@ -155,7 +159,10 @@ void BundleOrPrototypeComponentConfigurationImpl::UnbindReference(const std::str
                        std::current_exception());
     }
     
-    context->RemoveFromBoundServicesCache(refName, ref);
+    if (!context->RemoveFromBoundServicesCache(refName, ref)) {
+        GetLogger()->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
+            "Failure when removing a reference from the BoundServices Cache");
+    }
```
kevinleeMW, line 165: Perhaps I'll add a "return" here as well, in case any code is added after the if-statement in the future.

compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.cpp
```diff
@@ -152,7 +156,11 @@ void SingletonComponentConfigurationImpl::UnbindReference(const std::string& ref
                      std::current_exception());
   }
   auto context = GetComponentContext();
-  context->RemoveFromBoundServicesCache(refName, ref);
+  if (!context->RemoveFromBoundServicesCache(refName, ref)) {
+      GetLogger()->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
+          "Failure when removing a reference from the BoundServices Cache");
+  }
```
kevinleeMW, line 162: Perhaps I'll add a "return" here as well, in case any code is added after the if-statement in the future.

kevinleeMW (approved):

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -229,6 +238,9 @@ void ComponentContextImpl::RemoveFromBoundServicesCache(
     bc.GetServiceObjects(ServiceReferenceU(sRef));
 
   const auto removedService = sObjs.GetService();
+  if (!removedService) {
+      return false;
+  }
```
jeffdiclemente, line 243: This will leave a number of stale bound services in the boundServicesCache. How will this affect DS?

pelliott-mathworks: I realized that removedService not having a valid value can't happen here. The BoundServicesCache holds a copy of the InterfaceMapConstPtr which is a shared_ptr<InterfaceMap>. While we're holding this shared_ptr in our BoundServicesCache, the ServiceRegistrationBasePrivate object can't be destroyed. So in this case the sObjs.GetService will always return a value. I removed this test.

compendium/DeclarativeServices/src/manager/BundleOrPrototypeComponentConfiguration.cpp
```diff
@@ -146,23 +146,21 @@ void BundleOrPrototypeComponentConfigurationImpl::UnbindReference(const std::str
...
+          instance->InvokeUnbindMethod(refName, ref);
+      }
+      catch (const std::exception&) {
+          GetLogger()->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
+              "Exception received from user code while unbinding a "
+              "service reference.",
+              std::current_exception());
+      }
+
+      context->RemoveFromBoundServicesCache(refName, ref);
```
jeffdiclemente: it looks like the code formatting changed. indentations should be 2 spaces

jeffdiclemente (approved):
There are formatting issues and conflicts to resolve but other than that, looks good.

jeffdiclemente (approved):

PR #566, opened 2021-05-11
Format all code files using clang-format

achristoforides:
This PR has no functional or code changes. This PR merely formats all existing C, CPP, CXX, H, and HPP files in the repo (excluding any thirdparty libraries).

There are two exceptions to the formatting done by clang-format:

1) ComponentInstanceImpl.hpp: The '#include "Binders.hpp"' statement must be included after the includes above it since it relies on a using statement which they declare. Because clang-format will sort includes based on alphabetical order, this broke the code and so it was moved lower with a white space in between to keep it from sorting.

This seems like something we can create an issue out of since I don't think include order for "Binders.hpp" should matter.

2) ReferenceAutogenFiles.hpp: The order of the strings in the file matters. Clang-format was incorrectly changing the order of text within double quotes because an R string was used. This means that there were no quotes in front of the text on each line and thus clang-format treated it as actual c++ code.

The tests rely on a specific string which is defined in this file and so I manually reverted the formatting changes made.

### DIFF
--- a/cmake/CMakeResourceDependencies.cpp
+++ b/cmake/CMakeResourceDependencies.cpp
@@ -26,6 +26,6 @@ namespace {
 // empty object files.
 struct CMakeResourceDependencies
 {
-  CMakeResourceDependencies() { }
+  CMakeResourceDependencies() {}
 };
 }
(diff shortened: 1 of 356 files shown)


Jeeyuen-Rick (approved):

jeffdiclemente:
@achristoforides please create an issue for item 1, the order of the #include shouldn't matter. This seems like a bug.

jeffdiclemente (approved):

pelliott-mathworks (approved):

kevinleeMW (approved):

PR #572, opened 2021-06-01
Fixes #564, prevent nullptr from being passed to service constructors

### LINKED ISSUE #564: DS can pass nullptr to DS implementation-class constructor for a reference with cardinality 1..1 or 1..n
the expected behavior is that either the service component is satisfied or if the service reference cannot be bound (e.g. the service dependency's ctor throws an exception), the service component is not satisfied. A nullptr should never be passed as a service reference to the ctor of a service object if that service component requires a service dependency (1..1 or 1..n).

achristoforides:
This PR fixes #564 .

If a service depends on some upstream service and that service fails to construct or activate, a nullptr is passed to the dependent service and is still able to be created even if the cardinality is 1..1 or 1..n.

In the case where construction or activation of the upstream service fails, the dependent service shouldn't be able to be constructed.

This PR fixes this issue and adds tests which verify the desired behavior.


compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -117,9 +117,23 @@ std::shared_ptr<void> ComponentContextImpl::LocateService(
...
+      // that the construction or activation of the service failed.
+      if (!service && nullDueToFailure) {
+        std::string errMsg = "Service ";
+        errMsg += name;
+        errMsg += " with type ";
+        errMsg += type;
+        errMsg += " failed to construct or activate.";
+
+        throw ComponentException(errMsg);
+      }
```
achristoforides, line 133: Add check for 1..1/n cardinality requirement to throw. Add new test which tries to get a service to a service with cardinality 0..1 (should get back nullptr and not throw)

framework/include/cppmicroservices/ServiceInterface.h
```diff
@@ -354,6 +354,42 @@ inline std::shared_ptr<void> ExtractInterface(const InterfaceMapConstPtr& map,
...
+  }
+
+  auto iter = map->find(interfaceId);
+  if (iter != map->end()) {
+    nullDueToFailure = true;
+    return iter->second;
+  }
+  return nullptr;
+}
+
```
achristoforides, line 392: Hide this so users cannot call this function. Place it in ComponentContextImpl and make it uncallable.

compendium/test_bundles/TestBundleDSUpstreamDependencyB/src/ServiceImpl.cpp
```diff
@@ -0,0 +1,10 @@
+#include "ServiceImpl.hpp"
+
+#include <stdexcept>
+
+namespace dependent {
+TestBundleDSUpstreamDependencyImpl::TestBundleDSUpstreamDependencyImpl() {}
```
achristoforides, line 6: default this

framework/include/cppmicroservices/ServiceInterface.h
achristoforides, line 392: Add comment and potentially change parameter name to make the meaning of the parameter more clear.

framework/include/cppmicroservices/ServiceInterface.h
```diff
@@ -354,6 +354,42 @@ inline std::shared_ptr<void> ExtractInterface(const InterfaceMapConstPtr& map,
...
+ * @return The service interface pointer for the service interface id or
+ *         \c nullptr if \c map does not contain an entry for the given type.
+ *
+ * @see ExtractInterface(const InterfaceMapConstPtr&)
+ */
+inline std::shared_ptr<void> ExtractInterface(const InterfaceMapConstPtr& map,
+                                              const std::string& interfaceId,
+                                              bool& nullDueToFailure)
+{
+  if (!map) {
```
achristoforides, line 375: default nullDueToFailure to false.

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
achristoforides: This should actually return a valid pointer AND not throw.

compendium/DeclarativeServices/test/TestExceptionOnConstructionFailure.cpp
```diff
@@ -0,0 +1,148 @@
+/*=============================================================================
```
jeffdiclemente, line 1: I'd suggest naming this file similarly to the test fixture class; TestFailedBoundServiceActivation.cpp

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -103,6 +106,133 @@ ComponentContextImpl::GetProperties() const
...
+ * - interfaceId, The interface id string.
+ * - isSIPointerFoundOrMapEmpty, a boolean which represents if the return value was in the map
+ *     or the map is nullptr
+ * - return value: The service interface pointer for the service interface id or
+ *                 nullptr if "map" does not contain an entry for the given type.
+ */
+static std::shared_ptr<void> GetInterfacePointerAndCheckError(
+  const InterfaceMapConstPtr& map,
+  const std::string& interfaceId,
+  bool& isSIPointerFoundOrMapEmpty)
```
jeffdiclemente, line 130: I recommend following the cpp core guidelines as it relates to return values and "out" parameters, specifically: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rf-out and https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rf-out-multi.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -241,34 +233,27 @@ std::shared_ptr<void> ComponentContextImpl::LocateService(
...
+       */
+      bool isValid = (static_cast<bool>(lookupInfo.service) ==
+                      lookupInfo.wasFoundInMapOrMapEmpty);
       if (isValid) {
-        return service;
+        return lookupInfo.service;
       } else { // Checking for error condition and throwing if necessary
-        ValidateExtractedService(
-          configManagerPtr, name, type, service, isSIPointerFoundOrMapEmpty);
+        ValidateExtractedService(configManagerPtr, name, type, lookupInfo);
```
jeffdiclemente: It may be simpler to put all this logic into a single function that will either return the service object or throw.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -103,6 +104,146 @@ ComponentContextImpl::GetProperties() const
...
+ *   - true, if the service pointer was in the map
+ *       or if the map if null (service can be either
+ *       valid or a nullptr)
+*    - false if the service pointer was not found
+ *       in the map (implies service is nullptr)
+ */
+struct ServiceInterfacePointerLookupInfo
+{
+  std::shared_ptr<void> service;
+  bool wasFoundInMapOrMapEmpty;
```
jeffdiclemente, line 123: this should be default initialized to a value.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -103,6 +104,146 @@ ComponentContextImpl::GetProperties() const
...
+   * service pointer is not null and it was found in the map.
+   * 
+   * If this condition is not true, then we proceed with error checking and throw
+   * a ComponentException if necessary.
+   */
+  bool isValid = lookupInfo.IsValid();
+  if (isValid) {
+    return lookupInfo.service;
+  } else { // Checking for error condition and throwing if necessary
+    ValidateExtractedService(configManagerPtr, name, type, lookupInfo);
```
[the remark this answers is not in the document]
achristoforides: Without the extra function, we would be duplicating the code. In both areas where it is used, the same exact code is ran as well. Additionally, the function itself is roughly 30 lines and so in terms of code readability, it seems to make sense to me to extract the behavior into a separate function. If you would like me to remove the function, then I can do that as well.

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -103,6 +104,146 @@ ComponentContextImpl::GetProperties() const
...
+ * 
+ * This function has a similar implementation to ExtractService except it returns
+ * an instance of the ServiceInterfacePointerLookupInfo struct to do error handling
+ * if necessary.
+ * 
+ * - map, a InterfaceMap instance.
+ * - interfaceId, The interface id string.
+ * - return value: ServiceInterfacePointerLookupInfo
+ */
+static ServiceInterfacePointerLookupInfo GetInterfacePointer(
```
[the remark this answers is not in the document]
achristoforides: Since it is marked static at the namespace scope, it gives the function "internal linkage" which means that it is only accessible within the translation unit (barring use from any other translation units). This was included since we wanted to ensure that users of cppmicroservices weren't able to call this function.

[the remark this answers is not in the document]
jeffdiclemente: I see. my original concern was with public API users, not cppmicroservices maintainers accessing the function. This allows you to make this a private member function instead of static. I think making the function non-static is a better choice.

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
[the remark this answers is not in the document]
jeffdiclemente: what am I missing? `GetServicePointer` is the only function which calls `ValidateExtractedService`.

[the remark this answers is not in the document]
achristoforides: You are correct. I forgot that during the last refactoring, I made both LocateService and LocateServices use GetServicePointer. I will fix this. Thanks.

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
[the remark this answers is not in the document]
achristoforides: I agree with your decision on making it non-static, however I do not think it should be a private function member in the ComponentContextImpl class. It is only used as an implementation detail when using LocateService/Services and has a very niche use case. I'm going to remove the static but not add it in the ComponentContextImpl.hpp file.

carneyweb (approved):

kevinleeMW (approved):

PR #573, opened 2021-06-01
Fix -Wmissing-prototypes warning regarding macros

achristoforides:
This PR fixes the -Wmissing-prototypes warning which both the CPPMICROSERVICES_INITIALIZE_BUNDLE and CPPMICROSERVICES_EXPORT_BUNDLE_ACTIVATOR macros were producing.

To get rid of the error, prototypes for the functions being defined were included above the actual definitions.

Another way to approach this is to include pragma statements in the code to define how the compiler/linker should act but this would be tricky since you are not able to put preprocessor statements in a macro definition. Additionally, the fix for the warning is simple and the amount of code added is negligible.

### DIFF
--- a/framework/include/cppmicroservices/BundleActivator.h
+++ b/framework/include/cppmicroservices/BundleActivator.h
@@ -128,10 +128,14 @@ struct BundleActivator
  */
 #define CPPMICROSERVICES_EXPORT_BUNDLE_ACTIVATOR(_activator_type)              \
   extern "C" US_ABI_EXPORT cppmicroservices::BundleActivator*                  \
+    US_CREATE_ACTIVATOR_FUNC(US_BUNDLE_NAME)();                                \
+  extern "C" US_ABI_EXPORT cppmicroservices::BundleActivator*                  \
     US_CREATE_ACTIVATOR_FUNC(US_BUNDLE_NAME)()                                 \
   {                                                                            \
     return new _activator_type();                                              \
   }                                                                            \
+  extern "C" US_ABI_EXPORT void US_DESTROY_ACTIVATOR_FUNC(US_BUNDLE_NAME)(     \
+    cppmicroservices::BundleActivator * activator);                            \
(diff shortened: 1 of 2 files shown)


jeffdiclemente (approved):

framework/include/cppmicroservices/BundleActivator.h
```diff
@@ -127,11 +127,15 @@ struct BundleActivator
  * \snippet uServices-activator/main.cpp 0
  */
 #define CPPMICROSERVICES_EXPORT_BUNDLE_ACTIVATOR(_activator_type)              \
+  extern "C" US_ABI_EXPORT cppmicroservices::BundleActivator*                  \
+    US_CREATE_ACTIVATOR_FUNC(US_BUNDLE_NAME)();                                \
```
achristoforides, line 131: Add single line comment to explain why the declaration is included

pelliott-mathworks (approved):

framework/include/cppmicroservices/BundleInitialization.h
```diff
@@ -60,12 +60,16 @@ class BundleContextPrivate;
     std::atomic<cppmicroservices::BundleContextPrivate*> US_CTX_INS(           \
       US_BUNDLE_NAME){};                                                       \
                                                                                \
+    extern "C" cppmicroservices::BundleContextPrivate* US_GET_CTX_FUNC(        \
```
achristoforides, line 63: Add single line comment to explain why the declaration is included

framework/include/cppmicroservices/BundleActivator.h
achristoforides: Deciding not to do this, the inclusion of that line should be obvious (regarding what it syntactically means (prototype definition)) (even if somebody doesn't know what warning we are trying to avoid)

framework/include/cppmicroservices/BundleInitialization.h
achristoforides: Deciding not to do this, the inclusion of that line should be obvious (regarding what it syntactically means (prototype definition)) (even if somebody doesn't know what warning we are trying to avoid)

PR #577, opened 2021-06-04
Migrate tests from driver to gtest

apatil19:
- ResourceCompilerTest
- ServiceListenerTest


framework/test/gtest/ServiceListenerTest.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+  // Start libBundleTestSL1 to ensure that the Service interface is available.
+  EXPECT_NO_THROW(libSL1.Start()) << "Failed to load bundle";
+
+  // Start libBundleTestSL4 that will require the serivce interface and publish
+  // cppmicroservices::FooService
+  EXPECT_NO_THROW(libSL4.Start()) << "Failed to load bundle";
+
+  // Start libBundleTestSL3 that will require the serivce interface and get the service
+  EXPECT_NO_THROW(libSL3.Start()) << "Failed to load bundle";
```
jeffdiclemente, line 302: consistency; `EXPECT_NO_THROW` is used but not consistently in this file or in `usResourceCompilerTest.cpp` (where the original test points checked for exceptions being thrown). 

I can understand not using the `*_NO_THROW` macro in all of those instances since an exception would fail the test and that is the desired outcome. I'd suggest being consistent in the use of this type of test macro. It seems like the use of the macro is unnecessary so my suggestion would be to remove the use of it where it makes sense.

jeffdiclemente (changes requested):

framework/test/gtest/ResourceCompilerTest.cpp
```diff
@@ -38,44 +38,97 @@
 
 using namespace cppmicroservices;
 using namespace cppmicroservices::util;
-
+using namespace cppmicroservices::testing;
 namespace {
 
+TempDir uniqueTempdir = MakeUniqueTempDirectory();
```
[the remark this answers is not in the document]
apatil19: The Idea behind making it global that MakeUniqueTempDirectory() creates a unique temp dir with random alphanumeric name. 
However, we want that temp dir to be constant for all the test points.

So, if I move it to the Fixture, MakeUniqueTempDirectory() would create different dir for different test point causing them to fail.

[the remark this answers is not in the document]
jeffdiclemente: fair point. i'd suggest making a note of that in a comment. that's something that's not obvious when reading the code.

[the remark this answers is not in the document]
apatil19: Sure. Ill make sure to mention it in comment.

jeffdiclemente (approved):

framework/test/gtest/ResourceCompilerTest.cpp
```diff
@@ -38,44 +38,98 @@
...
+    mz_zip_writer_finalize_archive(&zip);
+    mz_zip_writer_end(&zip);
+  }
+
+  ManifestZipFile(const ManifestZipFile&) = delete;
+  void operator=(const ManifestZipFile&) = delete;
+  ManifestZipFile(const ManifestZipFile&&) = delete;
+  ManifestZipFile& operator=(const ManifestZipFile&&) = delete;
+
+  ~ManifestZipFile() {}
```
Jeeyuen-Rick, line 113: default?

framework/test/gtest/ResourceCompilerTest.cpp
```diff
@@ -448,12 +548,74 @@ void testBundleManifestZipAdd(const std::string& rcbinpath,
...
+  auto entryNames = zip.getNames();
+  testExists(entryNames, "mybundle/manifest.json");
+  testExists(entryNames, "mybundle/");
+  testExists(entryNames, "mybundle/resource1/resource1.txt");
+  testExists(entryNames, "mybundle/resource1/");
+  testExists(entryNames, "mybundle/resource2/resource2.txt");
+  testExists(entryNames, "mybundle/resource2/");
+
+  cmd.str("");
+  cmd.clear();
```
Jeeyuen-Rick, line 586: Are they doing the same thing?

Jeeyuen-Rick: How about cmd.str(std::string());

apatil19: yes changed it to cmd.str(std::string());

framework/test/gtest/ResourceCompilerTest.cpp
apatil19: Did not change it from original test.
Also, default doesn't work here.

Jeeyuen-Rick (approved):

PR #534, opened 2021-03-22
Fix another EOL issue in BundleResouceTest.cpp

achristoforides:
- Renamed TestResource.txt to TestResource.ptxt to preserve line
  endings

Signed-off-by: The MathWorks, Inc. <alchrist@mathworks.com>

PR #578, opened 2021-06-07
Fixes #575, Remove line ending dependency in StaticBundleResourceTest

### LINKED ISSUE #575: StaticBundleResourceTest relies on line endings
The StaticBundleResourceTest relies on line endings. It loads a bundle with resources and then checks the contents of the files.

This can be fixed by updating the .txt files to be .ptxt files and by modifying the test so that the files being opened have the correct name.

achristoforides:
Fixes #575 .

This PR removes the line ending dependency in the StaticBundleResourceTest.

The test tries to open the .txt files in the bundle and checks their contents. However, this can be a problem if .txt files adhere to the developer's git settings for changing line endings when pulling the files on various operating systems.

For instance, if the user clones the repo on a Windows machine, and then access the repo directory with WSL (Windows Subsystem for Linux) to build and test the software without having to run a separate VM, then the test fails because the file contents check expects there to be no "\r" (but it exists due to git switching line endings when pulling the files on Windows).

To fix this, this PR leverages an older change that was made. The files that have their contents checked were switched to ".ptxt" files which will retain UNIX line endings on any operating system or despite what git settings a user has. By doing this, platform specific logic is not required in the test and overall makes the test suite that we have more consistent.

For reference, the ".ptxt" change previously mentioned can be found here (https://github.com/CppMicroServices/CppMicroServices/pull/531) and another PR which uses this change (to serve as another example) can be found here (https://github.com/CppMicroServices/CppMicroServices/pull/534)


jeffdiclemente (approved):

kevinleeMW (approved):

pelliott-mathworks (approved):

carneyweb (commented):
Ship It!

PR #580, opened 2021-06-07
Fixes #579, Removes compiler warnings generated by MSVC due to usage of Boost

### LINKED ISSUE #579: Compiling on Windows using MSVC generates many warnings (not compiler warnings) about defines
When compiling CppMicroServices on Windows using MSVC 2019 (potentially earlier versions as well), the following message appears after many source files are compiled:

```
Please define _WIN32_WINNT or _WIN32_WINDOWS appropriately. For example:
- add -D_WIN32_WINNT=0x0601 to the compiler command line; or
- add _WIN32_WINNT=0x0601 to your project's Preprocessor Definitions.
Assuming _WIN32_WINNT=0x0601 (i.e. Windows 7 target).
```

These have been appearing since the inclusion of a subset of boost for the thread pool implementation for DS.

The fix for these is easy (adding a specific define to our compile flags).

The question that is still open is which version define should we use? Should we use what the compiler (and warning) message assumes (Windows 7 target) or should this be something different?

The full set of options can be found here: https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?redirectedfrom=MSDN&view=msvc-160

achristoforides:
Fixes #579 .

This PR introduces a small change to our top-level CMakeLists.txt file which removes the compiler warnings generated by MSVC due to our usage of Boost for the thread pool in DS.

The message looks like the following:
```
Please define _WIN32_WINNT or _WIN32_WINDOWS appropriately. For example:
- add -D_WIN32_WINNT=0x0601 to the compiler command line; or
- add _WIN32_WINNT=0x0601 to your project's Preprocessor Definitions.
Assuming _WIN32_WINNT=0x0601 (i.e. Windows 7 target).
```

Instead of relying on the assumed value for this define, we now explicitly define it as part of our compilation flags set by CMake.

The compiler message is no longer seen.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -401,6 +401,15 @@ set(US_TSAN_LINK_FLAGS )
 set(US_CXX_COVERAGE_FLAGS )
 
 if(MSVC)
+  # Since we are compiling with boost, there exists some code which looks
+  # at the value of this define. Since we previously did not specify the
+  # define, boost assumed a value (0x0601, meaning Windows 7). We will
+  # now explicitly include this define to remove the compiler messages
+  # it generates.
+  set(msvc_version_define
+    "_WIN32_WINNT=0x0601"
+  )
+
   set(disabled_warnings
     4251 # 'identifier' : class 'type' needs to have dll-interface to be used by clients of class 'type2'
   )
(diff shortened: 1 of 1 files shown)


jeffdiclemente (approved):

kevinleeMW (approved):
ship it!

pelliott-mathworks (approved):

carneyweb (approved):
Ship It!

PR #581, opened 2021-06-07
Fixes TSAN loading on Linux

achristoforides:
In a previous PR, the check for compiler flags related to TSAN was modified so that it checked for "-01 -fsanitize=thread" instead of "-O1 -fsanitize=thread".

This PR fixes that typo.


jeffdiclemente (approved):
I would have expected that `usFunctionCheckCompilerFlags` would have caught this typo the first time. any idea why it didn't?

achristoforides (commented):
I'm certain it would have (which is how I caught it now). My guess is that nobody used TSAN recently (myself included).

pelliott-mathworks (approved):

carneyweb (approved):
Ship It!

PR #586, opened 2021-06-18
fix segfault when service class ctor throws

### LINKED ISSUE #585: segfault when a DS service configuration implements Activate/Deactivate methods and the service impl ctor throws
reproduction steps:
- create a DS bundle which implements a service, `foo` that throws in its constructor
- implement the `Activate` and `Deactivate` methods for the `foo` service configuration
- call `GetService` for a 'foo` service
- a segfault should occur

The failure is caused by not checking for a valid service object pointer when calling `Activate` or `Deactivate`.

jeffdiclemente:
Fixes #585 

### DIFF
--- a/compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
+++ b/compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
@@ -445,12 +445,32 @@ ComponentConfigurationImpl::CreateAndActivateComponentInstanceHelper(
   auto componentInstance = CreateComponentInstance();
   auto ctxt =
     std::make_shared<ComponentContextImpl>(shared_from_this(), bundle);
+  /*
+   * Failing to construct the servce object is an unrecoverable 
+   * failure which will cause the service configuration to not
+   * be active.
+   */
   try {
-    componentInstance->CreateInstanceAndBindReferences(ctxt);
+    componentInstance->CreateInstance(ctxt);
   } catch (const std::exception&) {
     logger->Log(
       cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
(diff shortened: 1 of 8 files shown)


achristoforides (approved):

jeffdiclemente:
travis build failures are fixed in PR #588

carneyweb (approved):
Ship it.

kevinleeMW (approved):

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
```diff
@@ -445,12 +445,32 @@ ComponentConfigurationImpl::CreateAndActivateComponentInstanceHelper(
...
   try {
-    componentInstance->CreateInstanceAndBindReferences(ctxt);
+    componentInstance->CreateInstance(ctxt);
   } catch (const std::exception&) {
     logger->Log(
       cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
-      "Exception while creating component instance and binding references.",
+      "Exception while creating component instance.",
+      std::current_exception());
+    throw;
```
pelliott-mathworks, line 460: Just wondering if it's okay to leave the ComponentConfigurationImpl object in the state it's in. Maybe this component should be disabled since it can't be constructed?

pelliott-mathworks (approved):

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
jeffdiclemente: The OSGi spec doesn't make any assertion about moving the state to disabled. The state will remain in satisfied and never move to active.

PR #596, opened 2021-07-01
Migrate tests to Gtest

apatil19:
- ServiceFactoryTest
- BundleTest 


framework/test/gtest/BundleTest.cpp
```diff
@@ -0,0 +1,906 @@
...
+  ASSERT_NE(bundleB, bundleA);
+
+  bundleB = bundleA;
+
+  //Test bundle object assignment.
+  ASSERT_EQ(bundleB.GetBundleId(), bundleA.GetBundleId());
+  //Test that the bundle objects are equivalent.
+  ASSERT_EQ(bundleB, bundleA);
+}
+
```
jeffdiclemente, line 719: missing a test, it is `TestBundleStreamOperator()` in the test driver

framework/test/gtest/ServiceFactoryTest.cpp
```diff
@@ -201,3 +232,389 @@ TEST_F(ServiceFactoryTest, TestGetServiceReturnsNull)
...
+  bundleH.Stop();
+}
+
+#ifdef US_ENABLE_THREADING_SUPPORT
+
+// test that concurrent calls to ServiceFactory::GetService and ServiceFactory::UngetService
+// don't cause race conditions.
+TEST_F(ServiceFactoryTest, TestConcurrentServiceFactory)
+{
+  auto framework2 = framework;
```
jeffdiclemente, line 409: is this necessary? why is a copy being made?

framework/test/gtest/ServiceFactoryTest.cpp
```diff
@@ -201,3 +232,389 @@ TEST_F(ServiceFactoryTest, TestGetServiceReturnsNull)
...
+  //Test that one FrameworkEvent was sent
+  ASSERT_EQ(2, fwEvents.size());
+  //Test for correct framwork event type
+  ASSERT_EQ(fwEvents[0].GetType(), FrameworkEvent::FRAMEWORK_ERROR);
+  //Test for correct framwork event bundle
+  ASSERT_EQ(fwEvents[0].GetBundle(), context.GetBundle());
+
+  //Test for correct service exception type (recursion)
+  EXPECT_THROW(std::rethrow_exception(fwEvents[0].GetThrowable()),
+               cppmicroservices::ServiceException);
```
jeffdiclemente, line 583: there is a missing test point for :`

`US_TEST_CONDITION_REQUIRED(
        exc.GetType() == ServiceException::FACTORY_RECURSION,`

jeffdiclemente (changes requested):

framework/test/gtest/BundleTest.cpp
apatil19: TestBundleStreamOperator() simply installs 'TestBundleA' (which is being tested in other testpoints) and gives out bundle information (id, location, etc.) as an output.
There was not any validations as such. Hence, I removed it.

framework/test/gtest/ServiceFactoryTest.cpp
apatil19: line 417 : worker_threads.push_back(std::thread([framework2]() requires a variable instead of member ('framework' in this case).
So I can either start a new framework within this test-point or use a copy.

framework/test/gtest/BundleTest.cpp
jeffdiclemente: The operation of outputting the `Bundle` object is what is being tested. Unfortunately there wasn't a testpoint for it that validated a behavior. However running that code can uncover bugs in the `Bundle`'s operator<<.

apatil19: Ill add it to the test

PR #607, opened 2021-08-05
Fix clang build

achristoforides:
This PR fixes the clang build which was broken after GitHub updated the runners for GitHub Actions.

This issue is that there is a bug in CMake 3.21.0+ where if you are using clang on linux without lld installed, CMake will not use ld as the linker even if it is installed. See filed CMake issue: https://gitlab.kitware.com/cmake/cmake/-/issues/22500

This PR introduces changes to our CMake files which check to see if we are using lld; if we are, extra link flags are added since using lld requires specifying an emulation target (elf_amd64_fbsd).

Additionally, the clang build configuration was re-enabled and updated to always install lld so that we shouldn't experience this issue again.

### DIFF
--- a/cmake/usFunctionCheckResourceLinking.cmake
+++ b/cmake/usFunctionCheckResourceLinking.cmake
@@ -16,7 +16,7 @@ function(usFunctionCheckResourceLinking)
       set(_suffix .rc)
     elseif(UNIX)
       execute_process(
-        COMMAND ${CMAKE_LINKER} -r -b binary -o "${CMAKE_CURRENT_BINARY_DIR}/us_resource_link.o" "${CMAKE_COMMAND}"
+        COMMAND ${CMAKE_LINKER} -r -o "${CMAKE_CURRENT_BINARY_DIR}/us_resource_link.o" ${_ADDITIONAL_LINKER_FLAGS} --format=binary "${CMAKE_COMMAND}"
         RESULT_VARIABLE _result
       )
       if(_result EQUAL 0)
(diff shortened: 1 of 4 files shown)


kevinleeMW (approved):

cmake/usFunctionCheckResourceLinking.cmake
```diff
@@ -16,7 +16,7 @@ function(usFunctionCheckResourceLinking)
       set(_suffix .rc)
     elseif(UNIX)
       execute_process(
-        COMMAND ${CMAKE_LINKER} -r -b binary -o "${CMAKE_CURRENT_BINARY_DIR}/us_resource_link.o" "${CMAKE_COMMAND}"
+        COMMAND ${CMAKE_LINKER} -r -o "${CMAKE_CURRENT_BINARY_DIR}/us_resource_link.o" ${_ADDITIONAL_LINKER_FLAGS} --format=binary "${CMAKE_COMMAND}"
```
pelliott-mathworks, line 19: Didn't notice this defined in these files. Must be someplace else in the build files?

pelliott-mathworks (approved):
Great work!

jeffdiclemente (approved):

cmake/usFunctionCheckResourceLinking.cmake
achristoforides: _ADDITIONAL_LINKER_FLAGS is defined in the top-level CMakeLists.txt file

PR #612, opened 2021-08-11
Fix deadlock in ConfigAdmin Update, UpdateIfDifferent and Remove

pelliott-mathworks:
The Update, UpdateIfDifferent and Remove methods on Configuration objects send
an asynchronous notification to all services that have published a ConfigurationListener
interface. The asynchronous thread creates a std::future and saves it in a vector (incompleteFutures). 
A ConfigurationAdminImpl non-public method (WaitForAllAsync) can be used to wait for all futures in the vector to complete. The future created by the asynchronous notification thread is also shared (std::shared_future) and is returned to the caller of the Update, UpdateIfDifferent or Remove method so they can wait for the operation to complete. There was a bug in the initial version of ConfigurationAdmin that resulted in a deadlock under some circumstances because a std::future was being saved in the incompleteFutures vector instead of a std::shared_future. When the std::shared_future went out of scope, the destructor would stall because the std::future would still exist. This test confirms that
this deadlock no longer exists.

The actual situation in the field involved a complex problem where the notification of the updated configuration object in turn resulted in the code doing the updating being called. Instead of trying to duplicate this in user code, this tests creates a blocking ConfigurationListener and recreates the blocking circumstances that way. 

The code was tested by verifying that the test would stall (deadlock) when the value stored in the incompleteFutures and completeFutures vectors was of type std::future and does not stall in it’s present condition with the value stored in those two vectors of type std::shared_vector.

### DIFF
--- a/compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
+++ b/compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
@@ -66,6 +66,7 @@ struct BlockingConfigurationListener final
   void configurationEvent(
     const cppmicroservices::service::cm::ConfigurationEvent& event) noexcept
   {
+    auto type = event.type;  //just to avoid the unused parameter error on some platforms.
     {
       std::lock_guard<std::mutex> lg{ mutex_ };
     }
--- a/compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
+++ b/compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
@@ -141,7 +141,7 @@ TEST_F(TestConfigurationAdminImpl, VerifyNoUpdateDeadlock)
       EXPECT_EQ(std::future_status::timeout,
                 fut.wait_for(std::chrono::milliseconds(10)));
     }
-    //fut goes out of scope here. If the destructor incorrectly blocks, the test would stall now.
+    //fut goes out of scope here. If the destructor blocks, the test would stall now.
   }
   // Lock is out of scope here, so configurationEvent should be unblocked, and the promise should be set.
   callCompletedFut.get();
--- a/compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
+++ b/compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
@@ -41,6 +41,40 @@ MATCHER_P(AnyMapEquals, value, "")
   cppmicroservices::any_value_to_string(valueString, value);
   return (argString.str() == valueString.str());
 }
+/* This is a ConfigurationListener that blocks on a mutex. It is used by the TestConfigurationAdminImpl
+ * VerifyNoUpdateDeadlock test to recreate the blocking condition that caused a deadlock for the 
+ * Update, UpdateIfDifferent and Remove Configuration methods
+ */
+struct BlockingConfigurationListener final
+  : public cppmicroservices::service::cm::ConfigurationListener
+{
+
+  BlockingConfigurationListener() = default;
+  BlockingConfigurationListener(const BlockingConfigurationListener&) = delete;
+  BlockingConfigurationListener(BlockingConfigurationListener&&) = delete;
+  BlockingConfigurationListener& operator=(
+    const BlockingConfigurationListener&) = delete;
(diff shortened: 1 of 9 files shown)


Jeeyuen-Rick (approved):

compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
```diff
@@ -41,6 +41,40 @@ MATCHER_P(AnyMapEquals, value, "")
...
+  BlockingConfigurationListener& operator=(BlockingConfigurationListener&&) =
+    delete;
+  ~BlockingConfigurationListener() = default;
+
+  /*
+   * configurationEvent is the method called by Configuration Admin whenever a 
+   * a configuraton object is updated or removed.
+   *
+   */
+  void configurationEvent(
```
achristoforides, line 66: I would make the first letter of the function name ('c') capital to match the formatting we use throughout the codebase

compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
```diff
@@ -41,6 +41,40 @@ MATCHER_P(AnyMapEquals, value, "")
...
+    const cppmicroservices::service::cm::ConfigurationEvent& event) noexcept
+  {
+    {
+      std::lock_guard<std::mutex> lg{ mutex_ };
+    }
```
achristoforides, line 71: This isn't locking anything. Is there a reason we acquire a lock, let the lock go out of scope (releasing our locked access to shared data), and then calling `promise_.set_value()` in an unlocked manner?

compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
```diff
@@ -62,10 +96,57 @@ class TestConfigurationAdminImpl : public ::testing::Test
...
+
+  {
+    std::lock_guard<std::mutex> lg{ blockingConfigurationListener->mutex_ };
+
+    {
+      auto fut = configuration->Update(props);
+      EXPECT_EQ(std::future_status::timeout,
+                fut.wait_for(std::chrono::milliseconds(10)));
+    }
+    //fut goes out of scope here. If the destructor incorrectly blocks, the test would stall now.
```
achristoforides, line 144: The future destructor blocking here is not incorrect. It is part of the C++ standard. (see https://en.cppreference.com/w/cpp/thread/future/~future)

framework/src/util/Any.cpp
```diff
@@ -106,12 +106,20 @@ std::ostream& any_value_to_json(std::ostream& o,
         o << "\\t";
         break;
       default:
+// suppress type-limits error on linux arm 64 build
```
achristoforides, line 109: type-limits "warning", not "error". We just treat warnings as errors.

compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
pelliott-mathworks: configurationEvent is a method defined in the OSGI spec

compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
pelliott-mathworks: The combination of this lock and the std::future being saved in the incompleteFutures vector is what recreates the deadlock that we saw in the user code. Certainly not anything you would do in normal code but the easiest way to recreate it.

compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
pelliott-mathworks: Yes. And it contributed to the deadlock. Replacing the std::future with the shared_future in the vector eliminated the deadlock

jeffdiclemente: minor: I'm not sure "incorrectly" should be used in the comment. the future's destructor blocking is not necessarily incorrect in and of itself. consider removing "incorrectly" from the comment.

jeffdiclemente (approved):

achristoforides (approved):

PR #613, opened 2021-08-12
Fix bug when an AddingService method returns nullptr

### LINKED ISSUE #471: service tracker incorrectly tracks service if ServiceTrackerCustomizer::AddingService returns nullptr or throws
It appears that if a ServiceTrackerCustomizer's AddingService method returns nullptr or throws that the service is still "tracked" by the service tracker, as evidenced by walking the code and ServiceTracker::GetServiceReferences() returning a non-zero integer. Furthermore there is a performance related side-effect whereby ServiceTracker::WaitForService() will go into a tight loop until the timeout is reached instead of waiting.

The OSGi spec states [that a nullpr is returned from AddingService if the service is not tracked](https://docs.osgi.org/specification/osgi.core/7.0.0/util.tracker.html#org.osgi.util.tracker.ServiceTrackerCustomizer.addingService-ServiceReference-).

I believe the source of the problem comes from `BundleAbstractTracked.tpp` on line 120. `find` should be used instead of `operator[]`.

test case:

```
struct MyInterfaceOne
{
  virtual ~MyInterfaceOne() {}
};

class MyNullPtrCustomizer
  : public cppmicroservices::ServiceTrackerCustomizer<MyInterfaceOne>
{

public:
  MyNullPtrCustomizer(const BundleContext& context)
    : m_context(context)
  {}

  virtual std::shared_ptr<MyInterfaceOne> AddingService(
    const ServiceReference<MyInterfaceOne>& )
  {
    return nullptr;
  }

  virtual void ModifiedService(
    const ServiceReference<MyInterfaceOne>& ,
    const std::shared_ptr<MyInterfaceOne>& )
  {
  }

  virtual void RemovedService(const ServiceReference<MyInterfaceOne>& ,
                              const std::shared_ptr<MyInterfaceOne>& )
  {
  }

private:
  BundleContext m_context;
};

void TestNullPtrServiceTrackerCustomizer(BundleContext context)
{
  MyNullPtrCustomizer customizer(context);

  cppmicroservices::LDAPFilter filter(
    "(" + cppmicroservices::Constants::SERVICE_ID + ">=0)");
  cppmicroservices::ServiceTracker<MyInterfaceOne> tracker(
    context, filter, &customizer);
  tracker.Open();

  struct MyServiceOne : public MyInterfaceOne
  {};

  auto serviceOne = std::make_shared<MyServiceOne>();

  context.RegisterService<MyInterfaceOne>(serviceOne);

  US_TEST_CONDITION(tracker.GetServiceReferences().size() == 0,
                    "tracking count");

  auto trackedObj = tracker.WaitForService(std::chrono::seconds(1));
  US_TEST_CONDITION(nullptr == trackedObj, "tracked object should be nullptr");
}
```

jeffdiclemente:
Fixes #471

### DIFF
--- a/framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
+++ b/framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
@@ -119,7 +119,7 @@ void BundleAbstractTracked<S,TTT,R>::Track(S item, R related)
     }
     auto trackedItemIter = tracked.find(item);
     if (trackedItemIter != tracked.end()) {
-      object = (*trackedItemIter).second;
+      object = trackedItemIter->second;
     }
     if (!object)
     { /* we are not tracking the item */
@@ -192,7 +192,7 @@ void BundleAbstractTracked<S,TTT,R>::Untrack(S item, R related)
      * must remove from tracker before
      * calling customizer callback
      */
-    object = (*trackedItemIter).second;
+    object = trackedItemIter->second;
     tracked.erase(item);
     Modified(); /* increment modification count */
   }
--- a/framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
+++ b/framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
@@ -75,7 +75,7 @@ void BundleAbstractTracked<S,TTT,R>::TrackInitial()
        */
       item = initial.front();
       initial.pop_front();
-      if (tracked[item])
+      if (tracked.end() != tracked.find(item))
       {
         /* if we are already tracking this item */
         DIAG_LOG(*bc.GetLogSink()) << "BundleAbstractTracked::trackInitial[already tracked]: " << item;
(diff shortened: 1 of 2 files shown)


pelliott-mathworks (approved):

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -117,10 +117,13 @@ void BundleAbstractTracked<S,TTT,R>::Track(S item, R related)
     {
       return;
     }
-    object = tracked[item];
+    auto trackedItemIter = tracked.find(item);
+    if (trackedItemIter != tracked.end()) {
+      object = (*trackedItemIter).second;
```
achristoforides, line 122: Consider using `trackedItemIter->second` syntax here

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -180,16 +183,17 @@ void BundleAbstractTracked<S,TTT,R>::Untrack(S item, R related)
...
+    auto trackedItemIter = tracked.find(item);
+    // nothing to do, no item is being tracked.
+    if (trackedItemIter == tracked.end()) {
+      return;
+    }
     /*
      * must remove from tracker before
      * calling customizer callback
      */
+    object = (*trackedItemIter).second;
```
achristoforides, line 195: Consider using `trackedItemIter->second` syntax here

achristoforides (approved):

PR #614, opened 2021-08-23
Fix ConfigurationAdmin ListConfigurations

pelliott-mathworks:
There was an order dependency for services requiring configuration objects. If the configuration object was available before the component was initialized, DS would not realize the configuration object was available and would not satisfy the component. The problem was with the ListConfigurations method. It was not returning the configuration objects correctly even though they existed in the Configuration Admin repository. Fixed the ListConfigurations method and added two tests. Signed-off-by <pelliott@mathworks.com>


compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
```diff
@@ -29,6 +29,78 @@
...
+  ASSERT_TRUE(configAdminService)
+    << "GetService failed for ConfigurationAdmin.";
+
+  // Create configuration object and update properties BEFORE installing and
+  // starting the bundle which defines the service.
+  auto configuration = configAdminService->GetConfiguration("sample::ServiceComponentCA02");
+  configuration->UpdateIfDifferent(
+    std::unordered_map<std::string, cppmicroservices::Any>{ { "foo", true} });
+
+  // Install and start the bundle which has the server
```
jeffdiclemente, line 49: typo?: server => service ?

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
```diff
@@ -29,6 +29,78 @@
...
+    << "GetService failed for ConfigurationAdmin.";
+
+  // Confirm that the configuration object has been added to the Configuration
+  // Admin repository.
+  auto configObjects =
+    configAdminService->ListConfigurations("(pid=" + configObject + ")");
+  EXPECT_EQ(configObjects.size(), 1ul)
+      << "One configuration object expected.";
+
+  // Since the configuration object is present the state of the
```
jeffdiclemente, line 94: sentence fragment.

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
```diff
@@ -29,6 +29,78 @@
...
+    << "GetService failed for ConfigurationAdmin.";
+
+  // Create configuration object and update properties BEFORE installing and
+  // starting the bundle which defines the service.
+  auto configuration = configAdminService->GetConfiguration("sample::ServiceComponentCA02");
+  configuration->UpdateIfDifferent(
+    std::unordered_map<std::string, cppmicroservices::Any>{ { "foo", true} });
+
+  // Install and start the bundle which has the server
+  auto testBundle = ::test::InstallAndStartBundle(ctx, "TestBundleDSCA02");
```
jeffdiclemente, line 50: Doesn't this bundle need to have a "cm" section in it to accurately recreate the use case? The failure case was creating and updating a configuration and then installing and starting a bundle containing a DS service and a "cm" section.

jeffdiclemente (changes requested):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -310,32 +310,60 @@ ConfigurationAdminImpl::GetFactoryConfiguration(const std::string& factoryPid,
...
+ * be "pid=startup.options", (to search for a configuration object with a matching pid)
+ * "pid=virtualfilesystem~user1", (to search for a configuration object with a matching factory pid)
+ *  "key1=abc" (to search for a configuration object containing a property with key "key1" with a value "abc".
+ * Regular expressions are allowed. 
+ */
 std::vector<std::shared_ptr<cppmicroservices::service::cm::Configuration>>
 ConfigurationAdminImpl::ListConfigurations(
   const std::string& filter)
 {
+
```
achristoforides, line 325: nitpick: consider removing empty line at the top of the function.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -310,32 +310,60 @@ ConfigurationAdminImpl::GetFactoryConfiguration(const std::string& factoryPid,
...
+		// The pid wasn't a match but the properties might be. Check those.
+		auto props = it.second->GetProperties();
         if (ldap.Match(props)) {
-          result.push_back(it.second);
-        }
+        result.push_back(it.second);
+		}
       }
-    }
+    }  // end for
```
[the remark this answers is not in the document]
pelliott-mathworks: This is probably related to having to reinstall VS this week. I have Tools->Options->Text Editor->C++->Formatting set as follows:
Enable ClangFormat support (checked)
Run ClangFormat for all formatting scenarios (selected).
Is there some additional configuration I need to do to get the clang formatting to start working again in VS?

compendium/test_bundles/TestBundleDSCA27/src/ServiceImpl.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <iostream>
+
+namespace sample {
+
+void ServiceComponentCA27::Modified(
+  const std::shared_ptr<ComponentContext>& /*context*/,
+  const std::shared_ptr<cppmicroservices::AnyMap>& configuration)
+{
+  std::lock_guard<std::mutex> lock(propertiesLock);
+  properties = *configuration;
```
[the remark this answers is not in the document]
pelliott-mathworks: We did this with all of the bundles that we created for ConfigAdmin DS integration testing. They should be fixed too. Should I fix the rest of them as part of this pull request or handle that separately as part of a different task/pull request?

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
pelliott-mathworks: yes

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
```diff
@@ -29,6 +29,78 @@
 #include "TestInterfaces/Interfaces.hpp"
 
 namespace test {
+
+TEST_F(tServiceComponent, testUpdatePropsBeforeInstallResolvesService)
```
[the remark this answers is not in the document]
pelliott-mathworks: testUpdateConfigBeforeStartingBundleResolvesService

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
pelliott-mathworks: I think there are two use cases. In one use case the configuration object is created programmatically then the bundle is started. In the second use case the configuration object is in the manifest.json file and is created by ConfigurationAdmin automatically when the bundle is started. In both cases the service should be resolved because the configuration object is available. That's why there are two new tests. The first one creates the config object programmatically. The second creates it in the manifest.json file.

compendium/test_bundles/TestBundleDSCA27/src/ServiceImpl.cpp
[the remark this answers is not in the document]
achristoforides: Addressed by: https://github.com/CppMicroServices/CppMicroServices/pull/616

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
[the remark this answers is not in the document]
jeffdiclemente: VS 2019 should automatically see the .clang-format file in the source code root directory and use it. If not, take a look through VS documentation.

jeffdiclemente:
there are failures in the TSan and ASan sanitizer builds and appveyor that need to be investigated before merging this PR.

compendium/DeclarativeServices/test/TestUpdateConfiguration.cpp
jeffdiclemente: yes, i agree, however the reported reproduction steps has the configuration being created and updated before the bundle containing the "cm" section is installed and started. The test case here does those operations in reverse and I think we should have a test case that reproduces the same reproduction steps.

jeffdiclemente: cleared this up offline. there are two separate reproduction cases. I retract my previous comment.

in addition there is one more use case that should be tested which is basically the merging of these two test cases: a bundle with services and "cm" data with the configuration being updated before the bundle is installed and started.

pelliott-mathworks: Added another use case that programmatically creates the configuration object before the bundle defining the service is started. The bundle defining the service also contains the same configuration object. Confirmed the service is resolved and the resulting properties match those in the manifest.json file since that update happens last.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -310,32 +310,58 @@ ConfigurationAdminImpl::GetFactoryConfiguration(const std::string& factoryPid,
...
+      } else {
+        // The pid wasn't a match but the properties might be. Check those.
         auto props = it.second->GetProperties();
         if (ldap.Match(props)) {
-          result.push_back(it.second);
+          result.emplace_back(it.second);
         }
       }
-    }
+    } // end for
```
achristoforides, line 363: We do not usually mark the end of for loops like we do with namespaces. This should probably be removed. If the update you make is only this change, then you can add [ci skip] to the commit title and description and it _should_ not run CI. I know this works for AppVeyor but GitHub Actions may require you to cancel the build manually. Again, I only suggest doing this if removing this comment is the only change in a new commit.

PR #620, opened 2021-09-15
AsyncWorkService ConfigAdmin Integration

achristoforides:
This PR integrates the asynchronous work service into ConfigAdmin.

The same pattern that was used in the DecarativeServices integration is being re-used for this integration into ConfigAdmin.


compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -167,9 +167,11 @@ namespace cmimpl {
 
 ConfigurationAdminImpl::ConfigurationAdminImpl(
   cppmicroservices::BundleContext context,
-  std::shared_ptr<cppmicroservices::logservice::LogService> lggr)
+  const std::shared_ptr<cppmicroservices::logservice::LogService>& lggr,
+  const std::shared_ptr<cppmicroservices::async::AsyncWorkService>& asyncWS)
```
achristoforides, line 171: Check to see if a move would be possible rather than const&

achristoforides: Logger is used in multiple locations in CMActivator. Moving the logger would invalidate the logger instance for CMActivator (which it still needs after creating the ConfigurationAdminImpl object).

Additionally, the AsyncWorkService shared_ptr is needed in the CMActivator class (even after creating the ConfigurationAdminImpl) object since it needs to call `StopTracking()` on it. At the moment, it doesn't but this is incorrect and will be added.

These shouldn't be changed to be moves.

jeffdiclemente (approved):

PR #626, opened 2021-11-05
Improve the performance of removing service listeners

jeffdiclemente:
This change is to improve the performance of removing service listeners as the set of service listeners grows. The use case of removing a service listener is most prevalent when destroying a service tracker object. Consider a scenario where a service tracker object is created and destroyed many times.


Jeeyuen-Rick (approved):
Reviewed test.

achristoforides (approved):

pelliott-mathworks (approved):

carneyweb (approved):

PR #630, opened 2021-12-09
[security layer] bundle validation mechanism

jeffdiclemente:


framework/src/util/Any.cpp
```diff
@@ -67,6 +67,21 @@ std::ostream& any_value_to_string(std::ostream& os, const Any& any)
...
+}
+
+std::ostream& any_value_to_json(std::ostream& os,
+                                const std::function<bool(const std::string&)>&,
+                                const uint8_t,
+                                const int32_t)
+{
+  return os;
+}
+
```
achristoforides, line 84: Are these intentionally implemented as doing nothing?

framework/src/util/SecurityException.cpp
```diff
@@ -0,0 +1,37 @@
...
+  : std::runtime_error(std::move(what)),
+    origin(std::move(origin))
+{}
+
+Bundle SecurityException::GetBundle() const
+{
+  return origin;
+}
+
+}
```
achristoforides, line 37: This is pedantic but this file does not have a final newline.

achristoforides (approved):

framework/src/util/Any.cpp
jeffdiclemente: yes.

framework/src/util/SecurityException.cpp
jeffdiclemente: I will fix this.

compendium/ConfigurationAdmin/src/CMLogger.cpp
```diff
@@ -36,8 +36,11 @@ CMLogger::CMLogger(cppmicroservices::BundleContext context)
...
 CMLogger::~CMLogger()
 {
-  if (serviceTracker) {
-    serviceTracker->Close();
+  try {
+    if (serviceTracker) {
+      serviceTracker->Close();
+    }
+  } catch (...) {
   }
```
carneyweb, line 44: Should we issue a warning to the log an exception gets thrown rather than just swallowing it?

carneyweb (approved):

compendium/ConfigurationAdmin/src/CMLogger.cpp
jeffdiclemente: There is no logger at this point to log to since it would have been removed on `serviceTracker->Close()`.

I can refactor `CMLogger` similar to `SCRLogger` to close the service tracker outside of the destructor to be able to log exceptions. I would prefer to make that change separate to this PR. This change is only necessary to fix a crash in the tests.

compendium/DeclarativeServices/src/SCRBundleExtension.cpp
```diff
@@ -96,18 +100,25 @@ SCRBundleExtension::SCRBundleExtension(
...
                 bundleContext.GetBundle().GetSymbolicName());
   for (auto& compManager : *managers) {
     auto fut = compManager->Disable();
     registry->RemoveComponentManager(compManager);
-    fut
-      .get(); // since this happens when the bundle is stopped. Wait until the disable is finished on the other thread.
+    try {
+      fut.get(); // since this happens when the bundle is stopped. Wait until the disable is finished on the other thread.
+    } catch (...) {
+    }
```
jeffdiclemente, line 118: add log for exception

compendium/DeclarativeServices/src/manager/BundleOrPrototypeComponentConfiguration.cpp
```diff
@@ -130,10 +144,27 @@ void BundleOrPrototypeComponentConfigurationImpl::DeactivateComponentInstance(
...
+    auto compMgrs = compManagerRegistry->GetComponentManagers(
+      registration.GetReference().GetBundle().GetBundleId());
+    std::for_each(
+      compMgrs.begin(),
+      compMgrs.end(),
+      [](std::shared_ptr<cppmicroservices::scrimpl::ComponentManager> compMgr) {
+        try {
+          compMgr->Disable().get();
+        } catch (...) {
+        }
```
jeffdiclemente, line 164: log exceptions to logger

compendium/DeclarativeServices/test/TestBundleValidation.cpp
```diff
@@ -0,0 +1,393 @@
...
+  ASSERT_NO_THROW(enabledFuture.get());
+
+  svcRef = f.GetBundleContext().GetServiceReference<test::Interface2>();
+  ASSERT_TRUE(svcRef);
+  ASSERT_THROW(auto svcObj = f.GetBundleContext().GetService(svcRef),
+               cppmicroservices::SecurityException);
+
+  compDesc = dsRuntimeService->GetComponentDescriptionDTO(*bundleIter, "sample::ServiceComponent6");
+  ASSERT_FALSE(dsRuntimeService->IsComponentEnabled(compDesc));
+  Interface1SvcReg.Unregister();
```
jeffdiclemente, line 130: why am i unregistering here?

compendium/DeclarativeServices/test/TestBundleValidation.cpp
```diff
@@ -0,0 +1,393 @@
...
+  // service component not to be enabled
+  auto compDesc = dsRuntimeService->GetComponentDescriptionDTO(
+    *bundleIter, "sample::ServiceComponent6");
+  ASSERT_FALSE(dsRuntimeService->IsComponentEnabled(compDesc));
+  ASSERT_TRUE(receivedSecondBundleValidationErrorEvent);
+
+  f.GetBundleContext().RemoveListener(std::move(token));
+  f.Stop();
+  f.WaitForStop(std::chrono::milliseconds::zero());
+}
```
pelliott-mathworks, line 393: Just wondering if it would be worthwhile to have one or more concurrent tests to see what happens when the validation function returns a failure or an exception when multiple threads are accessing it at the same time. Maybe one of the GetService failures since we know that's a common use case.

framework/test/gtest/BundleValidationTest.cpp
```diff
@@ -0,0 +1,141 @@
...
+  ASSERT_EQ(bundleA.GetState(),
+            cppmicroservices::Bundle::State::STATE_RESOLVED);
+  ASSERT_TRUE(receivedBundleValidationErrorEvent);
+
+  f.GetBundleContext().RemoveListener(std::move(token));
+  f.Stop();
+  f.WaitForStop(std::chrono::milliseconds::zero());
+}
+
+#endif
```
pelliott-mathworks, line 141: Same comment here as for the DS tests. Maybe a concurrent GetService test for failures and exceptions would be useful?

pelliott-mathworks (approved):
Great job. Very thorough! Two small comments/questions about the tests.

PR #632, opened 2021-12-20
Fix DS race condition caused by latch removal

pelliott-mathworks:
Removing the latch from CCActiveState caused a race condition because a Deactivate wasn't waiting for prior Activate calls to complete. Restored the latch. Signed-off-by: The MathWorks, Inc. <pelliott@mathworks.com>

### DIFF
--- a/compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
+++ b/compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
@@ -23,6 +23,7 @@
 #include "CCActiveState.hpp"
 #include "../ComponentConfigurationImpl.hpp"
 #include "CCUnsatisfiedReferenceState.hpp"
+#include "cppmicroservices/detail/ScopeGuard.h"
 #include "cppmicroservices/SharedLibraryException.h"
 
 namespace cppmicroservices {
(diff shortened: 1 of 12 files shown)


compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
```diff
@@ -102,64 +117,84 @@ void CCActiveState::DoDeactivateWork(ComponentConfigurationImpl& mgr)
...
+        service::component::runtime::dto::ComponentState::ACTIVE) {
+      auto logger = mgr.GetLogger();
+      logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_WARNING,
+                  "Modified failed. Component no longer in Active State.");
+      return false;
+    }
+    if (!mgr.ModifyComponentInstanceProperties()) {
+      // Component instance does not have a Modified method. Deactivate
+      // and reactivate
+      latch.CountDown();
```
jeffdiclemente, line 147: this should be wrapped in a `try/catch` to guarantee that deactivation happens even if the latch throws.

compendium/DeclarativeServices/src/manager/states/CCActiveState.hpp
```diff
@@ -82,9 +83,15 @@ class CCActiveState final : public CCSatisfiedState
    * state represented by this object
    */
   ComponentState GetValue() const override { return ComponentState::ACTIVE; }
+  void WaitForTransitionTask() override {
+      latch.Wait(); 
+  }
+  void WaitForStateChangeCompletion() override { return; }
```
jeffdiclemente, line 89: I don't understand the purpose of `WaitForStateChangeCompletion()`, can you elaborate? What is this function's purpose?

compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
```diff
@@ -34,51 +35,61 @@ std::shared_ptr<ComponentInstance> CCActiveState::Activate(
...
-  // Make sure the state didn't change while we were waiting
-  if (mgr.GetState()->GetValue() !=
-      service::component::runtime::dto::ComponentState::ACTIVE) {
-    auto logger = mgr.GetLogger();
-    logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
-                "Deactivate failed. Component no longer in Active State.");
-    return;
-  }
-  DoDeactivateWork(mgr);
+   DoDeactivateWork(mgr);
```
jeffdiclemente, line 92: It doesn't seem like there's a need for a `DoDeactivateWork()` function. `Deactivate` can be called instead.

pelliott-mathworks: DoDeactivateWork is also called by the Modified method

compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
pelliott-mathworks: Yes. Fixed.

compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
jeffdiclemente: Modified can call Deactivate. I don't see a need for DoDeactiveWork when it just calls Deactivate now.

compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
jeffdiclemente: this still needs to be wrapped in a `try/catch`.

jeffdiclemente (approved):

PR #634, opened 2021-12-23
Fix warning suppression leakage to non-CppMicroServices code inside translation units 

achristoforides:
### Description
GlobalConfig.h (before this change) has some code which unconditionally suppresses MSVC warning C4996. This on its own is fine but the issue is that if GlobalConfig.h is directly or indirectly included in any translation unit, then the warning suppression will not only suppress warnings coming from CppMicroServices, but other libraries and code as well.

### Brief overview of changes
- Removed call to US_MSVC_DISABLE_WARNING (the comment above it seems to indicate that we thought it was for "deprecated, unsafe functions" but in reality it's for all deprecated functions)
- Disabled C4996 for MSVC compilers older than 2017 (US_MSVC_PUSH_DISABLE_WARNING and corresponding pop do not seem to work)

### Concrete example of what these changes fix
Functions `__declspec(deprecated) void foo(int a)` and `[[deprecated]] void foobar(char c)` and `__attribute__((deprecated)) void bonjour(const std::string& msg)` exist. `foo` and `foobar` hypothetically exist inside of the CppMicroServices library and are marked as deprecated. `bonjour` exists in another DSO (not CppMicroServices and not a library that a developer wrote (e.g., some-cool-library)).

The header in CppMicroServices where `foo` and `foobar` are declared includes GlobalConfig.h (directly or indirectly) for some of the utilities that it provides.

A developer wishes to use `foo` and `foobar` in their implementation file and since warning C4996 is disabled, they will not receive a compilation warning/error about using deprecated functions. The same developer also wishes to use `bonjour` in that same implementation file (doesn't have to be the same file but merely the same translation unit). The developer didn't check the documentation of some-cool-library and doesn't know that `bonjour` is actually a deprecated function. They write code which calls `bonjour` and they do not receive a warning/error saying that the function is deprecated because of GlobalConfig.h being directly/indirectly included and globally disabling the warning.

This is undesired and these changes aim to fix this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -472,8 +472,10 @@ if(MSVC)
 
   if (${MSVC_VERSION} LESS 1910)
     # For googletest on MSVC versions less than MSVC 2017
-    # 'identifier' : decorated name length exceeded, name was truncated
-    set(disabled_warnings ${disabled_warnings} 4503)
+    # - (4503) 'identifier' : decorated name length exceeded, name was truncated
+    # Additionally, on MSVC versions less than MSVC 2015
+    # - (4996) all deprecation warnings (pragma push/pop for disablign the warning does not work)
+    set(disabled_warnings ${disabled_warnings} 4503 4996)
   endif()
 
   foreach(wd ${disabled_warnings})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -474,7 +474,7 @@ if(MSVC)
     # For googletest on MSVC versions less than MSVC 2017
     # - (4503) 'identifier' : decorated name length exceeded, name was truncated
     # Additionally, on MSVC versions less than MSVC 2015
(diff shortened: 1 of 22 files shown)


Jeeyuen-Rick (approved):

CMakeLists.txt
```diff
@@ -472,8 +472,10 @@ if(MSVC)
 
   if (${MSVC_VERSION} LESS 1910)
     # For googletest on MSVC versions less than MSVC 2017
-    # 'identifier' : decorated name length exceeded, name was truncated
-    set(disabled_warnings ${disabled_warnings} 4503)
+    # - (4503) 'identifier' : decorated name length exceeded, name was truncated
+    # Additionally, on MSVC versions less than MSVC 2015
+    # - (4996) all deprecation warnings (pragma push/pop for disablign the warning does not work)
```
jeffdiclemente, line 477: typo: disablign

jeffdiclemente (approved):

PR #641, opened 2022-01-13
Test case for concurrent DS GetService calls

jeffdiclemente:
This test is designed to find race conditions in Declarative Services when multiple threads all try to get the same service.

### DIFF
--- a/compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
+++ b/compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
@@ -52,14 +52,7 @@ std::shared_ptr<ComponentInstance> CCActiveState::Activate(
         }
       });
       std::lock_guard<std::mutex> lock(oneAtATimeMutex);
-      // Make sure the state didn't change while we were waiting
-      auto currentState = mgr.GetState();
-      if (currentState->GetValue() !=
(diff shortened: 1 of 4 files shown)


compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
```diff
@@ -52,14 +52,7 @@ std::shared_ptr<ComponentInstance> CCActiveState::Activate(
...
       std::lock_guard<std::mutex> lock(oneAtATimeMutex);
-      // Make sure the state didn't change while we were waiting
-      auto currentState = mgr.GetState();
-      if (currentState->GetValue() !=
-          service::component::runtime::dto::ComponentState::ACTIVE) {
-        logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_WARNING,
-                    "Activate failed. Component no longer in Active State.");
-        return nullptr;
-      }
+
```
pelliott-mathworks, line 55: Is it possible that while we're waiting for the lock here that the Modified method is running, resulting in Deactivate being called? If so, we might have to keep this check in because the state would have changed while we were waiting.

pelliott-mathworks (approved):
one question to answer

carneyweb (approved):

PR #642, opened 2022-01-21
update minimum required cmake version

### LINKED ISSUE #445: update minimum required CMake version in CMakeLists.txt
we currently have 3.2 set as the minimum version of CMake requried to build CppMicroServices however CMake features which require a more recent version are required.

update the minimum required CMake version to be inline with the CMake features currently used.

jeffdiclemente:
Fixes #445

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Extract the current version from the VERSION file
 file(STRINGS VERSION _version LIMIT_COUNT 1)
 
-set(US_CMAKE_MINIMUM_REQUIRED_VERSION 3.2)
+set(US_CMAKE_MINIMUM_REQUIRED_VERSION 3.12.4)
 
 cmake_minimum_required(VERSION ${US_CMAKE_MINIMUM_REQUIRED_VERSION})
 
--- a/doc/src/build_instructions.rst
+++ b/doc/src/build_instructions.rst
@@ -11,7 +11,7 @@ project or *Make* files.
 Prerequisites
 -------------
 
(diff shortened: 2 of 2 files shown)


pelliott-mathworks (approved):

achristoforides (approved):

CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ endforeach()
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_ENABLE_EXPORTS ON)
```
jeffdiclemente, line 37: as I understand it after reading https://cmake.org/cmake/help/latest/prop_tgt/ENABLE_EXPORTS.html, `CMAKE_ENABLE_EXPORTS` only affects executables. is that correct?

jeffdiclemente (commented):
If my understanding of https://cmake.org/cmake/help/latest/policy/CMP0065.html is correct, this only affects executables. In which case I would like to see `CMAKE_ENABLE_EXPORTS` enabled for individual CMake sub projects instead of globally. However this can be done later as a follow up task.

achristoforides:
You are correct that CMP0065 only affects executables. I initially tried looking for a solution to only target specific executables (potentially with set_target_property), but I was unable to find what exactly to set based on the CMake documentation.

I agree that this can be done later as a follow-up task as this was already the behavior before switching the minimum CMake version.

PR #643, opened 2022-01-25
Cppms patch

pelliott-mathworks:
According to OSGI, Update notifications are only supposed to be sent by ConfigurationAdmin when a configuration object has been updated. A configuration object which has not yet been Updated should not result in an Update Notification. Also, a Removed Notification may only be sent if an Updated notification was previously sent. There were several places in our ConfigurationAdmin code where Updated and Removed notifications were being sent to services in error. This change fixes this. 

Each Configuration object contains a changeCount. The changeCount is set equal to 1 when the Configuration object is created and incremented each time the Configuration object is Updated. Updated and Removed notifications will only be sent for Configuration objects with a changeCount > 1. The AddConfigurations method, adds Configuration objects for configurations read from the manifest.json file. These Configuration objects are added to the repository with changeCount = 2 since the construction of these objects is equivalent to a creation and an update.  

The Configuration object now has a method called IsUpdated which returns true for Configuration objects with a changeCount > 1 and false otherwise. 

ConfigurationAdminImpl::NotifyConfigurationUpdated now accepts a second input parameter called removeUpdate. This is true when the Configuration object has been removed but IsUpdated was true before the object was removed. NotifyConfigurationUpdate is used for both Updated and Removed operations. For a Removed operation the Configuration object has already been removed and IsUpdated cannot be called so this input parameter controls whether or not a Removed notification will be sent. 

ConfigurationAdmin has service trackers that monitor services that implement the ManagedService and ManagedServiceFactory interfaces. It receives an AddingService notification when one of these services is registered. ConfigurationAdmin is responsible for notifying services of Configuration objects that were added to the repository (and updated) prior to the service being registered. There was a problem with our code because the configuration repository was being changed while the AddingService methods were accessing the repository. The race condition resulted in two Updated notifications being sent for the same Update. This change fixes this problem by locking the configuration repository at the beginning of the AddingService method so that no changes can occur while the AddingService method is in process.

### DIFF
--- a/compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
+++ b/compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
@@ -386,6 +386,7 @@ std::vector<ConfigurationAddedInfo> ConfigurationAdminImpl::AddConfigurations(
   std::vector<ConfigurationAddedInfo> pidsAndChangeCountsAndIDs;
   std::vector<bool> createdOrUpdated;
   std::vector<std::shared_ptr<ConfigurationImpl>> configurationsToInvalidate;
+  unsigned long changeCount{ 0u };
   {
     std::lock_guard<std::mutex> lk{ configurationsMutex };
     for (const auto& configMetadata : configurationMetadata) {
@@ -394,16 +395,21 @@ std::vector<ConfigurationAddedInfo> ConfigurationAdminImpl::AddConfigurations(
       if (it == std::end(configurations)) {
         auto factoryPid = getFactoryPid(pid);
         AddFactoryInstanceIfRequired(pid, factoryPid);
+        auto newConfig = std::make_shared<ConfigurationImpl>(
+                           this,
+                           pid,
+                           std::move(factoryPid),
+                           std::move(configMetadata.properties));
+        // GetChangeCount can throw an exception if the configuration has
+        // been removed from the repository. No exception is possible here
+        // because the configuration object has not yet been added to the repository.
+        changeCount = newConfig->GetChangeCount();
         it = configurations
                .emplace(pid,
-                        std::make_shared<ConfigurationImpl>(
-                          this,
-                          pid,
-                          std::move(factoryPid),
-                          std::move(configMetadata.properties)))
+                        newConfig)
                .first;
         pidsAndChangeCountsAndIDs.emplace_back(
-          pid, it->second->GetChangeCount(), reinterpret_cast<std::uintptr_t>(it->second.get()));
+          pid, changeCount, reinterpret_cast<std::uintptr_t>(it->second.get()));
         createdOrUpdated.push_back(true);
         continue;
       }
@@ -412,6 +418,7 @@ std::vector<ConfigurationAddedInfo> ConfigurationAdminImpl::AddConfigurations(
         const auto updatedAndChangeCount =
           it->second->UpdateWithoutNotificationIfDifferent(
             configMetadata.properties);
+        changeCount = updatedAndChangeCount.second;
         pidsAndChangeCountsAndIDs.emplace_back(
           pid,
           std::get<1>(updatedAndChangeCount),
@@ -425,7 +432,7 @@ std::vector<ConfigurationAddedInfo> ConfigurationAdminImpl::AddConfigurations(
         it->second = std::make_shared<ConfigurationImpl>(
           this, pid, getFactoryPid(pid), std::move(configMetadata.properties));
         pidsAndChangeCountsAndIDs.emplace_back(
-          pid, it->second->GetChangeCount(), reinterpret_cast<std::uintptr_t>(it->second.get()));
+          pid, changeCount, reinterpret_cast<std::uintptr_t>(it->second.get()));
         createdOrUpdated.push_back(true);
       }
     }
--- a/compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
+++ b/compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
@@ -400,13 +400,10 @@ std::vector<ConfigurationAddedInfo> ConfigurationAdminImpl::AddConfigurations(
                            pid,
                            std::move(factoryPid),
                            std::move(configMetadata.properties));
-        // GetChangeCount can throw an exception if the configuration has
-        // been removed from the repository. No exception is possible here
-        // because the configuration object has not yet been added to the repository.
         changeCount = newConfig->GetChangeCount();
         it = configurations
                .emplace(pid,
-                        newConfig)
+                        std::move(newConfig))
                .first;
         pidsAndChangeCountsAndIDs.emplace_back(
           pid, changeCount, reinterpret_cast<std::uintptr_t>(it->second.get()));
--- a/compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
+++ b/compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
@@ -26,6 +26,7 @@
 
 #include "cppmicroservices/Constants.h"
 #include "cppmicroservices/cm/ConfigurationException.hpp"
+#include "cppmicroservices/detail/ScopeGuard.h"
 
 #include "CMConstants.hpp"
 #include "ConfigurationAdminImpl.hpp"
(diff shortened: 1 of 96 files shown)


compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
```diff
@@ -424,6 +424,8 @@ TEST_F(TestConfigurationAdminImpl, VerifyRemoveConfigurations)
 
   EXPECT_NO_THROW(configAdmin.RemoveConfigurations(std::move(result)));
 
+  configAdmin.WaitForAllAsync();
```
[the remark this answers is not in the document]
pelliott-mathworks: It's not needed. I thought a wait had to happen before lines 429 through 433 were executed but RemoveConfigurations does a WaitForAllAsync before returning

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -511,27 +523,38 @@ void ConfigurationAdminImpl::RemoveConfigurations(
 }
 
 std::shared_future<void> ConfigurationAdminImpl::NotifyConfigurationUpdated(
-  const std::string& pid)
+  const std::string& pid, bool removeUpdate)
```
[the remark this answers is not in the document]
pelliott-mathworks: yes. making the changes now.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -511,27 +523,38 @@ void ConfigurationAdminImpl::RemoveConfigurations(
...
       const auto it = configurations.find(pid);
       if (it == std::end(configurations)) {
         removed = true;
+        if (removeUpdate) {
+          isUpdated = true;
+        }
       } else {
         try {
           properties = it->second->GetProperties();
+          isUpdated = it->second->IsUpdated();
```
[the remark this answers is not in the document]
pelliott-mathworks: switched the order so isUpdated will be calculated first.

compendium/ConfigurationAdmin/src/ConfigurationImpl.cpp
```diff
@@ -46,6 +46,11 @@ ConfigurationImpl::ConfigurationImpl(ConfigurationAdminPrivate* configAdmin,
 {
   assert(configAdminImpl != nullptr &&
          "Invalid ConfigurationAdminPrivate pointer");
+  // constructing a configuration object with properties is the equivalent
+  // of a Create and an Update operation.
+  if (properties.size() > 0) {
+    changeCount = 2u;
```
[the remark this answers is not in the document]
pelliott-mathworks: yes

compendium/ConfigurationAdmin/src/ConfigurationImpl.hpp
```diff
@@ -118,6 +118,19 @@ class ConfigurationImpl final
    */
   void Invalidate() override;
 
+  /** Internal method used by {@code ConfigurationAdminImpl} to determine if a configuration
+   * object has been updated. 
+   *
+   * See {@code ConfigurationPrivate#Invalidate}
+   */
+  bool IsUpdated() override {
```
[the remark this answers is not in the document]
pelliott-mathworks: yes. Making the changes now.

compendium/ConfigurationAdmin/src/ConfigurationImpl.hpp
```diff
@@ -118,6 +118,19 @@ class ConfigurationImpl final
    */
   void Invalidate() override;
 
+  /** Internal method used by {@code ConfigurationAdminImpl} to determine if a configuration
+   * object has been updated. 
+   *
+   * See {@code ConfigurationPrivate#Invalidate}
+   */
+  bool IsUpdated() override { 
+    if (changeCount > 1) {
```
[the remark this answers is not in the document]
pelliott-mathworks: yes

compendium/ConfigurationAdmin/src/ConfigurationPrivate.hpp
```diff
@@ -67,6 +67,13 @@ class ConfigurationPrivate
    * ConfigurationAdminImpl is shutting down or when it is deleting a Configuration.
    */
   virtual void Invalidate() = 0;
+
+  /** Internal method used by {@code ConfigurationAdminImpl} to determine if a configuration
+   * object has been updated. 
+   *
+   * See {@code ConfigurationPrivate#Invalidate}
+   */
+  virtual bool IsUpdated() = 0;
```
achristoforides, line 76: This should also be updated to whatever you decide to change the name of this function to based on Jeff's suggestion.

compendium/ConfigurationAdmin/test/TestConfigAdmin.cpp
```diff
@@ -326,7 +333,7 @@ TEST_F(ConfigAdminTests, testServiceRemoved)
   configuration = m_configAdmin->GetConfiguration("cm.testservice");
   EXPECT_TRUE(configuration->GetProperties().empty());
   EXPECT_EQ(service->getCounter(), expectedCount);
-
+
```
achristoforides, line 336: Remove extra whitespace

compendium/ConfigurationAdmin/test/TestConfigurationAdminImpl.cpp
```diff
@@ -536,7 +538,7 @@ TEST_F(TestConfigurationAdminImpl, VerifyManagedServiceFactoryNotification)
   ASSERT_TRUE(conf);
   auto props = conf->GetProperties();
   props["foo"] = std::string{ "bar" };
-  EXPECT_NO_THROW(conf->Update(props));
+  EXPECT_NO_THROW(auto fut = conf->Update(props); fut.get());
```
achristoforides, line 541: You should surround the code with `{ ... }`. (Add curly braces)

achristoforides (changes requested):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -493,8 +507,10 @@ void ConfigurationAdminImpl::RemoveConfigurations(
   auto idx = 0u;
   for (const auto& pidAndChangeCountAndID : pidsAndChangeCountsAndIDs) {
     const auto& pid = pidAndChangeCountAndID.pid;
-    if (removed[idx]) {
-      NotifyConfigurationUpdated(pid);
+    if (removedAndUpdated[idx].first) {
+      if (removedAndUpdated[idx].second) {
```
jeffdiclemente, line 511: minor: this can be combined into one `if` statement

jeffdiclemente (changes requested):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
pelliott-mathworks: I don't think it can be. If removedAndUpdated[idx].first is true and removedAndUpdated[idx] is false we still have to log a debug message.

jeffdiclemente: I missed that. you're right.

compendium/ConfigurationAdmin/src/ConfigurationImpl.hpp
```diff
@@ -118,6 +118,30 @@ class ConfigurationImpl final
...
+  }
+  /** Internal method used by {@code ConfigurationAdminImpl} to get the value of the 
+   *  changeCount
+   *
+   * See {@code ConfigurationPrivate#GetChangeCount}
+   */
+  unsigned long GetChangeCount() override
+  {
+    std::lock_guard<std::mutex> lk{ propertiesMutex };
+    return changeCount;
```
jeffdiclemente, line 142: per the OSGi spec section 104.14.3.6, this function should thrown an exception if the configuration was removed. In our case you can use the standard `throw std::runtime_error(REMOVED_EXCEPTION_MESSAGE);`.

jeffdiclemente (changes requested):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -394,16 +395,21 @@ std::vector<ConfigurationAddedInfo> ConfigurationAdminImpl::AddConfigurations(
...
         AddFactoryInstanceIfRequired(pid, factoryPid);
+        auto newConfig = std::make_shared<ConfigurationImpl>(
+                           this,
+                           pid,
+                           std::move(factoryPid),
+                           std::move(configMetadata.properties));
+        // GetChangeCount can throw an exception if the configuration has
+        // been removed from the repository. No exception is possible here
+        // because the configuration object has not yet been added to the repository.
+        changeCount = newConfig->GetChangeCount();
```
jeffdiclemente, line 406: I'm not sure I understand the "why" of this comment. There is no way that `GetChangeCount()` can throw here. I don't understand why it's mentioned.

jeffdiclemente (changes requested):

carneyweb (approved):
Ship it!

jeffdiclemente (approved):

jeffdiclemente:
ConfigAdminTests.testServiceUpdated failed in a couple of the builds

jeffdiclemente (approved):

achristoforides (approved):

PR #651, opened 2022-03-29
Fix deadlock

jeffdiclemente:
### Problem
There are two problems fixed by this PR:
1. A deadlock that can occur due to re-entrancy while a `ManagedService/ManagedServiceFactory` `AddingService` callback is executing
2. A race condition which causes a second configuration object update notification to be sent to the `ManagedService/ManagedServiceFactory` when the Configuration object has not been modified.

Problem 1 happens when the `GetService` call in `AddingService` causes the start of a bundle containing a ManagedService/ManagedServiceFactory. This results in a deadlock on `std::lock_guard<std::mutex> lk{ configurationsMutex };` within `AddingService`.
Also, in general, we cannot hold locks while calling into user code and `GetService` can result in calling into user code.

Problem 2 happens because two or more threads race to send configuration object update notifications to the `ManagedService/ManagedServiceFactory` object. By maintaining a change count within the `TrackedServiceWrapper` object which is the config object's change count at the last time the `ManagedService/ManagedServiceFactory` object received a notification, multiple threads can check to see if a config update needs to happen. If the change count for the config object is greater than the change count at the last time the `ManagedService/ManagedServiceFactory` object received an update, then a notification should happen, otherwise another thread won the race and no update is necessary.


compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -684,21 +684,24 @@ ConfigurationAdminImpl::AddingService(
...
-    const auto serviceID = std::to_string(cppmicroservices::any_cast<long>(
-      reference.GetProperty(Constants::SERVICE_ID)));
-    logger->Log(
-      SeverityLevel::LOG_WARNING,
-      "Ignoring ManagedService with ID " + serviceID + " from bundle " +
-        (bundle ? bundle.GetSymbolicName() : "Unknown") +
-        " as it does not have a service.pid or component.name property");
-    return nullptr;
+  std::string pid;
+  {
```
carneyweb, line 688: Perhaps a comment explaining that the additional level of scoping here is to limit where the lock is applied?

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -783,20 +789,22 @@ ConfigurationAdminImpl::AddingService(
   const ServiceReference<cppmicroservices::service::cm::ManagedServiceFactory>&
     reference)
 {
+  std::string pid;
+  {
```
carneyweb, line 793: Same here.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -684,21 +684,24 @@ ConfigurationAdminImpl::AddingService(
...
-    const auto bundle = reference.GetBundle();
-    const auto serviceID = std::to_string(cppmicroservices::any_cast<long>(
-      reference.GetProperty(Constants::SERVICE_ID)));
-    logger->Log(
-      SeverityLevel::LOG_WARNING,
-      "Ignoring ManagedService with ID " + serviceID + " from bundle " +
-        (bundle ? bundle.GetSymbolicName() : "Unknown") +
-        " as it does not have a service.pid or component.name property");
-    return nullptr;
+  std::string pid;
```
carneyweb, line 687: Perhaps a comment here explaining the extra block to scope the lock and avoid a deadlock? Just thinking about future maintainers.

carneyweb (approved):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -684,21 +684,24 @@ ConfigurationAdminImpl::AddingService(
...
+        reference.GetProperty(Constants::SERVICE_ID)));
+      logger->Log(
+        SeverityLevel::LOG_WARNING,
+        "Ignoring ManagedService with ID " + serviceID + " from bundle " +
+          (bundle ? bundle.GetSymbolicName() : "Unknown") +
+          " as it does not have a service.pid or component.name property");
+      return nullptr;
+    }
   }
   auto managedService = cmContext.GetService(reference);
```
pelliott-mathworks, line 706: I think releasing the lock here may result in the double Updated race condition happening again. If another thread creates and updates the configuration object while the GetService is happening, then an Updated notification will be sent for that update and the AddingService code will send another one at the end of the AddingService method (line 751)  To avoid the second Updated notification, AddingService can only send notifications for configuration objects that were created and updated before the AddingService method was called. One solution might be to move the check about whether or not the configuration object exists and has been updated  into the locked code before the lock is released for the GetService call.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -707,8 +710,11 @@ ConfigurationAdminImpl::AddingService(
                 "obtained from the BundleContext");
     return nullptr;
   }
-  // Ensure there's a Configuration for this PID if one doesn't exist already.
+  // Lock the configurations repository so no configuration objects can be
+  // added or removed while AddingService is processing the new service.
+  std::lock_guard<std::mutex> lk{ configurationsMutex };
 
+  // Ensure there's a Configuration for this PID if one doesn't exist already.
```
pelliott-mathworks, line 717: I'm not sure we should be creating a configuration object here. There are two ways that valid configuration objects are created. A configuration object can appear in a manifest.json file (in that case it is added by AddConfigurations and an Updated notification will be sent at that time.) The other way is in user code when a user gets a configuration object and then updates it. This code here is creating a blank configuration object which is meaningless. Unless the configuration object is updated by a user or updated by AddConfigurations, the configuration object has no value. It doesn't hurt to add it here. It just might not be necessary.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -783,20 +789,22 @@ ConfigurationAdminImpl::AddingService(
...
+        reference.GetProperty(Constants::SERVICE_ID)));
+      logger->Log(
+        SeverityLevel::LOG_WARNING,
+        "Ignoring ManagedServiceFactory with ID " + serviceID +
+          " from bundle " + (bundle ? bundle.GetSymbolicName() : "Unknown") +
+          " as it does not have a service.pid or component.name property");
+      return nullptr;
+    }
   }
   auto managedServiceFactory = cmContext.GetService(reference);
```
pelliott-mathworks, line 809: same comments here as for ManagedService::AddingService

pelliott-mathworks (changes requested):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
jeffdiclemente: FYI, after locally modifying the `ConfigAdminTests.testDuplicateUpdated ` test to introduce multiple threads and running in a loop, I was able to get a double update on the config object:

```
Note: Google Test filter = ConfigAdminTests.testDuplicateUpdated
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from ConfigAdminTests
[ RUN      ] ConfigAdminTests.testDuplicateUpdated
D:\Dev\CppMicroServices\cppms-ds-deadlock\CppMicroServices\compendium\ConfigurationAdmin\test\TestConfigAdmin.cpp(553): error: Expected equality of these values:
  serviceFactory->getUpdatedCounter("cm.testfactory~0")
    Which is: 2
  1
```
so this change definitely regresses the fix made in commit 664f5f93479c31f4c2d4592145002ac54fe69a67

jeffdiclemente: side note;

after removing the scoping of the lock in the `AddingService` ServiceTracker callbacks (essentially reverting back to the same state as 664f5f93479c31f4c2d4592145002ac54fe69a67), I get this failure after running many iterations of the test:

```
Repeating all tests (iteration 340042) . . .

Note: Google Test filter = ConfigAdminTests.testDuplicateUpdated
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from ConfigAdminTests
[ RUN      ] ConfigAdminTests.testDuplicateUpdated
D:\Dev\CppMicroServices\cppms-ds-deadlock\CppMicroServices\compendium\ConfigurationAdmin\test\TestConfigAdmin.cpp(553): error: Expected equality of these values:
  serviceFactory->getUpdatedCounter("cm.testfactory~0")
    Which is: 0
  1
```

achristoforides (approved):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
jeffdiclemente: to confirm, there is an pre-existing race condition that causes the config update to be missed which manifest when running ComfigAdminTests.testConcurrentDuplicateUpdated in a loop and observing the following test failure:

```
[ RUN      ] ConfigAdminTests.testDuplicateUpdated
D:\Dev\CppMicroServices\cppms-ds-deadlock\CppMicroServices\compendium\ConfigurationAdmin\test\TestConfigAdmin.cpp(553): error: Expected equality of these values:
  serviceFactory->getUpdatedCounter("cm.testfactory~0")
    Which is: 0
  1
```

This race happens when the configuration object has not been created/updated and the service tracker adding the ManagedServiceFactory service has not yet added the TrackedServiceWrapper object to the internal tracking map.  The `Configuration::UpdateIfDifferent` gets all the tracked ManagedServiceFactory objects and if it hasn't been added yet AND the ServiceTracker::AddingService completes before the configuration object is created, then the config update is missed completely.

carneyweb (approved):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -734,35 +739,50 @@ ConfigurationAdminImpl::AddingService(
...
       // Configuration is being removed
+      logger->Log(SeverityLevel::LOG_WARNING,
+                  "Attempted to update a configuration which has been removed.",
+                  std::current_exception());
     }
     // Only send notifications for configuration objects that have been
     // Updated. 
     if (it->second->HasBeenUpdatedAtLeastOnce()) {
       PerformAsync([this, pid, managedService, properties] {
         notifyServiceUpdated(pid, *managedService, properties, *logger);
```
pelliott-mathworks, line 768: Is this bypassing the new mechanism in the managedServiceWrapper by calling notifyServiceUpdated directly the change count in the managedServiceWrapper doesn't get checked or incremented. I think there might be other cases in the code where notifyServiceUpdated and notifyServiceRemoved are called directly.

jeffdiclemente: managedServiceWrapper doesn't exist at this point since `AddingService` hasn't returned yet. So I don't think it's bypassing the new mechanism. When the managedServiceWrapper is returned, it will contain the change count at the point in which notifyServiceUpdated was called.

pelliott-mathworks (approved):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
jeffdiclemente: opened issue #716 to track this.

PR #656, opened 2022-04-08
Fixes #653, Ensure that any BundleContext functions do not segfault if the bundle context is invalid

### LINKED ISSUE #653: BundleContextPrivate::CheckValid() segfaults if BundleContext is retrieved from Framework after shutdown
# Description

The call to `BundleContextPrivate::CheckValid()` segfaults when the `BundleContext` is retrieved from `Framework::GetBundleContext()` after the framework has been shutdown via `Stop()/WaitForStop()`.

# Desired Behavior

As the documentation states for `BundleContext`, if a function is called on the `BundleContext` (notably `GetService()`, `GetServiceReference()`, and `GetServiceReferences()` (amongst others)) and the `BundleContext` is no longer valid, it should throw a `std::runtime_error`.

Rather than segfaulting, `BundleContextPrivate::CheckValid()` should properly throw a `std::runtime_error`.

# Reproduction Steps

```cpp
TEST(ServiceReferenceTestStandalone,
     TestGetServiceReferenceAfterFrameworkShutdown)
{
  auto framework = std::make_shared<cppmicroservices::Framework>(FrameworkFactory().NewFramework());
  ASSERT_TRUE(framework);
  framework->Start();

  auto context = framework->GetBundleContext();

  ASSERT_TRUE(context);

  framework->Stop();
  framework->WaitForStop(std::chrono::milliseconds::zero());
  
  auto context2 = framework->GetBundleContext();
  ASSERT_FALSE(context2);
  EXPECT_THROW({ // NOTE: This fails. Instead of throwing, it segfaults.
      (void)context2.GetServiceReference<ServiceNS::ITestServiceA>(); // NOTE: if type is void, it does not segfault
    }, std::runtime_error);
}
```

achristoforides:
- Fixes bug where any functions called on an invalid BundleContext()
  (if retrieved after the framework was shutdown) causes a segfault.

- Fixes bug where BundleContext::GetServiceObjects() did not throw
  if the bundle context was invalid.

- Added new test for verifying that BundleContext functions properly
  throw instead of segfaulting when the bundle context is invalid.

### DIFF
--- a/framework/src/bundle/BundleContext.cpp
+++ b/framework/src/bundle/BundleContext.cpp
@@ -74,12 +74,20 @@ BundleContext& BundleContext::operator=(std::nullptr_t)
 
 std::shared_ptr<detail::LogSink> BundleContext::GetLogSink() const
 {
+  if (!d) {
+    throw std::runtime_error("The bundle context is no longer valid");
+  }
+
   d->CheckValid();
   return d->bundle->coreCtx->sink->shared_from_this();
 }
 
 Any BundleContext::GetProperty(const std::string& key) const
 {
+  if (!d) {
+    throw std::runtime_error("The bundle context is no longer valid");
+  }
+
   d->CheckValid();
   auto b = (d->Lock(), d->bundle);
 
@@ -94,6 +102,10 @@ Any BundleContext::GetProperty(const std::string& key) const
 
 AnyMap BundleContext::GetProperties() const
 {
+  if (!d) {
+    throw std::runtime_error("The bundle context is no longer valid");
+  }
+
   d->CheckValid();
   auto b = (d->Lock(), d->bundle);
 
@@ -107,6 +119,10 @@ AnyMap BundleContext::GetProperties() const
 
 Bundle BundleContext::GetBundle() const
 {
+  if (!d) {
+    throw std::runtime_error("The bundle context is no longer valid");
+  }
+
   d->CheckValid();
   auto b = (d->Lock(), d->bundle);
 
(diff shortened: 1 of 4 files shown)


framework/src/bundle/BundleContext.cpp
```diff
@@ -74,12 +74,20 @@ BundleContext& BundleContext::operator=(std::nullptr_t)
...
+  }
+
   d->CheckValid();
   return d->bundle->coreCtx->sink->shared_from_this();
 }
 
 Any BundleContext::GetProperty(const std::string& key) const
 {
+  if (!d) {
+    throw std::runtime_error("The bundle context is no longer valid");
```
jeffdiclemente, line 88: The doxygen comment in `BundleContext.h` needs to be updated to indicate this now throws.

framework/src/bundle/BundleContext.cpp
```diff
@@ -94,6 +102,10 @@ Any BundleContext::GetProperty(const std::string& key) const
 
 AnyMap BundleContext::GetProperties() const
 {
+  if (!d) {
+    throw std::runtime_error("The bundle context is no longer valid");
```
jeffdiclemente, line 106: The doxygen comment in `BundleContext.h` needs to be updated to indicate this now throws.

jeffdiclemente (changes requested):
the doxygen in `BundleContext.h` for `template<class S>
  std::vector<ServiceReference<S>> GetServiceReferences(
    const std::string& filter = std::string())` says that it returns `std::logic_error` if the `BundleContext` is invalid. This is inconsistent and needs to be changed in the doxygen comment (or changed in the code if it throws a `std::logic_error`)

pelliott-mathworks (approved):

PR #663, opened 2022-05-04
Fix ListConfigurations returning empty config objects

jeffdiclemente:
Per OSGi standard, ListConfigurations should only return config objects containing properties (i.e. not an empty config).

### DIFF
--- a/compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
+++ b/compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
@@ -69,7 +69,7 @@ void ConfigurationManager::Initialize()
     for (const auto& pid : metadata->configurationPids) {
       if (configProperties.find(pid) == configProperties.end()) {
         auto config = configAdmin->ListConfigurations("(pid=" + pid + ")");
-        if (config.size() > 0) {
+        if (config.size() > 0 && config.front()->GetChangeCount() > 0) {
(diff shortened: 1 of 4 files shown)


pelliott-mathworks (approved):

pelliott-mathworks (commented):
Just wondering if you noticed one of the build configurations failed. The BuildAndTestOnWIndows

jeffdiclemente:
yes, I'm looking into it.

jeffdiclemente:
missing part of the fix. apparently I wrote the test but didn't actually fix the bug... oops.

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
```diff
@@ -69,7 +69,7 @@ void ConfigurationManager::Initialize()
     for (const auto& pid : metadata->configurationPids) {
       if (configProperties.find(pid) == configProperties.end()) {
         auto config = configAdmin->ListConfigurations("(pid=" + pid + ")");
-        if (config.size() > 0) {
+        if (config.size() > 0 && config.front()->GetChangeCount() > 0) {
```
achristoforides, line 72: Everything after `&&` is not necessary. ListConfigurations will not return configurations that are empty.

achristoforides (changes requested):

PR #664, opened 2022-05-06
Fix check-then-act race in GetServiceFromFactory

achristoforides:

### DIFF
--- a/framework/src/service/ServiceReferenceBase.cpp
+++ b/framework/src/service/ServiceReferenceBase.cpp
@@ -103,8 +103,9 @@ Bundle ServiceReferenceBase::GetBundle() const
 
   auto l = p->registration->Lock();
   US_UNUSED(l);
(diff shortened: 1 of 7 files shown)


framework/src/service/ServiceReferenceBase.cpp
```diff
@@ -103,9 +103,9 @@ Bundle ServiceReferenceBase::GetBundle() const
 
   auto l = p->registration->Lock();
   US_UNUSED(l);
-  if (p->registration->bundle == nullptr)
+  if (p->registration->bundle.lock() == nullptr)
     return Bundle();
```
jeffdiclemente, line 107: please add braces, even for single line if statements

jeffdiclemente (changes requested):

pelliott-mathworks (approved):

jeffdiclemente (approved):

PR #665, opened 2022-05-07
Fix check-then-act race for BundleContext

achristoforides:

### DIFF
--- a/framework/src/bundle/BundleUtils.cpp
+++ b/framework/src/bundle/BundleUtils.cpp
@@ -80,8 +80,9 @@ namespace cppmicroservices {
 // Private util function to return system bundle's log sink
 std::shared_ptr<detail::LogSink> GetFrameworkLogSink()
 {
+  auto bundle_ = GetPrivate(GetBundleContext())->bundle.lock();
   // The following is a hack, we need a cleaner solution in the future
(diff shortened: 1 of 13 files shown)


pelliott-mathworks (approved):

framework/src/bundle/BundleUtils.cpp
```diff
@@ -80,8 +80,9 @@ namespace cppmicroservices {
 // Private util function to return system bundle's log sink
 std::shared_ptr<detail::LogSink> GetFrameworkLogSink()
 {
+  auto bundle_ = GetPrivate(GetBundleContext())->bundle.lock();
   // The following is a hack, we need a cleaner solution in the future
-  return GetPrivate(GetBundleContext())->bundle->coreCtx->sink;
+  return bundle_->coreCtx->sink;
```
achristoforides, line 85: Need if check.

framework/src/service/ServiceReferenceBase.cpp
```diff
@@ -103,9 +103,9 @@ Bundle ServiceReferenceBase::GetBundle() const
 
   auto l = p->registration->Lock();
   US_UNUSED(l);
-  if (p->registration->bundle == nullptr)
+  if (p->registration->bundle.lock() == nullptr)
     return Bundle();
```
achristoforides, line 107: Put in curly brackets.

PR #667, opened 2022-05-11
Fix bug in FindResources() for data-only bundles

achristoforides:
When a bundle is installed with the optional manifest argument, the bundle isn't fopen'd automatically. In the case of data-only bundles, it is never fopen'd and the BundleResourceContainer is never initialized.

A call to FindResources() on a data-only bundle who was installed via `InstallBundles()` with the optional manifest will result in 0 resources always being returned.

### DIFF
--- a/framework/src/bundle/BundleResourceContainer.cpp
+++ b/framework/src/bundle/BundleResourceContainer.cpp
@@ -65,7 +65,7 @@ BundleResourceContainer::BundleResourceContainer(
       m_SortedToplevelDirs.insert(b.first);
     }
   } else {
-    OpenContainer();
+    OpenAndInitializeContainer();
   }
 }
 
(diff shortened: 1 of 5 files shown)


framework/src/bundle/BundleResourceContainer.cpp
```diff
@@ -265,11 +260,17 @@ bool BundleResourceContainer::Matches(const std::string& name,
   return true;
 }
 
-void BundleResourceContainer::OpenContainer()
+void BundleResourceContainer::OpenContainer() const
```
jeffdiclemente, line 263: this no longer only opens the container. I suggest a method name change to describe the intent. suggestion: `OpenAndInitializeContainer`

jeffdiclemente (approved):

pelliott-mathworks (approved):

PR #672, opened 2022-05-13
Add using statement to Binders.hpp

### LINKED ISSUE #567: Include order of Binders.hpp matters
The include order of "Binders.hpp" matters as it relies on a using statement which is not present in the file itself. It tries using "ComponentContext" but this does not exist in the file and is currently relying on other includes before it to define it.

This manifests itself in a build error when you switch around the include statements order in "ComponentInstanceImpl.hpp".

shane-riley:
Add a using statement to "Binders.hpp" to remove dependence on include order.

Fixes #567


pelliott-mathworks (approved):

carneyweb (approved):

achristoforides (approved):
Great job!

jeffdiclemente (approved):

PR #680, opened 2022-06-15
Integrate LogService into the core framework and add more detail to exception messsages

achristoforides:
This PR integrates the LogService into the core framework for logging messages pertaining to shared library loading. As part of this, the shared library that was created for LogService has been removed and instead has been converted into a header-only component, similar to cm.

More "events" or "actions" can be logged in the core framework but this PR only instruments the code for logging library loading.

Additionally, this PR adds more details to many of the exception messages thrown from BundlePrivate. In almost all exception messages, the bundle ID and bundle location are now included as part of the exception message.


jeffdiclemente (approved):

compendium/DeclarativeServices/src/manager/BundleLoader.cpp
```diff
@@ -78,7 +100,15 @@ GetComponentCreatorDeletors(const std::string& compName,
...
+                  "Loading shared library for Bundle #" +
+                    ToString(fromBundle.GetBundleId()) +
+                    " (location=" + bundleLoc + ")");
       sh.Load(any_cast<int>(opts));
+      logger->Log(logservice::SeverityLevel::LOG_INFO,
+                  "Finished loading shared library for Bundle #" +
+                    ToString(fromBundle.GetBundleId()) +
+                    " (location=" + bundleLoc + ")");
     } catch (const std::system_error& ex) {
       // SharedLibrary::Load() will throw a std::system_error when a shared library
```
jiechenma, line 113: It helps to also add a log when exception happens, something like "Failed loading shared library for ..."

achristoforides: Thanks for the comment. This code was already merged so I will add a task for myself for a follow-up submission with this added.

PR #681, opened 2022-06-15
Fix leaked file handle

jeffdiclemente:
A file handle can be leaked if a file is opened with embedded ZIP data in a layout CppMicroServices does not expect.

### DIFF
--- a/framework/src/bundle/BundleResourceContainer.cpp
+++ b/framework/src/bundle/BundleResourceContainer.cpp
@@ -245,16 +245,18 @@ bool BundleResourceContainer::Matches(const std::string& name,
                                       const std::string& filePattern) const
 {
   // short-cut
-  if (filePattern == "*")
+  if (filePattern == "*") {
     return true;
+  }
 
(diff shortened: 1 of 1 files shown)


achristoforides (approved):

achristoforides:
Should a test be added (similar to those in framework/test/gtest/OpenFileHandleTest.cpp) to validate that for a shared library which is loaded, if no zip is detected, that the number of file handles does not increase but rather is identical to the starting number of file handles?

jeffdiclemente:
> Should a test be added (similar to those in framework/test/gtest/OpenFileHandleTest.cpp) to validate that for a shared library which is loaded, if no zip is detected, that the number of file handles does not increase but rather is identical to the starting number of file handles?

yes, it would be worthwhile to have. A specific file would have to be created for the test, namely a file which has embedded ZIP data but not in the format CppMicroServices expects. I will create a follow on task and a separate submission for that.

pelliott-mathworks (approved):

PR #688, opened 2022-06-19
Upgrade miniz to v3.0 (unreleased, using git commit on master)

### LINKED ISSUE #592: upgrade to latest miniz implementation
evaluate if we should upgrade and if so, upgrade to the latest version of miniz: https://github.com/richgel999/miniz/releases/tag/2.1.0

achristoforides:
Fixes #592 .

The difference between the miniz version we were using and the v3.0 version (unreleased) is quite large so the cppms-specific changes for miniz were integrated where appropriate.

This version (not the v2.2 released version) already has the utf8 fixes included :)


jeffdiclemente (commented):
I have a couple concerns:

- This is an unreleased version, how do we know it is "production ready" and should be integrated in CppMicroServices from a code quality perspective?
- Were any of the cppmicroservices google benchmark tests run (e.g. `framework\test\bench\bundleinstall.cpp`) to gauge whether there are any performance regressions?

jeffdiclemente (approved):

PR #690, opened 2022-06-20
Add log messages when shared library loading throws an exception

achristoforides:

### DIFF
--- a/compendium/DeclarativeServices/src/manager/BundleLoader.cpp
+++ b/compendium/DeclarativeServices/src/manager/BundleLoader.cpp
@@ -110,6 +110,10 @@ GetComponentCreatorDeletors(
                     ToString(fromBundle.GetBundleId()) +
                     " (location=" + bundleLoc + ")");
     } catch (const std::system_error& ex) {
+      logger->Log(logservice::SeverityLevel::LOG_INFO,
+                  "Failed loading shared library for Bundle #" +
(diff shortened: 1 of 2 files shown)


compendium/DeclarativeServices/src/manager/BundleLoader.cpp
```diff
@@ -110,6 +110,10 @@ GetComponentCreatorDeletors(
                     ToString(fromBundle.GetBundleId()) +
                     " (location=" + bundleLoc + ")");
     } catch (const std::system_error& ex) {
+      logger->Log(logservice::SeverityLevel::LOG_INFO,
```
jeffdiclemente, line 113: I think we want to use the `Log` overload that takes an `exception_ptr` so that the exception is also logged.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -507,6 +507,9 @@ std::exception_ptr BundlePrivate::Start0()
       res = std::make_exception_ptr(cppmicroservices::SharedLibraryException(
         ex.code(), ex.what(), thisBundle));
     } catch (...) {
+      coreCtx->logger->Log(logservice::SeverityLevel::LOG_INFO,
```
jeffdiclemente, line 510: I think we want to use the Log overload that takes an exception_ptr so that the exception is also logged.

jeffdiclemente (approved):

PR #694, opened 2022-06-24
Add support for null values to AnyMap

pelliott-mathworks:
rapidjson supports null values but AnyMap did not. This change adds support for null values to the AnyMap class. Signed-off-by: The MathWorks, Inc. <pelliott@mathworks.com>

### DIFF
--- a/framework/include/cppmicroservices/AnyMap.h
+++ b/framework/include/cppmicroservices/AnyMap.h
@@ -28,6 +28,22 @@
 #include <string>
 #include <unordered_map>
 
+/* rapidjson supports null values in a .json file. The AnyMap class does not support null values. 
+ * A value saved in the AnyMap must have a valid type and null is not a valid type. The NullValue 
+ * object allows us to save a null value in the AnyMap. 
+ */
+namespace cppmicroservices {
+  struct NullValue
+  {
+    std::string ToString() { return ""; };
+    friend std::ostream& operator<<(std::ostream& os, const NullValue& nv)
+    {
+      return os << ""
+                << "\n";
+    }
+  };
+}
+
 namespace cppmicroservices {
 
 namespace detail {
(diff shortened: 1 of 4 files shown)


framework/include/cppmicroservices/AnyMap.h
```diff
@@ -28,6 +28,22 @@
 #include <string>
 #include <unordered_map>
 
+/* rapidjson supports null values in a .json file. The AnyMap class does not support null values. 
+ * A value saved in the AnyMap must have a valid type and null is not a valid type. The NullValue 
+ * object allows us to save a null value in the AnyMap. 
+ */
+namespace cppmicroservices {
+  struct NullValue
```
jeffdiclemente, line 36: should we have a function in the `Any` class to test is an `Any` is `NullValue`?

Currently, I don't think `Any::Empty()` and `Any::ToJSON()` are in alignment with this new concept of `NullValue`. There might also be other `Any` member functions which are not consistent with `NullValue`.

```
  /**
   * returns true if the Any is empty
   */
  bool Empty() const { return !_content; }


  /**
   * Returns a JSON representation for the content.
   *
   * Custom types should specialize the any_value_to_json template function for meaningful output.
   * The values of increment and indent are passed around to be able to be used for nicer
   * formatting. The code that makes use of this is in the any_value_to_json specializations for the
   * various containers.
   *
   * To get pretty output, simply pass a value greater than zero in as the first argument of ToJSON
   * and the rest of the code will take care of things.
   *
   * @param increment The amount of extra indentation to add for each level of JSON. An increment of
   *                  zero indicates no special formatting
   * @param indent    The current amount of indent to apply to the current line. 
   */
  std::string ToJSON(const uint8_t increment, const int32_t indent) const
  {
    return Empty() ? "null" : _content->ToJSON(increment, indent);
  }
```

pelliott-mathworks: Do you think Empty should return true if the value is null?

achristoforides (approved):

jeffdiclemente (changes requested):
I would like to see test coverage for `NullValue` within the core framework's unit tests, i.e. within framework\test\gtest\AnyTest.cpp. Take a look at the existing tests in here for guidance on what to test. off the top of my head, testing Any::Type() for `NullValue` should be done.

carneyweb (approved):

PR #695, opened 2022-06-27
Remove manual ref counting for BundleResource

achristoforides:

### DIFF
--- a/framework/src/bundle/BundleResource.cpp
+++ b/framework/src/bundle/BundleResource.cpp
@@ -129,13 +129,6 @@ void BundleResource::InitializeChildren()
   }
 }
 
-BundleResource& BundleResource::operator=(const BundleResource& resource)
-{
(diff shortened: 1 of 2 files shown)


framework/src/bundle/BundleResource.cpp
```diff
@@ -137,20 +129,9 @@ void BundleResource::InitializeChildren()
...
 {
-  BundleResourcePrivate* curr_d = d;
   d = resource.d;
-  ++d->ref;
-
-  if (!--curr_d->ref)
-    delete curr_d;
 
   return *this;
 }
```
jeffdiclemente, line 137: can the copy assignment operator be `=default`? I don't think we need a user-defined implementation for this now.

pelliott-mathworks (approved):

jeffdiclemente (approved):

PR #697, opened 2022-07-01
Disable code signing for bundle with no c++ code

achristoforides:
- Without this, cannot build on ARM macOS machines

### DIFF
--- a/framework/test/bundles/libTestModuleWithEmbeddedZip/CMakeLists.txt
+++ b/framework/test/bundles/libTestModuleWithEmbeddedZip/CMakeLists.txt
@@ -12,6 +12,11 @@ set(_srcs foo.cpp)
 
 add_library(${PROJECT_NAME} ${_srcs})
 
+set_target_properties(${PROJECT_NAME}
+  PROPERTIES
+  XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY ""
+)
+
 if(CMAKE_CROSSCOMPILING)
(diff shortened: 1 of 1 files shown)


framework/test/bundles/libTestModuleWithEmbeddedZip/CMakeLists.txt
```diff
@@ -12,6 +12,11 @@ set(_srcs foo.cpp)
 
 add_library(${PROJECT_NAME} ${_srcs})
 
+set_target_properties(${PROJECT_NAME}
+  PROPERTIES
+  XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY ""
+)
```
jeffdiclemente, line 18: I should have added this initially. I missed this since I didn't use the generic cmake function `usFunctionEmbedResources` to embed data because I would have had to modify `usFunctionEmbedResources` to fit my needs and didn't think it necessary for this one use case. 

I recommend scoping this to macOS only, see `cmake/usFunctionEmbedResources.cmake`:

```
# Disable code-signing on macOS if appending resources.
	if(APPLE)
	  set_target_properties(${US_RESOURCE_TARGET} PROPERTIES
		XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "")
	endif()
```

jeffdiclemente (changes requested):

pelliott-mathworks (approved):

jeffdiclemente (approved):

PR #698, opened 2022-07-01
Multiple services and factory services in bundle dependent on same configuration pid

pelliott-mathworks:
Multiple services and service factories can be dependent on the same configuration object and should be notified if the configuration object changes. Right now only one will be notified. This pull request fixes this problem in Configuration Admin for any service implementing the ManagedService interface or the ManagedServiceFactory interface. All services implementing those interfaces should be notified when configuration objects on which they are dependent change.

### DIFF
--- a/compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
+++ b/compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
@@ -602,12 +602,15 @@ std::shared_future<void> ConfigurationAdminImpl::NotifyConfigurationUpdated(
     std::for_each(managedServiceWrappers.begin(),
                   managedServiceWrappers.end(),
                   [&](const auto& managedServiceWrapper) {
-                  if (managedServiceWrapper->pid == pid) {
-                    notifyServiceUpdated(pid,
-                                      *(managedServiceWrapper->trackedService),
-                                      properties,
-                                      *logger);
-                  }
+                    // The ServiceTracker will return a default constructed shared_ptr for each ManagedService
+                    // that we aren't tracking. We must be careful not to dereference these!
+                    if ((managedServiceWrapper) && (managedServiceWrapper->pid == pid)) {
+                        notifyServiceUpdated(
+                          pid,
+                          *(managedServiceWrapper->trackedService),
+                          properties,
+                          *logger);
+                    }
                   });
 
     const auto factoryPid = getFactoryPid(pid);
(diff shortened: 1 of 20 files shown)


compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -599,37 +599,41 @@ std::shared_future<void> ConfigurationAdminImpl::NotifyConfigurationUpdated(
...
+           // that we aren't tracking. We must be careful not to dereference these!
+           return (managedServiceWrapper ? (pid == managedServiceWrapper->pid)
+                                         : false);
+         })) != managedServiceWrappers.end()) {
       const auto& managedServiceWrapper = *it;
       notifyServiceUpdated(
-        pid, *(managedServiceWrapper->trackedService), properties, *logger);
+          pid, *(managedServiceWrapper->trackedService), properties, *logger);
+       it++;
     }
```
jeffdiclemente, line 617: is there a better way to do this that uses the C++ standard algorithms (`#include <algorithm>`)? This is iterating over all managed services and also running `std::find_if` for each iteration. Using `std::for_each` eliminates the manual iterator manipulation and reduces the time complexity,.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -599,37 +599,41 @@ std::shared_future<void> ConfigurationAdminImpl::NotifyConfigurationUpdated(
...
         // that we aren't tracking. We must be careful not to dereference these!
         return (managedServiceFactoryWrapper
                   ? (factoryPid == managedServiceFactoryWrapper->pid)
                   : false);
-      });
-    if (factoryIt != std::end(managedServiceFactoryWrappers)) {
+      })) != managedServiceFactoryWrappers.end()) {
       const auto& managedServiceFactoryWrapper = *factoryIt;
       if (removed) {
         notifyServiceRemoved(
```
jeffdiclemente, line 639: same comment as above.

jeffdiclemente (changes requested):

compendium/ConfigurationAdmin/test/TestConfigAdmin.cpp
```diff
@@ -94,13 +94,39 @@ std::shared_ptr<::test::TestManagedServiceInterface> getManagedService(
...
 
+std::vector<std::shared_ptr<::test::TestManagedServiceInterface>>
+getManagedServices(cppmicroservices::BundleContext& ctx)
+{
+  auto srs = ctx.GetServiceReferences<::test::TestManagedServiceInterface>();
+  std::vector<std::shared_ptr<::test::TestManagedServiceInterface>> services(
+    srs.size());
+  for (uint32_t i = 0; i < services.size(); ++i) {
+    services[i] = ctx.GetService<::test::TestManagedServiceInterface>(srs[i]);
+  }
```
jeffdiclemente, line 105: recommend using a range based for loop

compendium/ConfigurationAdmin/test/TestConfigAdmin.cpp
```diff
@@ -94,13 +94,39 @@ std::shared_ptr<::test::TestManagedServiceInterface> getManagedService(
...
 
+std::vector<std::shared_ptr<::test::TestManagedServiceFactory>>
+getManagedServiceFactories(cppmicroservices::BundleContext& ctx)
+{
+  auto srs = ctx.GetServiceReferences<::test::TestManagedServiceFactory>();
+  std::vector<std::shared_ptr<::test::TestManagedServiceFactory>> factories(
+    srs.size());
+  for (uint32_t i = 0; i < factories.size(); ++i) {
+    factories[i] = ctx.GetService<::test::TestManagedServiceFactory>(srs[i]);
+  }
```
jeffdiclemente, line 125: recommend using a range based for loop

compendium/test_bundles/TestBundleMultipleManagedService/src/ManagedServiceImpl.cpp
```diff
@@ -0,0 +1,49 @@
...
+namespace cm {
+namespace test {
+
+TestManagedServiceImpl3::TestManagedServiceImpl3()
+  : m_counter{ 0 }
+{}
+
+TestManagedServiceImpl3::~TestManagedServiceImpl3() = default;
+
+void TestManagedServiceImpl3::Updated(AnyMap const& properties)
```
jeffdiclemente, line 17: look out for function parameter values which aren't used in the function body. this will cause a build error on linux/mac

jeffdiclemente (changes requested):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
pelliott-mathworks: Replaced code with a for_each loop. Please review again when you get a chance.

jeffdiclemente (approved):

PR #699, opened 2022-07-02
Fix LDAP performance for matching [DO NOT MERGE]

achristoforides:
# Performance Summary
(NOTE: The benchmark being ran is `MatchFilterWithBundle/Complex`)
| Platform | Build Config | Time/CPU (ns) (dev) | Time/CPU (ns) (optimized) | Iterations | ~Speedup |
| --- | --- | --- | --- | --- | --- |
| win64 | Debug | `22779\|22741` | `6118\|6090` | `32` | `~3.72x` |
| win64 | Release | `1988\|1984` | `481\|481` | `32` | `~4.13x` |
| maca64 | Debug | `6005\|6004`| `3157\|3157` |  `32` | `~1.90x` |
| maca64 | Release | `723\|723` | `320\|320` | `32` | `~2.25x` |
| glnxa64 | Release | `721\|714` | `380\|378` | `32 ` | `~1.89x` |

# Benchmark Machine Info
### Win64 Benchmarking Details:
```
Run on (12 X 3600 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 1024 KiB (x6)
  L3 Unified 8448 KiB (x1)
```

### Maca64 Benchmarking Details (benchmarks performed on an M1 Pro chip):
(NOTE: According to google benchmark output, values for CPU speed are wrong due to an incompatibility with M1 Macs)
```
Run on (10 X 24.1001 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x10)
  L1 Instruction 128 KiB (x10)
  L2 Unified 4096 KiB (x5)
```

### Glnxa64 Benchmarking Details:
```
Run on (8 X 2600 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x8)
  L1 Instruction 64 KiB (x8)
  L2 Unified 512 KiB (x8)
  L3 Unified 16384 KiB (x8)
```

### DIFF
--- a/framework/src/service/ServiceRegistrationBase.cpp
+++ b/framework/src/service/ServiceRegistrationBase.cpp
@@ -151,11 +151,11 @@ void ServiceRegistrationBase::SetProperties(const ServiceProperties& props)
 
     auto propsCopy(props);
     propsCopy[Constants::SERVICE_ID] =
-      d->properties.Value_unlocked(Constants::SERVICE_ID);
-    objectClasses = d->properties.Value_unlocked(Constants::OBJECTCLASS);
+      d->properties.Value_unlocked(Constants::SERVICE_ID).first;
+    objectClasses = d->properties.Value_unlocked(Constants::OBJECTCLASS).first;
     propsCopy[Constants::OBJECTCLASS] = objectClasses;
     propsCopy[Constants::SERVICE_SCOPE] =
-      d->properties.Value_unlocked(Constants::SERVICE_SCOPE);
+      d->properties.Value_unlocked(Constants::SERVICE_SCOPE).first;
 
     auto itr = propsCopy.find(Constants::SERVICE_RANKING);
     if (itr != propsCopy.end()) {
@@ -169,13 +169,14 @@ void ServiceRegistrationBase::SetProperties(const ServiceProperties& props)
       }
     }
 
-    auto oldRankAny = d->properties.Value_unlocked(Constants::SERVICE_RANKING);
+    auto oldRankAny =
+      d->properties.Value_unlocked(Constants::SERVICE_RANKING).first;
     if (!oldRankAny.Empty()) {
       // since the old ranking is extracted from existing service properties
       // stored in the service registry, no need to type check before casting
       old_rank = any_cast<int>(oldRankAny);
     }
-    d->properties = Properties(std::move(AnyMap(std::move(propsCopy))));
+    d->properties = Properties(AnyMap(std::move(propsCopy)));
   }
   if (old_rank != new_rank) {
     auto classes = any_cast<std::vector<std::string>>(objectClasses);
--- a/framework/src/service/ServiceRegistrationBase.cpp
+++ b/framework/src/service/ServiceRegistrationBase.cpp
@@ -148,6 +148,7 @@ void ServiceRegistrationBase::SetProperties(const ServiceProperties& props)
 
     auto l2 = d->properties.Lock();
     US_UNUSED(l2);
+
     auto propsCopy(props);
     propsCopy[Constants::SERVICE_ID] =
       d->properties.Value_unlocked(Constants::SERVICE_ID);
@@ -174,7 +175,7 @@ void ServiceRegistrationBase::SetProperties(const ServiceProperties& props)
       // stored in the service registry, no need to type check before casting
       old_rank = any_cast<int>(oldRankAny);
     }
-    d->properties = Properties(std::move(propsCopy));
+    d->properties = Properties(std::move(AnyMap(std::move(propsCopy))));
   }
   if (old_rank != new_rank) {
     auto classes = any_cast<std::vector<std::string>>(objectClasses);
--- a/framework/src/util/AnyMap.cpp
+++ b/framework/src/util/AnyMap.cpp
@@ -309,48 +309,6 @@ any_map::const_iter::iterator any_map::const_iter::operator++(int)
   return tmp;
 }
 
-any_map::const_iter::iterator& any_map::const_iter::operator--()
-{
-  switch (type) {
-    case ORDERED:
-      --o_it();
-      break;
-    case UNORDERED:
-      --uo_it();
-      break;
-    case UNORDERED_CI:
-      --uoci_it();
-      break;
-    case NONE:
-      throw std::logic_error("cannot decrement an invalid iterator");
-    default:
-      throw std::logic_error("invalid iterator type");
-  }
-
-  return *this;
-}
-
-any_map::const_iter::iterator any_map::const_iter::operator--(int)
-{
-  iterator tmp = *this;
-  switch (type) {
-    case ORDERED:
-      o_it()--;
-      break;
-    case UNORDERED:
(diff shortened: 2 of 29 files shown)


framework/src/util/Properties.cpp
```diff
@@ -24,100 +24,105 @@
...
 US_MSVC_PUSH_DISABLE_WARNING(4996)
 
 namespace cppmicroservices {
 
 const Any Properties::emptyAny;
 
+Properties::Properties(AnyMap& p)
+  : props(p)
+{
+  if (p.size() > static_cast<std::size_t>(std::numeric_limits<int>::max())) {
```
pelliott-mathworks, line 39: Just wondering if this check is needed. p is a reference to an AnyMap which already has a size limit.

framework/src/util/PropsCheck.h
```diff
@@ -0,0 +1,67 @@
+/*=============================================================================
```
pelliott-mathworks, line 1: Our last SW at Google book club discussed documentation. Maybe with brand new files we could consider increasing our documentation? Some comments to consider, file comment, class comments and function comments.

framework/src/util/AnyMap.cpp
```diff
@@ -575,18 +575,36 @@ any_map::any_map(const ordered_any_map& m)
   map.o = new ordered_any_map(m);
 }
 
+any_map::any_map(ordered_any_map&& m)
+  : type(map_type::ORDERED_MAP)
+{
+  map.o = new ordered_any_map(std::move(m));
```
carneyweb, line 581: std::move is unnecessary as m is already a &&

framework/src/util/AnyMap.cpp
```diff
@@ -575,18 +575,36 @@ any_map::any_map(const ordered_any_map& m)
...
 any_map::any_map(const unordered_any_map& m)
   : type(map_type::UNORDERED_MAP)
 {
   map.uo = new unordered_any_map(m);
 }
 
+any_map::any_map(unordered_any_map&& m)
+  : type(map_type::UNORDERED_MAP)
+{
+  map.uo = new unordered_any_map(std::move(m));
```
carneyweb, line 593: std::move is unnecessary as m is already a &&

framework/src/util/AnyMap.cpp
```diff
@@ -575,18 +575,36 @@ any_map::any_map(const ordered_any_map& m)
...
 any_map::any_map(const unordered_any_cimap& m)
   : type(map_type::UNORDERED_MAP_CASEINSENSITIVE_KEYS)
 {
   map.uoci = new unordered_any_cimap(m);
 }
 
+any_map::any_map(unordered_any_cimap&& m)
+  : type(map_type::UNORDERED_MAP_CASEINSENSITIVE_KEYS)
+{
+  map.uoci = new unordered_any_cimap(std::move(m));
```
carneyweb, line 605: std::move is unnecessary as m is already a &&

pelliott-mathworks (approved):

framework/src/service/ServiceRegistrationBase.cpp
```diff
@@ -168,13 +169,14 @@ void ServiceRegistrationBase::SetProperties(const ServiceProperties& props)
...
-    auto oldRankAny = d->properties.Value_unlocked(Constants::SERVICE_RANKING);
+    auto oldRankAny =
+      d->properties.Value_unlocked(Constants::SERVICE_RANKING).first;
     if (!oldRankAny.Empty()) {
       // since the old ranking is extracted from existing service properties
       // stored in the service registry, no need to type check before casting
       old_rank = any_cast<int>(oldRankAny);
     }
-    d->properties = Properties(std::move(propsCopy));
+    d->properties = Properties(AnyMap(std::move(propsCopy)));
```
carneyweb, line 179: Is calling the AnyMap constructor explicitly required here?

framework/src/util/AnyMap.cpp
carneyweb, line 581: std::move is unnecessary since m is already a &&

framework/src/util/Properties.cpp
```diff
@@ -24,100 +24,105 @@
...
+
+#include "PropsCheck.h"
 
 US_MSVC_PUSH_DISABLE_WARNING(4996)
 
 namespace cppmicroservices {
 
 const Any Properties::emptyAny;
 
+Properties::Properties(AnyMap& p)
```
carneyweb, line 36: Should this be a const&? Or perhaps a &&? A reference implies that the values can be changed, which should not happen with the way it's coded, right?

framework/src/util/Properties.cpp
carneyweb, line 39: Why is this check here? Not that it's bad, but have we ever run across an AnyMap with more keys than INT_MAX?

framework/src/util/Properties.cpp
carneyweb: I now see the other versions of the constructor below... I think this constructor should probably be removed.

framework/src/util/AnyMap.cpp
achristoforides: This was necessary for test failures on macOS.

framework/src/util/AnyMap.cpp
achristoforides: This was necessary for test failures on macOS.

framework/src/util/AnyMap.cpp
achristoforides: This was necessary for test failures on macOS.

framework/src/util/AnyMap.cpp
achristoforides: This was necessary for test failures on macOS.

framework/src/util/Properties.h
```diff
@@ -36,24 +36,22 @@ class Properties : public detail::MultiThreaded<>
 {
 
 public:
+  explicit Properties(AnyMap& props);
```
carneyweb, line 39: Is this constructor necessary? If it doesn't actually modify props, it can be removed... the const& constructor will be called instead.

carneyweb (approved):

framework/src/util/Properties.cpp
achristoforides: Fixed.

framework/src/util/PropsCheck.h
achristoforides: Added.

framework/src/util/Properties.cpp
achristoforides: Fixed.

framework/src/util/Properties.cpp
achristoforides: Removed.

framework/src/util/Properties.h
achristoforides: Not necessary, removed.

carneyweb (approved):

pelliott-mathworks (approved):

PR #703, opened 2022-07-17
Improve service dependency error logging

kkumar45:
This PR improves service dependency error logging by enabling DEBUG log to show the cause of error when a service fails to find it's dependency because of typo/misspelled in it's LDAP filter or the service properties.

Signed-off-by: The MathWorks, Inc. kunalk@mathworks.com

### DIFF
--- a/compendium/DeclarativeServices/test/CMakeLists.txt
+++ b/compendium/DeclarativeServices/test/CMakeLists.txt
@@ -78,6 +78,7 @@ set(_declarativeservices_tests
   TestRegistrationManager.cpp
   TestSCRBundleExtension.cpp
   TestServiceComponentRuntimeImpl.cpp
+  TestServiceDependencyErrorLogging.cpp
   TestServiceMetadataParserV1.cpp
   TestSetConfiguration.cpp
   TestSingletonComponentConfiguration.cpp
@@ -102,6 +103,14 @@ include_directories(${PROJECT_BINARY_DIR}/include
   ${CppMicroServices_BINARY_DIR}/compendium/CM/include
   )
 
+set(_logservice_additional_srcs
+  ${CppMicroServices_SOURCE_DIR}/compendium/LogServiceImpl/src/LogServiceImpl.cpp
+  )
+
+set(_logservice_additional_hdrs
+  ${CppMicroServices_SOURCE_DIR}/compendium/LogServiceImpl/src/LogServiceImpl.hpp
+  )
+
 set(_additional_srcs
   TestUtils.cpp
   ../src/SCRActivator.cpp
@@ -123,7 +132,9 @@ usFunctionGetResourceSource(
 
 add_executable(${us_declarativeservices_test_exe_name}
   ${_declarativeservices_tests}
-  ${_additional_srcs})
+  ${_additional_srcs}
+  ${_logservice_additional_srcs}
+  ${_logservice_additional_hdrs})
 
 if (US_COMPILER_MSVC AND BUILD_SHARED_LIBS)
   target_compile_options(${us_declarativeservices_test_exe_name} PRIVATE -DGTEST_LINKED_AS_SHARED_LIBRARY)
@@ -140,7 +151,9 @@ set_property(
   )
 
 target_include_directories(${us_declarativeservices_test_exe_name}
-  PRIVATE $<TARGET_PROPERTY:util,INCLUDE_DIRECTORIES>)
+  PRIVATE $<TARGET_PROPERTY:util,INCLUDE_DIRECTORIES>
+  ${CppMicroServices_SOURCE_DIR}/compendium/LogServiceImpl/src
+  ${CppMicroServices_SOURCE_DIR}/third_party/spdlog/include)
 
 # Disable deprecation warnings.
 if (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") OR
--- a/compendium/DeclarativeServices/test/CMakeLists.txt
+++ b/compendium/DeclarativeServices/test/CMakeLists.txt
@@ -103,14 +103,6 @@ include_directories(${PROJECT_BINARY_DIR}/include
   ${CppMicroServices_BINARY_DIR}/compendium/CM/include
   )
 
-set(_logservice_additional_srcs
-  ${CppMicroServices_SOURCE_DIR}/compendium/LogServiceImpl/src/LogServiceImpl.cpp
-  )
-
-set(_logservice_additional_hdrs
-  ${CppMicroServices_SOURCE_DIR}/compendium/LogServiceImpl/src/LogServiceImpl.hpp
-  )
-
 set(_additional_srcs
   TestUtils.cpp
   ../src/SCRActivator.cpp
@@ -132,9 +124,7 @@ usFunctionGetResourceSource(
 
 add_executable(${us_declarativeservices_test_exe_name}
   ${_declarativeservices_tests}
-  ${_additional_srcs}
-  ${_logservice_additional_srcs}
-  ${_logservice_additional_hdrs})
+  ${_additional_srcs})
 
 if (US_COMPILER_MSVC AND BUILD_SHARED_LIBS)
   target_compile_options(${us_declarativeservices_test_exe_name} PRIVATE -DGTEST_LINKED_AS_SHARED_LIBRARY)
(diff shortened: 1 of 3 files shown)


compendium/DeclarativeServices/test/CMakeLists.txt
```diff
@@ -102,6 +103,14 @@ include_directories(${PROJECT_BINARY_DIR}/include
   ${CppMicroServices_BINARY_DIR}/compendium/CM/include
   )
 
+set(_logservice_additional_srcs
```
jeffdiclemente, line 106: in general I'd like to avoid having a dependency on the log service implementation code. I think you can write tests without it by using fake/mock log service objects

framework/src/service/ServiceListeners.cpp
```diff
@@ -446,6 +446,15 @@ void ServiceListeners::GetMatchingServiceListeners(const ServiceEvent& evt,
...
       if (ldapExpr.IsNull() || ldapExpr.Evaluate(props, false)) {
         set.insert(sse);
+      } else {
+        if (evt.GetType() == ServiceEvent::SERVICE_REGISTERED) {
+          const std::string msg =
+            sse.GetBundleContext().GetBundle().GetSymbolicName() +
+            " reference LDAP filter doesn't match " +
+            ref.GetBundle().GetSymbolicName() + " component's properties.\n";
+          coreCtx->logger->Log(
+            cppmicroservices::logservice::SeverityLevel::LOG_DEBUG, msg);
```
jeffdiclemente, line 456: I would like to see the debug message be more descriptive about the reason why someone is seeing the debug message. From a users standpoint, I don't understand what this debug message is telling me.

framework/src/service/ServiceListeners.cpp
```diff
@@ -446,6 +446,15 @@ void ServiceListeners::GetMatchingServiceListeners(const ServiceEvent& evt,
       const LDAPExpr& ldapExpr = sse.GetLDAPExpr();
       if (ldapExpr.IsNull() || ldapExpr.Evaluate(props, false)) {
         set.insert(sse);
+      } else {
+        if (evt.GetType() == ServiceEvent::SERVICE_REGISTERED) {
```
jeffdiclemente, line 450: should this be an else if instead of an else containing an if statement?

jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/CMakeLists.txt
kkumar45: removed log service impl dependency

framework/src/service/ServiceListeners.cpp
kkumar45: modified

framework/src/service/ServiceListeners.cpp
kkumar45: added verbose details

framework/src/service/ServiceListeners.cpp
```diff
@@ -446,7 +446,18 @@ void ServiceListeners::GetMatchingServiceListeners(const ServiceEvent& evt,
...
         set.insert(sse);
-      }
+      } else if (evt.GetType() == ServiceEvent::SERVICE_REGISTERED) {
+        const std::string msg =
+          "References LDAP filter value in key 'target' of " +
+          sse.GetBundleContext().GetBundle().GetSymbolicName() +
+          " manifest doesn't match service reference properties'." +
+          "\nPlease verify ldap filter value '" + sse.GetLDAPExpr().ToString() +
+          "' in " + ref.GetBundle().GetSymbolicName() +
+          " manifest components 'properties' map.";
```
jeffdiclemente, line 456: the use of the word "_target_" is specific to declarative services and this log message could be logged for services that are not using declarative services. It would make more sense to the user if the debug message was more generic in terms of what it's trying to tell the user. e.g. describe instead that a service listener matching the LDAP filter could not be found using the LDAP filter _X_.

I'm also not sure the bundle symbolic name is relevant for this debug message. I would suggest leaving it out of the debug message.

jeffdiclemente (changes requested):

framework/src/service/ServiceListeners.cpp
```diff
@@ -446,7 +446,11 @@ void ServiceListeners::GetMatchingServiceListeners(const ServiceEvent& evt,
       const LDAPExpr& ldapExpr = sse.GetLDAPExpr();
       if (ldapExpr.IsNull() || ldapExpr.Evaluate(props, false)) {
         set.insert(sse);
-      }
+      } else if (evt.GetType() == ServiceEvent::SERVICE_REGISTERED) {
+        const std::string msg = "Service listener matching LDAP filter could not be found using LDAP filter: " + sse.GetFilter();
+          coreCtx->logger->Log(
+            cppmicroservices::logservice::SeverityLevel::LOG_DEBUG, msg);
```
jeffdiclemente, line 452: thanks, this is much easier to understand. 
The service properties, in the `props` variable, should also be logged to show users the properties compared against when the LDAP filter failed to match.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/TestServiceDependencyErrorLogging.cpp
```diff
@@ -0,0 +1,155 @@
...
+private:
+  std::shared_ptr<CustomLogger> logger_;
+
+
+public:
+  cppmicroservices::Framework framework;
+};
+
+
+TEST_F(ServiceDependencyErrorLoggingTest, TestServiceDependencyLDAPFilter)
```
pelliott-mathworks, line 92: This seems like a complicated test for this new code. It creates several mock objects including a SingletonComponentConfigurationImpl object. Just wondering if a simpler test would have been sufficient. GetMatchingServiceListeners is called from ServiceRegistrationBase::SetProperties, ServiceRegistrationBase::Unregister and ServiceRegistrationBase::CreateServiceProperties. Just wondering if a test using a ServiceRegistrationBase object would have been sufficient. It would be simpler.

framework/src/service/ServiceListeners.cpp
```diff
@@ -446,7 +446,24 @@ void ServiceListeners::GetMatchingServiceListeners(const ServiceEvent& evt,
...
+          propsKeyValues += k + " : " + val + ", ";
+        }
+
+        const std::string msg =
+          "Service listener matching LDAP filter could "
+          "not be found using LDAP filter: " +
+          sse.GetFilter() + ". Please verify the LDAP filter value in service properties: " + "{" + propsKeyValues + "}";
+
+          coreCtx->logger->Log(
+            cppmicroservices::logservice::SeverityLevel::LOG_DEBUG, msg);
```
pelliott-mathworks, line 465: There are many cases where not finding the matching service listener is the normal case. Just wondering if this is the right place to be logging this. The normal cases will be cluttering up the log. Would it make more sense to have the caller of GetMatchingServiceListeners do the logging? They are the ones that know when not finding a match is normal or not.

pelliott-mathworks: I think it's normal for someone to use an LDAP filter to search for services with a specific service property. It's not an error if they don't succeed in finding a service. It just means the service hasn't been registered yet. Wouldn't this logging code log those cases as well?

jeffdiclemente: yes, it will log in those cases. the problem is, we don't know what the intent is. the goal is to log something for users to see in the log files when it's the case that they didn't find the service they expected due to a typo in a LDAP filter and/or service property. the downside is that there are more log entries.

compendium/DeclarativeServices/test/TestServiceDependencyErrorLogging.cpp
kkumar45: `mockMetadata` and `refmockMetadata` creates dummy metadata values needed for the use case. i.e creating non-matching LDAP filter value to enable logging. `SingletonComponentConfigurationImpl` creates registration & reference manager that automatically handle the service registration in case of dependencies by triggering the call to Register at the appropriate time. 
I looked into `ServiceRegistrationBase` and it looks like it doesn't handle dependencies tracking and doesn't have a logger to verify what information is logged. We may have to handle all those cases manually.

pelliott-mathworks: Jeff sent me a link to the Solution Review document for this feature [Solution Review : Improve service dependency error logging](https://confluence.mathworks.com/display/I18N/Solution+Review+%3A+Improve+service+dependency+error+logging)
That helped me understand what problem you are trying to solve and why the complexity is necessary. Looks good.

pelliott-mathworks (approved):

jeffdiclemente (approved):

achristoforides:
Hi @kkumar45 , merging this into development completely broke all builds on CI due to compilation failures. Can you please take a look? I reverted the merge.

jeffdiclemente:
what was missed? this PR had no CI failures.

kkumar45:
Looks like my branch was not synced with recent commits wherein `Properties::Value_unlocked` definition is changed recently to return pair `std::pair<Any, bool> Properties::Value_unlocked`. I will sync and modify my changes accordingly.

PR #704, opened 2022-07-18
Improve performance of LDAP matching

achristoforides:
This PR improves the performance of LDAP matching.

The performance improvements in https://github.com/CppMicroServices/CppMicroServices/pull/699 are characteristic of the performance gains exhibited by this PR except the performance gains in this PR are in-fact faster (by ~80ns on win64 when benchmarked on my machine, and generally faster on the other platforms).

### DIFF
--- a/framework/include/cppmicroservices/AnyMap.h
+++ b/framework/include/cppmicroservices/AnyMap.h
@@ -221,8 +221,11 @@ public:
 
   any_map(map_type type);
   any_map(const ordered_any_map& m);
+  any_map(ordered_any_map&& m);
   any_map(const unordered_any_map& m);
+  any_map(unordered_any_map&& m);
   any_map(const unordered_any_cimap& m);
+  any_map(unordered_any_cimap&& m);
 
   any_map(const any_map& m);
   any_map& operator=(const any_map& m);
--- a/framework/include/cppmicroservices/AnyMap.h
+++ b/framework/include/cppmicroservices/AnyMap.h
@@ -29,6 +29,8 @@
 #include <unordered_map>
 
 namespace cppmicroservices {
+class Properties;
+class LDAPExpr;
 
 namespace detail {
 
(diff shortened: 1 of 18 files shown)


framework/src/util/Properties.cpp
```diff
@@ -24,99 +24,140 @@
...
+Properties::Properties(const AnyMap& p)
+  : props(p)
 {
-  keys = std::move(o.keys);
-  values = std::move(o.values);
-  return *this;
+  if (p.GetType() != AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS) {
+    props_check::ValidateAnyMap(p);
+
+    PopulateCaseInsensitiveLookupMap();
```
achristoforides, line 62: Do the validation check as part of PopulateCaseInsensitiveLookupMap() (if faster...)

framework/src/util/PropsCheck.cpp
```diff
@@ -0,0 +1,75 @@
...
+ * those stored in Property objects have.
+ * 
+ * The provided AnyMap is said to be valid if there exists no pairs of two keys
+ * which differ in case only (e.g., "service.feature", "Service.feature"). If this
+ * condition is not true, this function throws as defined below.
+ * 
+ * @param am The AnyMap to validate
+ * @throws std::runtime_error Thrown when `am` is invalid (described above)
+ */
+void ValidateAnyMap(const cppmicroservices::AnyMap& am)
```
achristoforides, line 42: Check to see if using a std::unordered_set (case-insensitive) is a better solution for checking validity.

framework/src/util/Properties.cpp
```diff
@@ -24,99 +24,140 @@
...
+
+    if (!matchCase) {
+      auto ciItr = caseInsensitiveLookup.find(key);
+      if (ciItr != caseInsensitiveLookup.end()) {
+        return std::make_pair(
+          props.findUO_TypeChecked(ciItr->second.data())->second, true);
+      } else {
+        return std::make_pair(emptyAny, false);
+      }
     }
```
achristoforides, line 123: This needs to happen in the ORDERED_MAP case.

framework/src/util/Properties.cpp
```diff
@@ -24,99 +24,140 @@
...
+          props.findUOCI_TypeChecked(ciItr->second.data())->second, true);
+      } else {
+        return std::make_pair(emptyAny, false);
+      }
+    }
+
+    return std::make_pair(emptyAny, false);
+  } else {
+    throw std::runtime_error("Unknown AnyMap type.");
   }
```
achristoforides, line 145: This logic should be for ORDERED_MAP

framework/include/cppmicroservices/AnyMap.h
```diff
@@ -221,8 +223,11 @@ class US_Framework_EXPORT any_map
 
   any_map(map_type type);
```
pelliott-mathworks, line 224: Some additional comments to explain some of the background and reasoning for the decisions that were made to improve efficiency would be helpful.

framework/src/util/PropsCheck.cpp
achristoforides: It's not.

jeffdiclemente (approved):

PR #706, opened 2022-07-19
CFRlogger accessviolation

pelliott-mathworks:
A couple of small bugs were detected in CFRLogger and some inconsistencies with the way CFRLogger was being initialized in CoreBundleContext.cpp, as compared to other Service Trackers (like ServiceHook) so this PR is addressing those issues. 

In the existing code, the CoreBundleContext::Init method initialized the logger as follows:
logger = std::make_shared<cppmicroservices::cfrimpl::CFRLogger>(MakeBundleContext(systemBundle->bundleContext.Load().get()));
Every time the CoreBundleContext::Init method ran it would destroy the old CFRLogger object and create a new one. This was inefficent. Also I think the intention was to create one CFRLogger object when CoreBundleContext was constructed and destroy it when it was destructed. 

To fix this, Open and Close methods were introduced. The CFRLogger is constructed when the CoreBundleContext object is constructed, opened when the CoreBundleContext::Init method is run and closed when the CoreBundleContext::Uninit0 method is run. It is destroyed when the CoreBundleContext object is destroyed. Only one object exists for the lifetime of CoreBundleContext. This implementation is consistent with the ServiceHooks implementation

### DIFF
--- a/framework/src/bundle/CoreBundleContext.h
+++ b/framework/src/bundle/CoreBundleContext.h
@@ -98,14 +98,7 @@ public:
   */
   std::shared_ptr<detail::LogSink> sink;
 
-  /**
-   * A LogService for logging framework messages via
-   * a default or user-provided LogService that are intended to be
-   * visible outside of the framework.
-   */
-  std::shared_ptr<cppmicroservices::cfrimpl::CFRLogger> logger;
-
-  /**
+   /**
    * Bundle Storage
    */
   std::unique_ptr<BundleStorage> storage;
@@ -124,6 +117,13 @@ public:
    * All registered services in this framework.
    */
   ServiceRegistry services;
+
+  /**
+   * A LogService for logging framework messages via
+   * a default or user-provided LogService that are intended to be
+   * visible outside of the framework.
+   */
+  std::shared_ptr<cppmicroservices::cfrimpl::CFRLogger> logger;
 
   /**
    * All service hooks.
--- a/framework/src/bundle/CoreBundleContext.h
+++ b/framework/src/bundle/CoreBundleContext.h
@@ -70,6 +70,18 @@ class CoreBundleContext
 {
 public:
   /**
+  * Please note: The order of the member variables in this class is important. When the 
+  * CoreBundleContext object is destroyed, it will call the destructors for the member 
+  * variables in the reverse order in which they are listed here. For example, serviceHooks 
+  * will be destroyed before the logger. The logger will be destroyed before the listeners, etc. 
+  * 
+  * The logger has a ServiceTracker member variable. When it is destroyed, the ServiceTracker::Close
+  * method is called to remove the ServiceListener. The logger object must be destroyed before the 
+  * listeners member variable is destroyed because when the listeners member variable is destroyed 
+  * it leaves the ServiceListeners data structures in an unusable state. If the logger object 
+  * destructor runs after the listeners object destructor, it results in an access violation. 
+  */
+  /**
    * Framework id.
    */
   int id;
--- a/framework/src/util/CFRLogger.cpp
+++ b/framework/src/util/CFRLogger.cpp
@@ -29,7 +29,6 @@ namespace cfrimpl {
 CFRLogger::CFRLogger()
   : serviceTracker()
   , logService(nullptr)
-  , bOpen(false)
 {
 }
 
@@ -125,7 +124,6 @@ void CFRLogger::Open()
   }
   serviceTracker = std::make_unique<cppmicroservices::ServiceTracker<cppmicroservices::logservice::LogService>> (cfrContext, this);
   serviceTracker->Open();
-  bOpen = true;
 }
 
 void CFRLogger::Close() 
@@ -136,7 +134,6 @@ void CFRLogger::Close()
     serviceTracker->Close();
     serviceTracker.reset();
   }
-  bOpen = false;
 }
 
 } // cfrimpl
--- a/framework/src/util/CFRLogger.cpp
+++ b/framework/src/util/CFRLogger.cpp
@@ -48,7 +48,7 @@ CFRLogger::AddingService(
   auto currLogger = std::atomic_load(&logService);
   std::shared_ptr<cppmicroservices::logservice::LogService> logger;
   if (!currLogger && reference) {
-    logger = GetBundleContext().GetService<cppmicroservices::logservice::LogService>(
+    logger = cfrContext.GetService<cppmicroservices::logservice::LogService>(
       reference);
     std::atomic_store(&logService, logger);
   }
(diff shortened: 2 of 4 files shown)


achristoforides:
> The fix for the access violation on shutdown was a MathWorks specific fix in ServiceTracker.tpp which is not included in this pull request. While fixing the access violation a couple of small bugs were detected in CFRLogger and some inconsistencies with the way CFRLogger was being initialized in CoreBundleContext.cpp, as compared to other Service Trackers (like ServiceHook) so this PR is addressing those issues.

Hi Patty. The CFRLogger took its structure from the SCRLogger and CMLogger classes. The code as currently presented is more inconsistent with the existing patterns now than it was originally. I'm not suggesting any change with this comment but just wanted to get some clarification as to what inconsistencies you saw given that it was meant to mimic SCR/CMLogger rather than ServiceHook.

framework/src/util/CFRLogger.cpp
```diff
@@ -112,10 +115,38 @@ void CFRLogger::Log(const cppmicroservices::ServiceReferenceBase& sr,
...
+    auto currLogger = std::atomic_load(&logService);
+    if (currLogger) {
+      currLogger->Log(sr, level, message, ex);
+    }
   }
 }
+
+void CFRLogger::Open()
+{
+  auto l = this->Lock();
```
achristoforides, line 128: There is only ever 1 instance of CFRLogger in the process. It is opened in the constructor of `CoreBundleContext` and closed in the destructor. Additionally, the only class who uses the CFRLogger is `CoreBundleContext`. Code outside of that class should not be able to call open on it.

Why is a lock here necessary given that there is only 1 call to open and 1 call to close in the entire library/process?

framework/src/util/CFRLogger.cpp
```diff
@@ -112,10 +115,38 @@ void CFRLogger::Log(const cppmicroservices::ServiceReferenceBase& sr,
...
+  US_UNUSED(l);
+  serviceTracker.reset(
+    new cppmicroservices::ServiceTracker<cppmicroservices::logservice::LogService> (GetBundleContext(), this));
```
achristoforides, line 131: Why can't this be done in the constructor initializer lists as originally written?

achristoforides (changes requested):
Based on the description in the PR, it is unclear why a lot of these changes are being made in addition to what issues they are addressing. I see that from the CI results, they work, but the reason for these are lost on me without further clarification.

jeffdiclemente (changes requested):
I'm confused as to what problem this change is solving. A description of the root cause would help to provide context for the solution.

framework/src/util/CFRLogger.h
```diff
@@ -46,11 +46,12 @@ namespace cfrimpl {
  */
 class CFRLogger final
   : public cppmicroservices::logservice::LogService
+  , public detail::MultiThreaded<>
   , public cppmicroservices::ServiceTrackerCustomizer<
       cppmicroservices::logservice::LogService>
 {
 public:
-  explicit CFRLogger(cppmicroservices::BundleContext context);
+  CFRLogger(cppmicroservices::CoreBundleContext*);
```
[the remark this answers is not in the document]
pelliott-mathworks: If we construct the CFRLogger object as part of the CoreBundleContext construction, the BundleContext doesn't yet exist.

framework/src/util/CFRLogger.cpp
```diff
@@ -21,25 +21,22 @@
...
+#include "cppmicroservices/GetBundleContext.h"
 
 namespace cppmicroservices {
 namespace cfrimpl {
-CFRLogger::CFRLogger(cppmicroservices::BundleContext context)
-  : cfrContext(std::move(context))
-  , serviceTracker(
-      std::make_unique<cppmicroservices::ServiceTracker<
-        cppmicroservices::logservice::LogService>>(cfrContext, this))
+CFRLogger::CFRLogger(cppmicroservices::CoreBundleContext*)
```
jeffdiclemente, line 29: `cppmicroservices::CoreBundleContext*` is never used. this can be removed as a parameter.

framework/src/util/CFRLogger.cpp
```diff
@@ -117,5 +114,32 @@ void CFRLogger::Log(const cppmicroservices::ServiceReferenceBase& sr,
...
+    serviceTracker->Close();
+    serviceTracker.reset();
+  }
+  bOpen = false;
+}
+
+bool CFRLogger::IsOpen() const 
+{
+  return bOpen;
+}
```
jeffdiclemente, line 143: there are no call sites to `IsOpen`, I suggest removing this method.

framework/src/util/CFRLogger.cpp
```diff
@@ -117,5 +114,32 @@ void CFRLogger::Log(const cppmicroservices::ServiceReferenceBase& sr,
     currLogger->Log(sr, level, message, ex);
   }
 }
+
+void CFRLogger::Open()
+{
+  auto l = this->Lock();
+  US_UNUSED(l);
+  cfrContext = GetBundleContext();
```
jeffdiclemente, line 122: however unlikely, `GetBundleContext()` can return an "invalid" bundle context which, if used, will cause failures. This failure mode needs to be handled.

jeffdiclemente (changes requested):

framework/src/util/CFRLogger.cpp
```diff
@@ -117,5 +114,32 @@ void CFRLogger::Log(const cppmicroservices::ServiceReferenceBase& sr,
...
+  serviceTracker.reset(
+    new cppmicroservices::ServiceTracker<cppmicroservices::logservice::LogService> (cfrContext, this));
```
jeffdiclemente, line 124: prefer using `serviceTracker = std::make_shared<cppmicroservices::cfrimpl::CFRLogger>(...)` instead of `reset()` to avoid calling `new` and leverage better memory management by the c++ runtime.

jeffdiclemente (changes requested):

framework/src/util/CFRLogger.h
```diff
@@ -83,13 +84,18 @@ class CFRLogger final
...
+  void Open();
+  void Close();
 
 private:
   cppmicroservices::BundleContext cfrContext;
   std::unique_ptr<
     cppmicroservices::ServiceTracker<cppmicroservices::logservice::LogService>>
     serviceTracker;
   std::shared_ptr<cppmicroservices::logservice::LogService> logService;
+  std::atomic<bool> bOpen;
```
jeffdiclemente, line 98: `bOpen` can be removed completely, I don't see it used anywhere.

framework/src/bundle/CoreBundleContext.h
```diff
@@ -125,6 +118,13 @@ class CoreBundleContext
    */
   ServiceRegistry services;
 
+  /**
+   * A LogService for logging framework messages via
+   * a default or user-provided LogService that are intended to be
+   * visible outside of the framework.
+   */
+  std::shared_ptr<cppmicroservices::cfrimpl::CFRLogger> logger;
```
jeffdiclemente, line 126: it's worthwhile, until we fix the root cause, to leave a comment warning others that the order of member declaration generally matters here.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

achristoforides (approved):

PR #709, opened 2022-08-01
Bundle tracker

shane-riley:


framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -31,23 +31,25 @@ namespace detail {
 
 template<class S, class TTT, class R>
 BundleAbstractTracked<S,TTT,R>::BundleAbstractTracked(BundleContext bc)
-  : closed(false), trackingCount(0), bc(bc)
+  : closed(false)
+  , trackingCount(0)
+  , bc(bc)
```
achristoforides, line 36: This works but I'd recommend changing the parameter name from `bc` to something else as the private member is also called `bc`. This should help with code readability.

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -31,23 +31,25 @@ namespace detail {
...
-= default;
+BundleAbstractTracked<S,TTT,R>::~BundleAbstractTracked() = default;
 
 template<class S, class TTT, class R>
 void BundleAbstractTracked<S,TTT,R>::SetInitial(const std::vector<S>& initiallist)
 {
   std::copy(initiallist.begin(), initiallist.end(), std::back_inserter(initial));
 
-  if (bc.GetLogSink()->Enabled())
+  if (bc.GetLogSink()->Enabled())
```
achristoforides, line 48: Remove extra whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -57,12 +59,12 @@ void BundleAbstractTracked<S,TTT,R>::SetInitial(const std::vector<S>& initiallis
 template<class S, class TTT, class R>
 void BundleAbstractTracked<S,TTT,R>::TrackInitial()
 {
-  while (true)
+  while (true)
```
achristoforides, line 62: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -57,12 +59,12 @@ void BundleAbstractTracked<S,TTT,R>::SetInitial(const std::vector<S>& initiallis
...
 void BundleAbstractTracked<S,TTT,R>::TrackInitial()
 {
-  while (true)
+  while (true) 
   {
     S item;
     {
       auto l = this->Lock(); US_UNUSED(l);
-      if (closed || (initial.size() == 0))
+      if (closed || (initial.size() == 0))
```
achristoforides, line 67: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -75,13 +77,13 @@ void BundleAbstractTracked<S,TTT,R>::TrackInitial()
        */
       item = initial.front();
       initial.pop_front();
-      if (tracked.end() != tracked.find(item))
+      if (tracked.end() != tracked.find(item))
```
achristoforides, line 80: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -75,13 +77,13 @@ void BundleAbstractTracked<S,TTT,R>::TrackInitial()
...
       initial.pop_front();
-      if (tracked.end() != tracked.find(item))
+      if (tracked.end() != tracked.find(item)) 
       {
         /* if we are already tracking this item */
         DIAG_LOG(*bc.GetLogSink()) << "BundleAbstractTracked::trackInitial[already tracked]: " << item;
         continue; /* skip this item */
       }
-      if (std::find(adding.begin(), adding.end(), item) != adding.end())
+      if (std::find(adding.begin(), adding.end(), item) != adding.end())
```
achristoforides, line 86: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -110,18 +112,21 @@ void BundleAbstractTracked<S,TTT,R>::Close()
 template<class S, class TTT, class R>
 void BundleAbstractTracked<S,TTT,R>::Track(S item, R related)
 {
-  std::shared_ptr<TrackedParamType> object;
+  bool isInMap = false;
+  TrackedParamType object;
   {
     auto l = this->Lock(); US_UNUSED(l);
-    if (closed)
+    if (closed)
```
achristoforides, line 119: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -110,18 +112,21 @@ void BundleAbstractTracked<S,TTT,R>::Close()
...
     }
     auto trackedItemIter = tracked.find(item);
-    if (trackedItemIter != tracked.end()) {
+    if (trackedItemIter != tracked.end()) 
+    {
       object = trackedItemIter->second;
+      isInMap = true;
     }
-    if (!object)
+    if (!isInMap)
```
achristoforides, line 129: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -130,19 +135,19 @@ void BundleAbstractTracked<S,TTT,R>::Track(S item, R related)
         return;
       }
       adding.push_back(item); /* mark this item is being added */
-    }
-    else
+    } 
+    else
```
achristoforides, line 139: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -130,19 +135,19 @@ void BundleAbstractTracked<S,TTT,R>::Track(S item, R related)
...
   }
 
-  if (!object)
+  if (!isInMap) 
   { /* we are not tracking the item */
     TrackAdding(item, related);
-  }
-  else
+  } 
+  else
```
achristoforides, line 150: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -156,12 +161,12 @@ void BundleAbstractTracked<S,TTT,R>::Track(S item, R related)
...
 void BundleAbstractTracked<S,TTT,R>::Untrack(S item, R related)
 {
-  std::shared_ptr<TrackedParamType> object;
+  TrackedParamType object;
   {
     auto l = this->Lock(); US_UNUSED(l);
     std::size_t initialSize = initial.size();
     initial.remove(item);
-    if (initialSize != initial.size())
+    if (initialSize != initial.size())
```
achristoforides, line 169: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -173,7 +178,7 @@ void BundleAbstractTracked<S,TTT,R>::Untrack(S item, R related)
 
     std::size_t addingSize = adding.size();
     adding.remove(item);
-    if (addingSize != adding.size())
+    if (addingSize != adding.size())
```
achristoforides, line 181: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -256,26 +265,26 @@ void BundleAbstractTracked<S,TTT,R>::CopyEntries_unlocked(TrackingMap& map) cons
...
   adding.remove(item);
-  if (addingSize != adding.size() && !closed)
+  if (addingSize != adding.size() && !closed) 
   {
     /*
      * if the item was not untracked during the customizer
      * callback
      */
-    if (custom)
+    if (custom)
```
achristoforides, line 279: Remove whitespace

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -285,15 +294,15 @@ template<class S, class TTT, class R>
...
   /* Call customizer outside of synchronized region */
-  try
+  try 
   {
     object = CustomizerAdding(item, related);
     becameUntracked = this->CustomizerAddingFinal(item, object);
-  }
-  catch (...)
+  } 
+  catch (...)
```
achristoforides, line 305: Remove whitespace

framework/include/cppmicroservices/detail/BundleTracker.tpp
```diff
@@ -0,0 +1,273 @@
...
+template<class T>
+template <typename S0, typename... S>
+constexpr typename BundleTracker<T>::BundleState BundleTracker<T>::CreateStateMask(S0 const &s0, S const &...s) {
+  static_assert((std::is_enum_v<S0> && ... && std::is_enum_v<S>),
+                "The function requires enumerations.");
+  static_assert((std::is_same_v<Bundle::State, std::decay_t<S0>> && ... &&
+                 std::is_same_v<Bundle::State, std::decay_t<S>>),
+                "All values must be Bundle States.");
+  return (s0 | ... | s);
+}
```
achristoforides, line 44: Mentioned earlier. Is it possible to remove the template type `T` dependency here as it is not needed?

framework/include/cppmicroservices/detail/BundleTracker.tpp
```diff
@@ -0,0 +1,273 @@
...
+    d->GetBundles_unlocked(bundles, t.get());
+  }
+  return bundles;
+}
+
+template<class T>
+std::optional<typename BundleTracker<T>::TrackedParamType> BundleTracker<T>::GetObject(
+  const Bundle& bundle)
+{
+  std::optional<typename BundleTracker<T>::TrackedParamType> optional;
```
achristoforides, line 164: Consider renaming the variable to something more descriptive.

framework/include/cppmicroservices/detail/BundleTracker.tpp
```diff
@@ -0,0 +1,273 @@
...
+    US_UNUSED(l);
+    optional = t->GetCustomizedObject_unlocked(bundle);
+  }
+  return optional;
+}
+
+template<class T>
+typename BundleTracker<T>::TrackingMap BundleTracker<T>::GetTracked()
+{
+  BundleTracker<T>::TrackingMap map;
```
achristoforides, line 180: Consider renaming the variable to something more descriptive.

framework/include/cppmicroservices/detail/BundleTrackerPrivate.h
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  inline const BundleTracker<T>* q_func() const
+  {
+    return static_cast<BundleTracker<T>*>(q_ptr);
+  }
+
+  friend class BundleTracker<T>;
+
+  BundleTracker<T>* const q_ptr;
```
achristoforides, line 127: `q_func()` and `q_ptr` should be renamed to be more descriptive. I can understand what is going on here but that is not immediately obvious as the name of the functions here are `q_func()` which in no way say that it is casting the back-pointer `q_ptr`.

framework/include/cppmicroservices/detail/BundleTrackerPrivate.tpp
```diff
@@ -0,0 +1,95 @@
...
+
+namespace detail {
+
+template<class TTT>
+BundleTrackerPrivate<TTT>::BundleTrackerPrivate(
+  BundleTracker<T>* bt,
+  const BundleContext& context,
+  StateType stateMask,
+  std::shared_ptr<BundleTrackerCustomizer<T>> customizer)
+  : context(std::move(context))
```
achristoforides, line 41: Why is the BundleContext being moved here?

framework/include/cppmicroservices/detail/BundleTrackerPrivate.tpp
```diff
@@ -0,0 +1,95 @@
...
+  , stateMask(stateMask)
+  , customizer(customizer)
```
achristoforides, line 43: Change parameter names such that the member and the parameter name do not clash (even though this works fine)

framework/include/cppmicroservices/detail/BundleTrackerPrivate.tpp
```diff
@@ -0,0 +1,95 @@
...
+  , listenerToken()
+  , trackedBundle()
+  , q_ptr(bt)
+{
+  this->customizer = customizer; // customizer ? customizer : q_func();
```
achristoforides, line 48: I'm not sure what this comment is trying to explain. Is this old code that was just not removed?

framework/include/cppmicroservices/detail/BundleTrackerPrivate.tpp
```diff
@@ -0,0 +1,95 @@
...
+{
+  return trackedBundle.Load();
+}
+
+template<class TTT>
+std::vector<Bundle> BundleTrackerPrivate<TTT>::GetInitialBundles(
+  typename BundleTrackerPrivate<TTT>::StateType stateMask)
+{
+  std::vector<Bundle> result;
+  std::vector<Bundle> contextBundles = context.GetBundles();
```
achristoforides, line 66: switch to use `auto`

framework/include/cppmicroservices/detail/BundleTrackerPrivate.tpp
```diff
@@ -0,0 +1,95 @@
...
+  for (Bundle bundle : contextBundles) {
```
achristoforides, line 67: switch to use `auto`

framework/include/cppmicroservices/detail/ServiceTracker.tpp
```diff
@@ -354,7 +355,12 @@ ServiceTracker<S,T>::GetService(const ServiceReference<S>& reference) const
   { /* if ServiceTracker is not open */
     return std::shared_ptr<TrackedParamType>();
   }
-  return (t->Lock(), t->GetCustomizedObject_unlocked(reference));
+  {
+    auto l = t->Lock();
+    US_UNUSED(l);
+    auto optional = t->GetCustomizedObject_unlocked(reference);
```
achristoforides, line 361: Rename variable to be more descriptive.

framework/include/cppmicroservices/detail/TrackedBundle.h
```diff
@@ -0,0 +1,125 @@
...
+   * called while synchronized on this object.
+   *
+   * @param bundle Tracked bundle.
+   * @param related Action related object.
+   * @param object Customized object for the tracked bundle.
+   */
+  void CustomizerRemoved(Bundle bundle,
+                         const BundleEvent& related,
+                         const TrackedParamType& object) override;
+};
```
achristoforides, line 117: Comments here should be explicit about when these callbacks are called (i.e., what it means for a Bundle to be Added, Modified, or Removed).

framework/include/cppmicroservices/detail/TrackedBundle.tpp
```diff
@@ -0,0 +1,136 @@
...
+
+namespace cppmicroservices {
+
+namespace detail {
+
+template<class TTT>
+TrackedBundle<TTT>::TrackedBundle(BundleTracker<T>* bundleTracker,
+                                  BundleTrackerCustomizer<T>* customizer)
+  : Superclass(bundleTracker->d->context)
+  , bundleTracker(bundleTracker)
```
achristoforides, line 31: rename parameter

framework/include/cppmicroservices/detail/TrackedBundle.tpp
```diff
@@ -0,0 +1,136 @@
...
+  , customizer(customizer)
```
achristoforides, line 32: rename parameter

framework/include/cppmicroservices/detail/TrackedBundle.tpp
```diff
@@ -0,0 +1,136 @@
...
+      << "TrackedService::BundleChanged[" << state << "]: " << bundle;
+  }
+
+  // Track iff state in mask
+  if (state & bundleTracker->d->stateMask) {
+    this->Track(bundle, event);
+    /*
+     * If the customizer throws an unchecked exception, it is
+     * safe to let it propagate
+     */
```
achristoforides, line 94: Move comment above the line of code for readability

framework/include/cppmicroservices/detail/TrackedBundle.tpp
```diff
@@ -0,0 +1,136 @@
...
+  } else {
+    this->Untrack(bundle, event);
+    /*
+     * If the customizer throws an unchecked exception, it is
+     * safe to let it propagate
+     */
```
achristoforides, line 100: Move comment above the line of code for readability

framework/include/cppmicroservices/detail/TrackedService.tpp
```diff
@@ -25,23 +25,24 @@ namespace cppmicroservices {
 namespace detail {
 
 template<class S, class TTT>
-TrackedService<S,TTT>::TrackedService(ServiceTracker<S,T>* serviceTracker,
-                  ServiceTrackerCustomizer<S,T>* customizer)
+TrackedService<S, TTT>::TrackedService(
+  ServiceTracker<S, T>* serviceTracker,
+  ServiceTrackerCustomizer<S, T>* customizer)
   : Superclass(serviceTracker->d->context)
   , serviceTracker(serviceTracker)
```
achristoforides, line 32: Rename parameter

framework/include/cppmicroservices/detail/TrackedService.tpp
```diff
@@ -25,23 +25,24 @@ namespace cppmicroservices {
...
   , customizer(customizer)
```
achristoforides, line 33: Rename parameter

framework/include/cppmicroservices/detail/TrackedService.tpp
```diff
@@ -109,44 +103,51 @@ void TrackedService<S,TTT>::ServiceChanged(const ServiceEvent& event)
...
-std::shared_ptr<typename TrackedService<S,TTT>::TrackedParamType>
-TrackedService<S,TTT>::CustomizerAdding(ServiceReference<S> item,
-                                        const ServiceEvent& /*related*/)
+std::optional<
+  std::shared_ptr<typename TrackedService<S, TTT>::TrackedParamType>>
+TrackedService<S, TTT>::CustomizerAdding(ServiceReference<S> item,
+                                         const ServiceEvent& /*related*/)
 {
-  return customizer->AddingService(item);
+  auto value = customizer->AddingService(item);
```
achristoforides, line 130: Rename variable to be more descriptive

framework/test/gtest/BundleTrackerCustomCallbackTest.cpp
```diff
@@ -0,0 +1,661 @@
...
+    .Start();
+
+  // Then the framework receives it
+  EXPECT_TRUE(receivedErrorEvent)
+    << "Framework did not receive error after RemovedBundle throw";
+
+  EXPECT_CALL(*customizer, RemovedBundle).Times(::testing::AnyNumber());
+  bundleTracker.Close();
+  framework.GetBundleContext().RemoveListener(std::move(token));
+}
```
achristoforides, line 661: Add newline at the end of the file

achristoforides (changes requested):
Great job!

Please run `git-clang-format` in your local CppMicroServices directory and push those changes. There are many formatting inconsistencies with the style that we use.

framework/test/gtest/BundleTrackerConcurrencyTest.cpp
```diff
@@ -0,0 +1,232 @@
...
+#include "gtest/gtest.h"
+
+using namespace cppmicroservices;
+
+class BundleTrackerConcurrencyTest : public ::testing::Test
+{
+protected:
+  Framework framework;
+  BundleContext context;
+  BundleTracker<>::BundleState all_states =
```
jeffdiclemente, line 41: make `constexpr`

framework/test/gtest/BundleTrackerCustomCallbackTest.cpp
```diff
@@ -0,0 +1,661 @@
...
+#endif
+
+using namespace cppmicroservices;
+
+class BundleTrackerCustomCallbackTest : public ::testing::Test
+{
+protected:
+  Framework framework;
+  BundleContext context;
+  BundleTracker<>::BundleState all_states =
```
jeffdiclemente, line 43: make `constexpr`

framework/test/gtest/BundleTrackerCustomCallbackTest.cpp
```diff
@@ -0,0 +1,661 @@
...
+  Bundle bundle = cppmicroservices::testing::InstallLib(
+    framework.GetBundleContext(), "TestBundleA");
+  bundle.Start();
+
+  // Then the object 5 should be the tracked
+  EXPECT_EQ(5, bundleTracker.GetObject(bundle))
+    << "The object returned by a customizer's AddingBundle should be tracked";
+
+  // Whatever happens after the bundle is added is out of scope for this test
+  EXPECT_CALL(*customizer, RemovedBundle).Times(::testing::AnyNumber());
```
jeffdiclemente, line 259: I see a number of these `EXPECT_CALL`s throughout these tests that follow the pattern of:
1. have a comment of the form: "Whatever happens after <XYZ> is out of scope for this test"
2. have `EXPECT_CALL(*customizer, RemovedBundle).Times(::testing::AnyNumber());`

If the behavior is out of scope, an EXPECT_CALL isn't necessary. What am I missing?

jeffdiclemente (changes requested):

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
pelliott-mathworks, line 36: can we use std::move here? The parameter bc is a copy of the BundleContext. Can we move that copy to the member variable?

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -110,18 +112,21 @@ void BundleAbstractTracked<S,TTT,R>::Close()
 template<class S, class TTT, class R>
 void BundleAbstractTracked<S,TTT,R>::Track(S item, R related)
 {
-  std::shared_ptr<TrackedParamType> object;
+  bool isInMap = false;
```
pelliott-mathworks, line 115: Just wondering if this is threadsafe. It's done outside the lock for the object.

framework/include/cppmicroservices/detail/BundleTracker.tpp
```diff
@@ -0,0 +1,273 @@
...
+  } catch (...) {
+  }
+}
+
+template<class T>
+BundleTracker<T>::BundleTracker(
+  const BundleContext& context,
+  BundleState stateMask,
+  std::shared_ptr<_BundleTrackerCustomizer> customizer)
+  : d(new _BundleTrackerPrivate(this, context, stateMask, customizer))
```
pelliott-mathworks, line 61: I think d is a std::unique_ptr. Can we use make_uniqueptr instead? new should be avoided at all costs since it requires a delete to recover the memory.

framework/include/cppmicroservices/BundleTracker.h
```diff
@@ -0,0 +1,266 @@
...
+   * 
+   * @param context The <code>BundleTrackerContext</code> from which tracking occurs.
+   * @param stateMask The bit mask which defines the bundle states to be tracked.
+   * @param customizer The customizer to call when bundles are added, modified, or removed.
+   *                   If the customizer is nullptr, then the callbacks in this <code>BundleTracker</code> will 
+   *                   be used instead (default or can be overridden).
+   */
+  BundleTracker(
+    const BundleContext& context,
+    BundleState stateMask,
```
pelliott-mathworks, line 114: Wondering if the stateMask can be a const. Does it get changed after it's initialized?

framework/include/cppmicroservices/detail/TrackedBundle.h
```diff
@@ -0,0 +1,125 @@
...
+ * the CppMicroServices bundle system.
+ */
+template<class TTT>
+class TrackedBundle
+  : public TrackedBundleListener
+  , public BundleAbstractTracked<Bundle, TTT, BundleEvent>
+{
+
+public:
+  using T = typename TTT::TrackedType;
```
pelliott-mathworks, line 48: Last time mentioning the using statements but please reconsider all of these.

framework/include/cppmicroservices/detail/TrackedBundle.tpp
```diff
@@ -0,0 +1,136 @@
...
+
+template<class TTT>
+void TrackedBundle<TTT>::BundleChanged(const BundleEvent& event)
+{
+  // Call track or untrack based on state mask
+
+  // Ignore events that do not correspond with
+  // Bundle state changes
+
+  (void)latch.CountUp();
```
pelliott-mathworks, line 51: I don't think this is the correct use of the latch. See Declarative Services CCActiveState.cpp for the correct usage. 
 Example code:
 * function1()
 * {
 *    if(latch.CountUp())
 *    {
 *       // do something critical here
 *       latch.CountDown();
 *    }
 * }
 * function2()
 * {
 *    latch.Wait(); // this blocks the thread until all threads that have exited function1
 *    // do something here.
 * }
 
CountUp is protecting some code. It can only continue if the return value is true.

pelliott-mathworks, line 51: This isn't the correct usage of the latch function. See DeclarativeServices CCActiveState.cpp for example of correct usage. 

 Example code:
 * function1()
 * {
 *    if(latch.CountUp())
 *    {
 *       // do something critical here
 *       latch.CountDown();
 *    }
 * }
 * function2()
 * {
 *    latch.Wait(); // this blocks the thread until all threads that have exited function1
 *    // do something here.
 * }
The CountUp function protects some code. That code can only be executed if CountUp returns true.

pelliott-mathworks (changes requested):

framework/include/cppmicroservices/Bundle.h
```diff
@@ -866,4 +868,9 @@ US_Framework_EXPORT std::ostream& operator<<(std::ostream& os,
                                              Bundle::State state);
 }
 
+// Hashing behavior for Bundles
+US_HASH_FUNCTION_BEGIN(cppmicroservices::Bundle)
```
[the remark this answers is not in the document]
shane-riley: The constref is included in the hashing macro. See GlobalConfig.h.in:142

framework/include/cppmicroservices/BundleTracker.h
```diff
@@ -0,0 +1,266 @@
...
+   * Create a <code>StateType</code> stateMask for a BundleTracker
+   * 
+   * @param s0 the <code>Bundle::State</code>.
+   * @param s the subsequent <code>Bundle::States5</code>.
+   * 
+   * @return The state mask.
+   * 
+   */
+  template <typename S0, typename... S>
+  static constexpr typename BundleTracker<T>::BundleState CreateStateMask(S0 const &s0, S const &...s);
```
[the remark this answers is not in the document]
shane-riley: It is true that `T` does not change the behavior of `BundleTracker<T>::CreateStateMask(...)`. However, it is not immediately obvious to me how to create a class method `BundleTracker::CreateStateMask(...)`, since class templates cannot share a name with a non-template classes.

framework/include/cppmicroservices/detail/TrackedService.tpp
shane-riley: Changed `value` to `serviceObjectPointer`, and added a comment to better explain what the return statement is doing.

framework/test/gtest/BundleTrackerConcurrencyTest.cpp
shane-riley: Changed to `static constexpr`

framework/test/gtest/BundleTrackerCustomCallbackTest.cpp
shane-riley: Changed to `static constexpr`

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
shane-riley: Changed to `context`

framework/include/cppmicroservices/detail/BundleTracker.tpp
shane-riley: Changed to `optionalTrackedObject`

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
shane-riley: The flag `isInMap` is declared outside of the synchronized block since it is required later to determine which customizer to call. Since the flag is local to the function call and the actual map-checking is synchronized, it should be okay.

framework/include/cppmicroservices/BundleTracker.h
```diff
@@ -0,0 +1,266 @@
...
+  bool IsEmpty();
+
+  /**
+   * Called when a <code>Bundle</code> is modified that is being tracked by this <code>BundleTracker</code>
+   * and the <code>BundleTrackerCustomizer</code> constructor argument was nullptr.
+   *
+   * When a tracked bundle changes states, this method is called.
+   *
+   * @param bundle The tracked <code>Bundle</code> whose state has changed.
+   * @param event The BundleEvent which was caught by the <code>BundleTracker</code>. Can be null.
```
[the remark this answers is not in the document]
shane-riley: That clause comes from the OSGi spec. I know that `RemovedBundle` could be called with an invalid `BundleEvent` if the bundle is removed from the map via `BundleTracker::Remove`. I can't think of a reason for `ModifiedBundle()` to have an invalid `BundleEvent`, so I can drop that clause

framework/include/cppmicroservices/BundleTracker.h
```diff
@@ -0,0 +1,266 @@
...
+   *
+   * Bundles that match the state mask will be tracked by this <code>BundleTracker</code>.
+   *
+   * @throws std::logic_error If the <code>BundleTrackerContext</code> used in the creation of this
+   * <code>BundleTracker</code> is no longer valid.
+   */
+  void Open();
+
+  /**
+   * Remove a bundle from this <code>BundleTracker</code>.
```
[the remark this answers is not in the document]
shane-riley: Yes and yes. Modified accordingly

framework/include/cppmicroservices/detail/BundleTracker.tpp
shane-riley: Changed to `trackingMap`

framework/include/cppmicroservices/detail/BundleTrackerPrivate.tpp
shane-riley: contextBundles is changed to using `auto`

framework/include/cppmicroservices/detail/ServiceTracker.tpp
shane-riley: Changed to `customObject`

framework/include/cppmicroservices/BundleTracker.h
```diff
@@ -0,0 +1,266 @@
...
+   * @param bundle the <code>Bundle</code> to be removed
+   */
+  void Remove(const Bundle&);
+
+  /**
+   * Called when a <code>Bundle</code> is being removed from this <code>BundleTracker</code>
+   * and the <code>BundleTrackerCustomizer</code> constructor argument was nullptr.
+   *
+   * @param bundle The tracked <code>Bundle</code> that is being removed.
+   * @param event The BundleEvent which was caught by the <code>BundleTracker</code>. Can be null.
```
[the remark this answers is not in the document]
shane-riley: Rewriting the doxygen comments for the customizer methods generally to improve clarity.

In this case, the `BundleEvent` can be null-constructed if the `RemovedBundle()` is called by a manual call to `BundleTracker::Remove()`. This language also comes from the OSGi spec.

framework/include/cppmicroservices/detail/BundleAbstractTracked.h
```diff
@@ -200,10 +201,9 @@ class BundleAbstractTracked
    * @param related Action related object.
    * @param object Customized object for the tracked item.
    */
-  virtual void CustomizerModified(
-    S item,
-    const R& related,
-    const std::shared_ptr<TrackedParamType>& object) = 0;
+  virtual void CustomizerModified(S item,
+                                  const R& related,
+                                  const TrackedParamType& object) = 0;
```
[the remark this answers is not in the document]
shane-riley: Actually no. If `std::nullopt` is returned by `AddingBundle()`, then the `Bundle` is not tracked, meaning therefore that a `CustomizerModified` will never be called on it.

As a result, if `CustomizerModified` is called on a `Bundle` we can guarantee that "something" (i.e. a `Bundle` or a custom object) is stored as the `object` and provided to the customizer.

framework/include/cppmicroservices/detail/BundleAbstractTracked.h
```diff
@@ -213,10 +213,9 @@ class BundleAbstractTracked
    * @param related Action related object.
    * @param object Customized object for the tracked item.
    */
-  virtual void CustomizerRemoved(
-    S item,
-    const R& related,
-    const std::shared_ptr<TrackedParamType>& object) = 0;
+  virtual void CustomizerRemoved(S item,
+                                 const R& related,
+                                 const TrackedParamType& object) = 0;
```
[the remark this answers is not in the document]
shane-riley: (Repeated from note on `CustomizerModified`)

Actually no. If `std::nullopt` is returned by `AddingBundle()`, then the Bundle is not tracked, meaning therefore that a `CustomizerRemoved` will never be called on it.

As a result, if `CustomizerRemoved`is called on a Bundle we can guarantee that "something" (i.e. a `Bundle` or a custom object) is stored as the object and provided to the customizer.

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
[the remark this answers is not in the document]
shane-riley: The tracking map should never store nullopts, since an `AddingBundle` call that returns a `std::nullopt` necessitates that the `Bundle` is not tracked. That is why `std::optional` is used to determine whether to track, but the value type in the map is just `TrackedParamType`.

`isInMap` is solving a slightly different problem here. Previously, the tracking map was Bundles/ServiceReferences -> shared pointers. `BundleAbstractTracked` was relying on the fact that a null-constructed shared pointer is falsy as a means to check whether the key was found in the map. When we change to storing value types as the value, we do not want to force the user to define `bool()` for their tracked objects. As a result, `isInMap` is used to flag whether the key was found in the map. It is unrelated to `std::nullopt` here.

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
```diff
@@ -218,12 +223,16 @@ bool BundleAbstractTracked<S,TTT,R>::IsEmpty_unlocked() const
...
 {
   typename TrackingMap::const_iterator i = tracked.find(item);
-  if (i != tracked.end()) return i->second;
-  return std::shared_ptr<TrackedParamType>();
+  if (i != tracked.end())
+  {
+    return i->second;
+  }
+
+  return std::nullopt;
```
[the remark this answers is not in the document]
shane-riley: As I understand it, the type implicitly converts to a `std::optional` here. To make it explicit however, the line is modified to construct a `std::optional`

framework/include/cppmicroservices/detail/BundleTracker.tpp
```diff
@@ -0,0 +1,273 @@
...
+  auto t = d->Tracked();
+  if (!t) { /* If BundleTracker is not open */
+    return true;
+  }
+  {
+    auto l = t->Lock();
+    US_UNUSED(l);
+    isEmpty = t->IsEmpty_unlocked();
+  }
+  return isEmpty;
```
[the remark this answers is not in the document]
shane-riley: Modified to `return (t->Lock(), t->IsEmpty_unlocked());`

framework/include/cppmicroservices/detail/BundleTracker.tpp
```diff
@@ -0,0 +1,273 @@
...
+  auto t = d->Tracked();
+  if (!t) { /* If BundleTracker is not open */
+    return 0;
+  }
+  {
+    auto l = t->Lock();
+    US_UNUSED(l);
+    size = t->Size_unlocked();
+  }
+  return size;
```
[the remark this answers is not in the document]
shane-riley: Modified to `return (t->Lock(), t->Size_unlocked());`

framework/include/cppmicroservices/detail/BundleTracker.tpp
```diff
@@ -0,0 +1,273 @@
...
+  auto t = d->Tracked();
+  if (!t) { /* If BundleTracker is not open */
+    return -1;
+  }
+  {
+    auto l = t->Lock();
+    US_UNUSED(l);
+    count = t->GetTrackingCount();
+  }
+  return count;
```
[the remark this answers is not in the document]
shane-riley: Modified to `return (t->Lock(), t->GetTrackingCount());`

framework/include/cppmicroservices/detail/BundleTracker.tpp
```diff
@@ -0,0 +1,273 @@
...
+  auto t = d->Tracked();
+  if (!t || !bundle) { /* If BundleTracker is not open or if bundle is invalid */
+    return std::nullopt;
+  }
+  {
+    auto l = t->Lock();
+    US_UNUSED(l);
+    optional = t->GetCustomizedObject_unlocked(bundle);
+  }
+  return optional;
```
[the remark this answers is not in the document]
shane-riley: Modified to 'return (t->Lock(), t->GetCustomizedObject_unlocked(bundle));'

framework/include/cppmicroservices/detail/BundleAbstractTracked.tpp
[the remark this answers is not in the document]
shane-riley: They are split since the first occurs in a locked region (manages map) and the second must not (calls customizers). Is there a better way to write this?

framework/include/cppmicroservices/detail/TrackedBundle.h
shane-riley: I've cleaned up the behavioral explanations in the BundleTracker and the BundleTrackerCustomizer for the callbacks. Since these methods just wrap the callbacks, I will add a @see tag to the doxygen linking to the API to describe the calling conditions

framework/include/cppmicroservices/detail/BundleTrackerPrivate.h
shane-riley: This pattern came from the ServiceTracker. Naming has been overhauled to better describe behavior:

`q_ptr` => `bundleTracker`
`q_func` => `getTrackerAsCustomizer`

Also the static_casts are made to target `BundleTrackerCustomizer` and not `BundleTracker`, since that is what they should have been in the first place.

framework/include/cppmicroservices/detail/TrackedBundle.tpp
shane-riley: My usage here is duplicated from a commit by Jeff to resolve a race in the `ServiceTracker` about 15 months ago where the tracker closes before the customizers are finished.

Right now, when `Close()` is called it calls `CounterLatch::Wait()`. As I understand it now, this should cause `CounterLatch::CountUp` to return false after that point, which we are not checking for currently and probably should be.

@jeffdiclemente any ideas on how to deal with this?

framework/include/cppmicroservices/BundleTracker.h
```diff
@@ -0,0 +1,266 @@
...
+  /**
+   * Return the number of bundles being tracked by this <code>BundleTracker</code>.
+   *
+   * @return The number of tracked bundles.
+   */
+  size_t Size();
+
+private:
+  using TypeTraits = typename BundleTrackerCustomizer<T>::TypeTraits;
+  using _BundleTracker = BundleTracker<T>;
```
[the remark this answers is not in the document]
shane-riley: Based on your comment and some discussions with @jeffdiclemente and @achristoforides, I think it's best that the TypeTraits struct gets dropped from the BundleTrackerCustomizer. The pattern still exists in the ServiceTracker, but could be removed as well. Now tracked objects in the `BundleTracker` are always referred to as `T`, not `TrackedType` or `TrackedParamType`. This allows us to greatly clean up the frivolous using statements, which I have done now.

framework/include/cppmicroservices/detail/BundleTracker.tpp
shane-riley: Moved to header

framework/include/cppmicroservices/detail/TrackedBundle.tpp
jeffdiclemente: in this instance `CountUp();` is used to ensure that all pending threads running `BundleChanged()` finish before the tracker cleans up and shuts down. There should be no need to check the return value of `CountUp()`. One could argue that a different synchronization mechanism should be used instead of the latch in this case.

framework/include/cppmicroservices/BundleTracker.h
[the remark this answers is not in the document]
shane-riley: An update: The current commit is able to deduce the type of the `BundleTracker` *sometimes*. When a `BundleTracker` is constructed without a customizer it can perform the type deduction so the `<>` default template instation notation can be dropped:
```
cppms::BundleTracker bundleTracker(context, stateMask); // builds
```

In the gtest's however, this breaks down when we try to use mock customizers. I'm not sure why, but it is struggling with converting the third argument:
```
class MockCustomizer : BundleTrackerCustomizer<>
{ /* ... */ }

MockCustomizer customizer;
cppms::BundleTracker bundleTracker(context, stateMask, customizer); // fails to build
```

Additionally, `CreateStateMask` fails to deduce the type of `BundleTracker`. I think this is because the parameters for that method are never the template type. Unfortunately, one can't simply add an additional declaration for a nontemplate `BundleTracker` (even though `CreateStateMask` really does not depend on the type). However, another option that could certainly work would be to simply change `CreateStateMask` from a static method to a first-class function in the `cppmicroservices` namespace, while keeping it in the `BundleTracker` header. This could set a bad precedent of namespace pollution, but would completely separate the state mask generation from the templated `BundleTracker`.

It is not obvious to me how to proceed in further enabling these type deductions, but it feels like there must be a way to fix the edge cases.

framework/include/cppmicroservices/detail/TrackedBundle.tpp
shane-riley: It is true that once the `Wait()` is called, we don't need to process any additional `BundleChanged()` events since the tracker is closing anyway. Definitely multiple ways to address this.

jeffdiclemente: No, once `Wait()` is called, it blocks the thread until the latch count reaches 0.
Closing the tracker unregisters the bundle listener, so no new bundle events can come into the tracker. the tracker then waits for all existing bundle listeners to complete.

framework/include/cppmicroservices/detail/BundleTrackerPrivate.tpp
```diff
@@ -0,0 +1,95 @@
+/*=============================================================================
```
jeffdiclemente, line 1: the .tpp files should be deleted and all content added to the .h file

framework/include/cppmicroservices/detail/TrackedBundle.tpp
```diff
@@ -0,0 +1,134 @@
+/*=============================================================================
```
jeffdiclemente, line 1: the .tpp files should be deleted and all content added to the .h file

jeffdiclemente (changes requested):

PR #713, opened 2022-08-04
Improve service dependency error logging

kkumar45:
This PR improves service dependency error logging by enabling DEBUG log to show the cause of error when a service fails to find it's dependency because of typo/misspelled in it's LDAP filter or the service properties.

Signed-off-by: The MathWorks, Inc. kunalk@mathworks.com

### DIFF
--- a/compendium/DeclarativeServices/test/CMakeLists.txt
+++ b/compendium/DeclarativeServices/test/CMakeLists.txt
@@ -78,6 +78,7 @@ set(_declarativeservices_tests
   TestRegistrationManager.cpp
   TestSCRBundleExtension.cpp
   TestServiceComponentRuntimeImpl.cpp
+  TestServiceDependencyErrorLogging.cpp
   TestServiceMetadataParserV1.cpp
   TestSetConfiguration.cpp
   TestSingletonComponentConfiguration.cpp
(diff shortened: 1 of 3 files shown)


jeffdiclemente (approved):

pelliott-mathworks (approved):

achristoforides (approved):

Burgch:
This change seems to be generating a LOT of noise in CppMicroServices logging.  As far as I can tell, the change to `ServiceListeners.cpp` results in the Framework logging every registered ServiceListener which has an `LDAPFilter` comprising more than just the `OBJECTCLASS` or `SERVICE_ID`, and which doesn't match the properties of the service being registered, _**and**_ it does this every time a service is registered. I believe this includes any `ServiceTracker` which is `Open`, as well as any DS services which have listeners for references with a `target`.

This is making it incredibly difficult to see anything useful in the logs, and when log rotation is in use, logs are overwritten incredibly quickly by this noise. Can we please remove it, or at least only log it with `DIAG_LOG`?

/cc @kkumar45 @jeffdiclemente @achristoforides @pelliott-mathworks

Burgch:
Great, thanks Jeff!

pelliott-mathworks (commented):
Approved

PR #714, opened 2022-08-11
702 ServiceTracker GetTrackingCount() when opened and then closed

shane-riley:
Ensure that the ServiceTracker::GetTrackingCount() method returns -1 if the tracker has been opened and then closed.


framework/include/cppmicroservices/detail/ServiceTracker.tpp
```diff
@@ -434,7 +434,7 @@ template<class S, class T>
 int ServiceTracker<S,T>::GetTrackingCount() const
 {
   auto t = d->Tracked();
-  if (!t)
+  if (!t || (t->Lock(), t->closed))
```
jeffdiclemente, line 437: I suggest placing `t->Lock()` in-between lines 436 and 437, it's easier to read.

jeffdiclemente (changes requested):

framework/include/cppmicroservices/detail/ServiceTracker.tpp
achristoforides: I'm not the biggest fan of using the comma operator here. I know we do it in other parts of the code but I'm of the opinion that it isn't good form.

Can this be changed to the following instead:
```cpp
auto t =d->Tracked()
auto lock(t->Lock());
US_UNUSED(lock);
if (!t || t->closed) {
  return -1;
}

return t->GetTrackingCount();
```

shane-riley: This snippet is adopted, with one important change: we have to check for `!t` before locking:
```
auto t =d->Tracked()
if (!t) {
  return -1;
}
auto lock(t->Lock());
US_UNUSED(lock);
if (t->closed) {
  return -1;
}

return t->GetTrackingCount();
```

shane-riley:
The BundleTracker PR has been updated to match the resolution to 702 in both implementation and in behavior.

PR #724, opened 2022-09-20
Update ComponentInstanceImpl.hpp

pelliott-mathworks:
Improve error message that is generated when an appropriate constructor isn't found for the Service Instance. Signed-off-by, The MathWorks, Inc. <pelliott@mathworks.com>

### DIFF
--- a/compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
+++ b/compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
@@ -447,7 +447,7 @@ public:
       "and/or the service implementation does not implement all of the "
       "service interface's methods. A constructor with service reference "
       "input parameters or a constructor with an AnyMap input parameter for "
-      "configuration properties plus service reference input parameters is "
+      "configuration properties and service reference input parameters is "
       "required when inject-references is true. ");
     return nullptr;
   }
--- a/compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
+++ b/compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
@@ -417,15 +417,11 @@ public:
   {
     static_assert(
       std::is_default_constructible<C>::value,
-      "An appropriate constructor was not found for the Service Instance. "
-      "The manifest.json inject-references property is false so the build "
-      "was not looking for a constructor with service reference input "
-      "parameters. It was looking for a default constructor or a constructor "
-      "with an AnyMap input parameter to receive the configuration properties. "
-      "Please check the constructor input parameters. "
-      "This failure can also be caused when all methods for the interface the "
-      "Service Instance is implementing are not present. Please check that all "
-      "interface methods are present.");
+      "An appropriate constructor was not found "
+      "and/or the service implementation does not implement all of the "
+      "service interface's methods. A default constructor or a constructor "
+      "with an AnyMap input parameter for the configuration properties is "
+      "required when inject-references is false. ");
     return nullptr;
   }
 
@@ -447,15 +443,12 @@ public:
     static_assert(
       std::is_constructible<C,
                             const std::shared_ptr<CtorInjectedRefs>&...>::value,
-      "An appropriate constructor was not found for the Service Instance. "
-      "The manifest.json inject-references property is true so the build "
-      "was looking for a constructor with input parameters to receive the "
-      "service references. The constructor may or may not have an AnyMap "
-      "input parameter to receive the configuration properties depending "
-      "on the configuration in the manifest.json file. Please check the "
-      "constructor input parameters. This failure can also be caused when "
-      "all methods for the interface the Service Instance is implementing "
-      "are not present. Please check that all interface methods are present.");
+      "An appropriate constructor was not found "
+      "and/or the service implementation does not implement all of the "
+      "service interface's methods. A constructor with service reference "
+      "input parameters or a constructor with an AnyMap input parameter for "
+      "configuration properties plus service reference input parameters is "
+      "required when inject-references is true. ");
     return nullptr;
   }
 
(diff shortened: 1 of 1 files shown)


compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
```diff
@@ -417,10 +417,15 @@ class ComponentInstanceImpl final
...
-      "appropriate constructor is not available or when the class being "
-      "validated is not a complete type.");
+      "An appropriate constructor was not found for the Service Instance. "
+      "The manifest.json inject-references property is false so the build "
+      "was looking for a default constructor or a constructor with an "
+      "AnyMap input parameter to receive the configuration properties. "
+      "All methods for the interfaces the Service Instance is implementing "
+      "must be present. Please see "
+      "https://confluence.mathworks.com/display/CppMicroServices/"
+      "Build+Failure+When+Constructing+Service+Instance "
```
jeffdiclemente, line 427: this is a MW internal website. We can't have this in the output.

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
```diff
@@ -442,9 +447,16 @@ class ComponentInstanceImpl final
...
-      "or when the class being validated is not a complete type.");
+      "An appropriate constructor was not found for the Service Instance. "
+      "The manifest.json inject-references property is true so the build "
+      "was looking for a constructor with input parameters to receive the "
+      "service parameters. The constructor may or may not have an AnyMap "
+      "input parameter to receive the configuration properties depending "
+      "on the configuration in the manifest.json file. All methods for the "
+      "interfaces the Service Instance is implementing must be present. Please see "
+      "https://confluence.mathworks.com/display/CppMicroServices/"
+      "Build+Failure+When+Constructing+Service+Instance for more "
```
jeffdiclemente, line 458: this is a MW internal website. We can't have this in the output.

jeffdiclemente (changes requested):

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
pelliott-mathworks: I think the content of the page is okay? It doesn't have anything MW specific. I just need to move it someplace else. Do you have a recommendation?

jeffdiclemente: Do we really need a URL that describes the problem in the error message? I suggest removing the URL from the error message.

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
```diff
@@ -419,13 +419,13 @@ class ComponentInstanceImpl final
...
-      "https://confluence.mathworks.com/display/CppMicroServices/"
-      "Build+Failure+When+Constructing+Service+Instance "
-      "for more information.");
+      "was not looking for a constructor with service reference input "
+      "parameters. It was looking for a default constructor or a constructor "
+      "with an AnyMap input parameter to receive the configuration properties. "
+      "Please check the constructor input parameters. "
+      "This failure can also be caused when all methods for the interface the "
+      "Service Instance is implementing are not present. Please check that all "
+      "interface methods are present.");
```
jeffdiclemente, line 428: I think the number of words can be cut down a bit. If the ways in which this error can occur is that either the service implementation doesn't implement all the service interface's methods OR the service implementation has no default constructor OR the service implementation has no constructor which takes an AnyMap, that should be simple to communicate. Something like:

```
An appropriate constructor was not found and/or the service implementation does not implement all of the service interface's methods. A default constructor or a constructor with an AnyMap input parameter is required when injection is false.
```

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
```diff
@@ -450,13 +450,12 @@ class ComponentInstanceImpl final
...
       "input parameter to receive the configuration properties depending "
-      "on the configuration in the manifest.json file. All methods for the "
-      "interfaces the Service Instance is implementing must be present. Please see "
-      "https://confluence.mathworks.com/display/CppMicroServices/"
-      "Build+Failure+When+Constructing+Service+Instance for more "
-      "information.");
+      "on the configuration in the manifest.json file. Please check the "
+      "constructor input parameters. This failure can also be caused when "
+      "all methods for the interface the Service Instance is implementing "
+      "are not present. Please check that all interface methods are present.");
```
jeffdiclemente, line 458: same general comment as above; work on the error message and making it concise and clear

jeffdiclemente (changes requested):

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
```diff
@@ -442,9 +443,12 @@ class ComponentInstanceImpl final
...
       std::is_constructible<C,
                             const std::shared_ptr<CtorInjectedRefs>&...>::value,
-      "Suitable constructor not found for constructor injection. "
-      "This error happens when an appropriate constructor is not available "
-      "or when the class being validated is not a complete type.");
+      "An appropriate constructor was not found "
+      "and/or the service implementation does not implement all of the "
+      "service interface's methods. A constructor with service reference "
+      "input parameters or a constructor with an AnyMap input parameter for "
+      "configuration properties plus service reference input parameters is "
```
jeffdiclemente, line 450: recommend changing `plus` to `and`

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #728, opened 2022-09-23
Performance improvement from `brian-performance` branch

achristoforides:


framework/src/util/Properties.cpp
```diff
@@ -158,6 +158,58 @@ std::pair<Any, bool> Properties::Value_unlocked(const std::string& key,
   }
 }
 
+const Any& Properties::Value(const std::string& key, bool matchCase) const
```
jeffdiclemente, line 161: This function should be named something more descriptive, e.g. `RefToValue`, `ValueByRef`, etc...

framework/include/cppmicroservices/detail/Threads.h
```diff
@@ -151,7 +151,7 @@ template<class MutexHost>
 class NoWaitCondition
 {};
 
-template<class LockingStrategy = MutexLockingStrategy<>,
+template<class LockingStrategy = MutexLockingStrategy<std::recursive_mutex>,
```
jeffdiclemente, line 154: this change requires more evaluation. I think this change is needed to support the changes in [framework/src/service/ServiceRegistrationBase.cpp](https://github.com/CppMicroServices/CppMicroServices/pull/728/files#diff-36be34fd05c509bdeed05610be9b89469c8d193ec6ce5431f47f6bfd468e4e72) and I'd like to understand why. Making the locking strategy use `recursive_mutex` is a significant change the locking strategy used within the core framework.

framework/src/service/ServiceReferenceBase.cpp
```diff
@@ -59,7 +59,12 @@ void ServiceReferenceBase::SetInterfaceId(const std::string& interfaceId)
 
 ServiceReferenceBase::operator bool() const
 {
-  return static_cast<bool>(GetBundle());
+  auto p = d.load();
+  if (p->registration == nullptr) {
+    return false;
+  }
+
+  return (!p->registration->bundle.expired());
```
jeffdiclemente, line 67: `std::weak_ptr::expired` is racy (https://en.cppreference.com/w/cpp/memory/weak_ptr/expired).  I'm not sure whether this matters or not for this usage. need to evaluate more thoroughly.

jeffdiclemente (changes requested):

framework/src/util/Properties.h
```diff
@@ -45,6 +45,8 @@ class Properties : public detail::MultiThreaded<>
   std::pair<Any, bool> Value_unlocked(const std::string& key,
                                       bool matchCase = false) const;
 
+  const Any& ValueByRef(const std::string& key, bool matchCase = false) const;
```
jeffdiclemente, line 48: This should be renamed to `ValueByRef_unlocked` to follow the same convention as `Value_unlocked` to indicate that this function doesn't hold a lock.

jeffdiclemente (approved):

carneyweb (approved):

jeffdiclemente:
It looks like the change in the scoping of the locks in `bool    ServiceReferenceBase::operator<(ServiceReferenceBase const& reference) const` from the PR causes a deadlock.

`d.load()->registration->properties.Lock();` is followed by a call to `reference.d.load()->registration->properties.Lock();`, which will deadlock if `d.load()->registration->properties` and `reference.d.load()->registration->properties` are the same object. This shouldn't never happen as there is a check to see if `d.load()->registration` == `reference.d.load()->registration` before entering the scoped block.

PR #731, opened 2022-10-03
Factory Configuration Bug Fix

pelliott-mathworks:
When factory configurations are created before the bundle containing the factory component is installed and started, DS should find those configurations in the Configuration Admin repository and register them as part of the factory component startup. Signed-off-by The MathWorks, Inc. <pelliott@mathworks.com>

### DIFF
--- a/compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
+++ b/compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
@@ -155,7 +155,7 @@ void ConfigurationNotifier::CreateFactoryComponent(
   // component except the factory component itself.
   newMetadata->configurationPids.clear();
   for (const auto& basePid : oldMetadata->configurationPids) {
-    if (basePid != factoryName) {
+    if (basePid != oldMetadata->configurationPids[0])  {
       newMetadata->configurationPids.emplace_back(basePid);
     }
   }
--- a/compendium/DeclarativeServices/test/TestFactoryPid.cpp
+++ b/compendium/DeclarativeServices/test/TestFactoryPid.cpp
@@ -137,6 +137,49 @@ TEST_F(tServiceComponent, testFactoryPidConstructionNameDifferentThanClass)
   auto instance = GetInstance<test::CAInterface>();
   ASSERT_TRUE(instance) << "GetService failed for CAInterface";
 
+
+}
+
+/* testFactoryConfigBeforeInstall
+   This test creates the configuration objects for the factory instances
+   before the factory component has been installed and started. 
+   Once the factory component is installed and started the instances 
+   associated with the configuration objects should be registered
+   automatically by DS.
+*/
+TEST_F(tServiceComponent, testFactoryConfigBeforeInstall)
+{
+  std::string configurationPid = "ServiceComponentPid";
+
+  // Get a service reference to ConfigAdmin to create the factory component instances.
+  auto configAdminService =
+      GetInstance<cppmicroservices::service::cm::ConfigurationAdmin>();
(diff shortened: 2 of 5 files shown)


compendium/DeclarativeServices/test/TestFixture.hpp
```diff
@@ -169,6 +169,21 @@ class tServiceComponent : public testing::Test
     return context.GetService<T>(instanceRef);
   }
 
+  template<class T>
+    std::vector<std::shared_ptr<T>> GetInstances()
+    {
+      std::vector<cppmicroservices::ServiceReference<T>> instanceRefs;
+      std::vector<std::shared_ptr<T>> instances;
+      instanceRefs = context.GetServiceReferences<T>();
+      if (instanceRefs.size() <= 0) {
```
jeffdiclemente, line 178: prefer using `empty()`

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -155,7 +155,7 @@ void ConfigurationNotifier::CreateFactoryComponent(
   // component except the factory component itself.
   newMetadata->configurationPids.clear();
   for (const auto& basePid : oldMetadata->configurationPids) {
-    if (basePid != factoryName) {
+    if (basePid != oldMetadata->configurationPids[0])  {
```
jeffdiclemente, line 158: not sure I understand this change. can you elaborate?

Also, `factoryName` is no longer used which will cause a build failure on linux/mac

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
pelliott-mathworks: The first pid in the configuration-pids for the factory component is the one used to create the factory instances. When creating the list of configuration-pids for the factory instance, it needs to be deleted from the list of configuration-pids and replaced with the specific configuration-pid that matches this factory instance. When I originally wrote this code I thought the pid used to create the factory instances was supposed to match the factory component name. That assumption was incorrect. Users have been using pids that do not match the factory component name to create the factory instances. So it's not correct to use the factoryName for the comparison here. I noticed this when testing the current change because I used a pid that was different than the factory component name to create the factory instances.

compendium/DeclarativeServices/test/TestFactoryPid.cpp
```diff
@@ -137,6 +137,49 @@ TEST_F(tServiceComponent, testFactoryPidConstructionNameDifferentThanClass)
...
+{
+  std::string configurationPid = "ServiceComponentPid";
+
+  // Get a service reference to ConfigAdmin to create the factory component instances.
+  auto configAdminService =
+      GetInstance<cppmicroservices::service::cm::ConfigurationAdmin>();
+  ASSERT_TRUE(configAdminService) << "GetService failed for ConfigurationAdmin";
+
+  // Create some factory configuration objects.
+  auto const count = 5;
```
jeffdiclemente, line 160: this can be `constexpr`

jeffdiclemente (approved):

jeffdiclemente:
there are [build failures on macosx](https://github.com/CppMicroServices/CppMicroServices/actions/runs/3177563883/jobs/5188664599) that appear to have something to do with generating archives. These fialures look unrelated to your changes however this issue should be logged to circle back on and investigate. @pelliott-mathworks Can you create a new issue? thanks.

Issue #738, opened 2022-10-05
build failures on macOS with generating archives


pelliott-mathworks:
There were build failures in the Build and Test[macOS-12, Configuration=3] build (https://github.com/CppMicroServices/CppMicroServices/actions/runs/3177563883/jobs/5188664599) when I was building PR 731 that weren't related to my changes and seemed to be related with generating archives. Here's a snippet:
[ 68%] Linking CXX executable ../../../bin/usFrameworkTests
ld: in ../../../lib/libTestBundleRA.a(PK), archive member 'PK' with length 0 is not mach-o or llvm bitcode file '../../../lib/libTestBundleRA.a' for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[ 68%] Built target usFrameworkTests

jeffdiclemente:
After taking an initial look, it appears as if something has changed with the macOS linker that no longer likes how we append a ZIP flie to the end of an archive file. Note the error message:

> d: in ../../../lib/libTestBundleRA.a(PK���), archive member 'PK���' with length 0 is not mach-o or llvm bitcode file '../../../lib/libTestBundleRA.a' for architecture x86_64

"PK" is the header signature for ZIP files.

PR #744, opened 2022-10-07
Revert PR #713

jeffdiclemente:
This change is causing an untenable amount of noise in log output. Backing out for now. A better solution to the problem of logging service dependency errors is needed.


kkumar45 (approved):

jeffdiclemente:
build failures are due to Issue #738

PR #754, opened 2022-10-26
configurations using the same pid are not updated properly

jeffdiclemente:
Fixed an issue whereby re-using a configuration pid did not cause the configuration to be sent to the ManagedService/ManagedServiceFactory correctly.

### DIFF
--- a/compendium/ConfigurationAdmin/src/ConfigurationImpl.cpp
+++ b/compendium/ConfigurationAdmin/src/ConfigurationImpl.cpp
@@ -143,7 +143,7 @@ std::shared_future<void> ConfigurationImpl::Remove()
   std::lock_guard<std::mutex> lk{ configAdminMutex };
   if (configAdminImpl) {
     auto fut = configAdminImpl->NotifyConfigurationRemoved(
-      pid, reinterpret_cast<std::uintptr_t>(this));
+      pid, reinterpret_cast<std::uintptr_t>(this), changeCount);
     configAdminImpl = nullptr;
     return fut;
   }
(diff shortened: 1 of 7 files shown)


compendium/ConfigurationAdmin/src/ConfigurationImpl.cpp
```diff
@@ -143,7 +143,7 @@ std::shared_future<void> ConfigurationImpl::Remove()
   std::lock_guard<std::mutex> lk{ configAdminMutex };
   if (configAdminImpl) {
     auto fut = configAdminImpl->NotifyConfigurationRemoved(
-      pid, reinterpret_cast<std::uintptr_t>(this));
+      pid, reinterpret_cast<std::uintptr_t>(this), changeCount);
```
achristoforides, line 146: Not a flag but how does the `reinterpret_cast<std::uintptr_t>(this)` work?

It looks like we're reinterpret-casting a `ConfigurationImpl` to a `std::uintptr_t`? Does this simply mean that rather than using a `ConfigurationImpl*` we are just using `std::uintptr_t` for type-erasure since the address would still be the same?

achristoforides (commented):
LGTM but make changes where appropriate so that tests pass.

pelliott-mathworks (approved):

compendium/ConfigurationAdmin/src/ConfigurationImpl.cpp
jeffdiclemente: essentially yes. I didn't write this bit of code and a cursory look makes me think this re-interpret case isn't necessary and the code could be refactored to remove it.

PR #777, opened 2023-01-09
Fix ub in any

carneyweb:
Fixes undefined behavior in:

template <typename ValueType>  bool  operator==(ValueType const& val) const;

Tested by:
1. enabled UBSAN
2. built
3. ran "bin/usFrameworkTests --gtest_filter="*AnyTest*" and saw the undefined behavior failure
4. updated code with fix
5. re-built
6. re-ran tests as above, no longer see failure


pelliott-mathworks (approved):

framework/include/cppmicroservices/Any.h
```diff
@@ -139,6 +139,12 @@ namespace cppmicroservices
     template <typename ValueType>
     ValueType* any_cast(Any* operand);
```
jeffdiclemente, line 140: you can probably remove this forward declaratoin since it's no longer being called in `operator==`

jeffdiclemente (approved):

framework/include/cppmicroservices/Any.h
carneyweb: Removed.

PR #801, opened 2023-02-08
Fix race with DS service object construction

jeffdiclemente:
A race occurs when DS is constructing a service object with injectable mandatory service dependencies whereby any bound service dependencies could no longer be available because another thread unregistered any of the bound service dependencies.

Given that a ComponentContext is constructed only after all the service dependencies has been satisfied, if during ComponetContext construction any of the bound service dependencies are nullptrs then throw and fail to construct the service object. This fixes a failure mode where a service constructor can be called with a nullptr service object when the service dependency is defined as mandatory and injection is true.

Also noticed that one of the test bundle project was exceeding the path limit on Windows, so shortened the project name to mitigate that problem.


compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -83,21 +84,36 @@ namespace cppmicroservices
                             auto& serviceMap = (*boundServicesCacheHandle)[refName];
                             if (refScope == cppmicroservices::Constants::SCOPE_BUNDLE)
                             {
-                                serviceMap.push_back(bc.GetService(sRefU));
+                                auto svc = bc.GetService(sRefU);
+                                if (svc)
+                                {
+                                    foundAtLeastOneValidBoundService = true;
+                                }
+                                serviceMap.push_back(svc);
```
carneyweb, line 92: Should this line be within the scope of the 'if' block above? See line 101 below.

pelliott-mathworks (approved):

carneyweb (commented):
Looks good except for a possible logic error I noted in a comment. If that's not a problem, things look fine to me.

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
jeffdiclemente: No, I'm maintaining the existing behavior which was to place the service object (even if it was `nullptr`) within the `serviceMap`

achristoforides (approved):

pelliott-mathworks (approved):

PR #808, opened 2023-02-15
Make a small performance improvement

jeffdiclemente:
stop copying the std::pair in the range-based for. This helps improve the speed of RegisterService


achristoforides (approved):

pelliott-mathworks (commented):
Looks good!

PR #809, opened 2023-03-07
701 - clang-tidy improvement for CMakeResourceDependencies.cpp

insi-eb:
CMakeResourceDependencies is often built by CMake in the context of users of CppMicroServices - if you've configured clang-tidy to run in the user project, it complains about the "old-style" constructor and requests a modernized `= default`.

Signed-off-by: Ingmar Sittl <Ingmar.Sittl@elektrobit.com>

I've tested that this produces the exact same output than the previous code on different gcc and clang versions.
I can also create a PR to bring this small improvement to `development`, if you want.


jeffdiclemente:
It looks like this change causes build failures on macOS:

```
/Users/runner/work/CppMicroServices/CppMicroServices/cmake/CMakeResourceDependencies.cpp:29:33: error: defaulted function definitions are a C++11 extension [-Werror,-Wc++11-extensions]
  CMakeResourceDependencies() = default;
                                ^
1 error generated.
```

This error is surprising, I thought we were compiling using c++11? It would appear that CMakeResourceDependencies.cpp isn't being compiled with c++11 or greater.

jeffdiclemente:
> > It looks like this change causes build failures on macOS:
> > ```
> > /Users/runner/work/CppMicroServices/CppMicroServices/cmake/CMakeResourceDependencies.cpp:29:33: error: defaulted function definitions are a C++11 extension [-Werror,-Wc++11-extensions]
> >   CMakeResourceDependencies() = default;
> >                                 ^
> > 1 error generated.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This error is surprising, I thought we were compiling using c++11? It would appear that CMakeResourceDependencies.cpp isn't being compiled with c++11 or greater.
> 
> This piece of code in `cmake\usFunctionEmbedResources.cmake` is what is compiling CMakeResourceDependencies.cpp (only on macOS):
> 
> ```
> add_custom_command(
>         OUTPUT ${_source_output}
>         COMMAND ${CMAKE_CXX_COMPILER} ${_us_resource_cxx_flags} -c ${US_CMAKE_RESOURCE_DEPENDENCIES_CPP} -o stub.o
>         COMMAND ${CMAKE_LINKER} -r -sectcreate __TEXT us_resources ${_zip_archive_name} stub.o -o ${_source_output}
>         DEPENDS ${_zip_archive}
>         WORKING_DIRECTORY ${_zip_archive_path}
>         COMMENT "Linking resources zip file ${_zip_archive} for ${US_RESOURCE_TARGET}"
>         VERBATIM
>        )
> ```
> 
> I haven't checked the compiler flags however I assume they don't contain `std=c++11`

confirmed, the compiler command doesn't have any C++ language version set, i.e. `--std=cXXXX` is not present on the command line. so [the default is c++98](https://opensource.apple.com/source/clang/clang-703.0.29/src/tools/clang/www/cxx_status.html#:~:text=By%20default%2C%20Clang%20builds%20C%2B%2B,11%20features%20accepted%20as%20extensions.)

insi-eb:
> > I haven't checked the compiler flags however I assume they don't contain `std=c++11`
> 
> confirmed, the compiler command doesn't have any C++ language version set, i.e. `--std=cXXXX` is not present on the command line. so [the default is c++98](https://opensource.apple.com/source/clang/clang-703.0.29/src/tools/clang/www/cxx_status.html#:~:text=By%20default%2C%20Clang%20builds%20C%2B%2B,11%20features%20accepted%20as%20extensions.)

a *simple* solution could be to prepend `--std=c++14` (or on `development` to c++17) to _us_resource_cxx_flags here - that also allows users to override the `--std` setting themselves in their CMAKE_CXX_FLAGS to a later standard without problems

insi-eb:
> I can also create a PR to bring this small improvement to `development`, if you want.

@jeffdiclemente: should I also create a corresponding PR for `development`?

jeffdiclemente:
> > I can also create a PR to bring this small improvement to `development`, if you want.
> 
> @jeffdiclemente: should I also create a corresponding PR for `development`?

@insi-eb sure, that sounds good to me. thanks!

PR #833, opened 2023-03-22
Brian decl services

BrianWeed:
Prototype performance improvements (and a few bug fixes)


framework/include/cppmicroservices/ServiceRegistrationBase.h
```diff
@@ -128,7 +128,7 @@ namespace cppmicroservices
          *         of <code>properties</code> exceeds the value returned by
          *         std::numeric_limits<int>::max().
          */
-        void SetProperties(ServiceProperties const& properties);
+        void SetProperties(ServiceProperties && properties);
```
jeffdiclemente, line 131: since this is a public API, this change will break callers. I suggest adding a backwards compatible SetProperties API such that existing call sites don't need to be changed.

jeffdiclemente (changes requested):

PR #834, opened 2023-03-23
ComponentManagermemory reallocation bug fix

pelliott-mathworks:
The ComponentManagerImpl object contained a shared_ptr to the vector of all ComponentManagerImpl objects. When adding an item to this vector, the resizing of the vector caused a memory allocation error. Fixed by removing the vector from the ComponentManagerImpl object. Signed-off-by: The MathWorks, Inc. <pelliott@mathworks.com>

### DIFF
--- a/compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
+++ b/compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
@@ -183,8 +183,8 @@ namespace cppmicroservices
             auto registry = mgr->GetRegistry();
             auto logger = mgr->GetLogger();
             auto configNotifier = mgr->GetConfigNotifier();
-            auto managers = mgr->GetManagers();
-
+            auto managers = std::make_shared<std::vector<std::shared_ptr<cppmicroservices::scrimpl::ComponentManager>>>(
+                mgr->GetRegistry()->GetComponentManagers());
             try
             {
                 auto compManager = std::make_shared<ComponentManagerImpl>(newMetadata,
@@ -192,8 +192,7 @@ namespace cppmicroservices
                                                                           bundle.GetBundleContext(),
                                                                           logger,
                                                                           asyncWorkService,
-                                                                          configNotifier,
-                                                                          managers);
+                                                                          configNotifier);
                 if (registry->AddComponentManager(compManager))
                 {
                     managers->push_back(compManager);
--- a/compendium/DeclarativeServices/test/gtest/TestFactoryPid.cpp
+++ b/compendium/DeclarativeServices/test/gtest/TestFactoryPid.cpp
@@ -165,4 +165,47 @@ namespace test
         auto instances = GetInstances<test::CAInterface>();
         EXPECT_EQ(instances.size(), count);
     }
+    /* test MultipleFactoryConfig. 
+     * This test creates 10 factory objects without waiting for the previous one to complete. 
+     */
+    TEST_F(tServiceComponent, testMultipleFactoryConfig)
+    {
+        std::string configurationPid = "ServiceComponentPid";
+
+        // Start the test bundle containing the factory component. 
+        cppmicroservices::Bundle testBundle = StartTestBundle(
+            "TestBundleDSCA21"); 
+        
+        // Get a service reference to ConfigAdmin to create the factory component instances.
+        auto configAdminService = GetInstance<cppmicroservices::service::cm::ConfigurationAdmin>();
+        ASSERT_TRUE(configAdminService) << "GetService failed for ConfigurationAdmin";
+
+        // Create some factory configuration objects. Don't wait for one to complete before 
+        // creating the next one.
+        constexpr auto count = 10;
+        std::vector<std::shared_future<void>> futures;
+
+        for (int i = 0; i < count; i++)
+        {
+            // Create the factory configuration object
+            auto factoryConfig = configAdminService->CreateFactoryConfiguration(configurationPid);
+
+            // Update the properties for the factory configuration object
+            cppmicroservices::AnyMap props(cppmicroservices::AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS);
+            const std::string instanceId { "instance" + std::to_string(i) };
+            props["uniqueProp"] = instanceId;
+            auto fut = factoryConfig->Update(props);
+            futures.push_back(fut);
+        }
+
+        // Wait for all factory objects to finish updating.
+        for (auto& item : futures) {
+            item.get();
+         } 
+
+        // Request service references to the new component instances. This will
+        // cause DS to construct the factory instances.
+        auto instances = GetInstances<test::CAInterface>();
+        EXPECT_EQ(instances.size(), count);
+    }
 } // namespace test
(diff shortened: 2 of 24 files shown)


jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.hpp
```diff
@@ -76,7 +77,8 @@ namespace cppmicroservices
              */
             ConfigurationNotifier(cppmicroservices::BundleContext const& context,
                                   std::shared_ptr<cppmicroservices::logservice::LogService> logger,
-                                  std::shared_ptr<cppmicroservices::async::AsyncWorkService> asyncWorkService_);
+                                  std::shared_ptr<cppmicroservices::async::AsyncWorkService> asyncWorkService_,
+                                  std::shared_ptr<SCRExtensionRegistry> extensionRegistry_);
```
jeffdiclemente, line 81: the parameter names have the trailing `_` which doesn't fit in with the convention where the members have the trailing `_`. I recommend putting the trailing `_` on the member variable name.

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -183,22 +186,32 @@ namespace cppmicroservices
...
                 if (registry->AddComponentManager(compManager))
                 {
-                    managers->push_back(compManager);
-                    compManager->Initialize();
-                }
+                    auto extension = extensionRegistry->Find(bundle.GetBundleId());
+                    if (extension)
+                    {
+                        extension->AddComponentManager(compManager);
+                        compManager->Initialize();
```
jeffdiclemente, line 203: what should happen if `extension->AddComponentManager(compManager);` fails? should `compManager->Initialize();` still be called?

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -183,22 +186,32 @@ namespace cppmicroservices
...
-                                                                          configNotifier,
-                                                                          managers);
+                                                                          configNotifier);
                 if (registry->AddComponentManager(compManager))
                 {
-                    managers->push_back(compManager);
-                    compManager->Initialize();
-                }
+                    auto extension = extensionRegistry->Find(bundle.GetBundleId());
+                    if (extension)
```
jeffdiclemente, line 200: use [initializers for if statements](https://en.cppreference.com/w/cpp/language/if)

compendium/DeclarativeServices/test/gtest/TestFactoryPid.cpp
```diff
@@ -160,6 +160,49 @@ namespace test
...
+            // Update the properties for the factory configuration object
+            cppmicroservices::AnyMap props(cppmicroservices::AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS);
+            const std::string instanceId { "instance" + std::to_string(i) };
+            props["uniqueProp"] = instanceId;
+            auto fut = factoryConfig->Update(props);
+            futures.push_back(fut);
+        }
+
+        // Wait for all factory objects to finish updating.
+        for (const auto& item : futures) {
```
jeffdiclemente, line 202: clang formatting: `const auto&` -> `auto const&`

jeffdiclemente (changes requested):

jeffdiclemente (changes requested):
@pelliott-mathworks please add a CHANGELOG.rst update for this change

compendium/DeclarativeServices/src/SCRActivator.cpp
```diff
@@ -143,30 +144,21 @@ namespace cppmicroservices
...
+            //create the extension which will load the components
+            if (!bundleRegistry->Find(bundle.GetBundleId()))
             {
                 logger->Log(SeverityLevel::LOG_DEBUG, "Creating SCRBundleExtension ... " + bundle.GetSymbolicName());
                 try
                 {
                     auto const& scrMap = ref_any_cast<cppmicroservices::AnyMap>(headers.at(SERVICE_COMPONENT));
-                    auto ba = std::make_unique<SCRBundleExtension>(bundle,
-                                                                   scrMap,
+                    auto ba = std::make_shared<SCRBundleExtension>(bundle,
```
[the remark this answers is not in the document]
pelliott-mathworks: If this is a unique_ptr then I would have to use std::move in the bundleRegistry->Add call and it wouldn't be available for the ba->Initialize call. 

The reason I had to split the functionality from the old SCRBundleExtension constructor into a constructor and an Initialize method is because factory components can be added to ConfigurationAdmin before a bundle is started. When the bundle is started, DS will find the configuration objects and call CreateFactoryComponent to create them. For CreateFactoryComponent to work it has be able to find the SCRBundleExtension object in the SCRExtensionRegistry 
 and then call the SCRBundleExtension::AddComponentManager method to add the ComponentManagerImpl objects to the SCRBundleExtension::managers map. 

So the new constructor creates the managers map. SCRActivator adds the SCRBundleExtension object to the SCRExtensionRegistry. At this point the Initialize method can be called to create the ComponentConfiguration objects and if CreateFactoryComponent is called it will work fine.

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
pelliott-mathworks: No but the only way extension->AddComponentManager can fail is if it throws an exception. Since this is in a try block, control will jump to the catch.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -208,6 +220,10 @@ namespace cppmicroservices
             {
                 throw;
             }
+            catch (std::runtime_error const&)
+            {
+                throw;
+            }
```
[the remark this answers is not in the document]
jeffdiclemente: what problem is being solved by converting the  `std::exception` into a `std::runtime_error`? This conversion seems unnecessary.

achristoforides (approved):

carneyweb (approved):

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -183,21 +186,29 @@ namespace cppmicroservices
...
+                        extension->AddComponentManager(compManager);
+                        compManager->Initialize();
+                    }
+                    else
+                    {
+                        logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
+                                    "Failed to find ComponentManager with name " + newMetadata->name
+                                        + " from bundle with Id "
+                                        + std::to_string(bundleContext.GetBundle().GetBundleId()));
+                        throw std::exception();
```
jeffdiclemente, line 210: I don't understand why there is a `throw` here. This will cause execution to proceed to the `catch (std::exception const&)` block and then log a message which is misleading (i.e. ComponentManager creation did not fail). I think the `throw` should be removed.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #841, opened 2023-04-19
Fixes #840: removes manual reference counting

### LINKED ISSUE #840: Remove manual reference counting from ServiceRegistrationBasePrivate and ServiceReferenceBasePrivate
Currently, destruction of ServiceRegistrationBasePrivate and ServiceReferenceBasePrivate occurs when manually tracked reference counters fall to 0. 

This should be updated to use built in types, like std::shared_ptr

tcormackMW:
This removes manual reference counting from ServiceRegistrationBasePrivate and ServiceReferenceBasePrivate and offloads some of ServiceRegistrationBasePrivate's data to be store in the new class ServiceRegistrationCoreInfo


framework/src/service/ServiceRegistrationCoreInfo.cpp
```diff
@@ -0,0 +1,56 @@
...
+        , bundle(bundle_->shared_from_this())
+        , properties(std::move(props))
+        , available(true)
+        , unregistering(false)
+    {
+    }
+
+    ServiceRegistrationCoreInfo::~ServiceRegistrationCoreInfo()
+    {
+        bundle.reset();
```
carneyweb, line 46: These are unnecessary. Without them the destructor could be defaulted.

carneyweb (changes requested):
Fix it, then ship it!

framework/src/service/ServiceRegistrationCoreInfo.cpp
tcormackMW: Fixed in commit "ServiceRegistrationCoreInfo now default" destructor(https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/e0df8235bc17141bfddd6abfbd8c549c6269b535)

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -43,34 +43,29 @@ namespace cppmicroservices
...
+        : registration(reg)
     {
-        if (registration)
-            ++registration->ref;
-    }
-
-    ServiceReferenceBasePrivate::~ServiceReferenceBasePrivate()
-    {
-        if (registration && !--registration->ref)
+        if (!reg.expired())
```
pelliott-mathworks, line 49: Can registration be a nullptr? If so, should be checked before using. The old code checked before using which implies it could be a nullptr

framework/src/service/ServiceRegistrationBase.cpp
```diff
@@ -175,27 +160,27 @@ namespace cppmicroservices
...
-            auto oldRankAny = d->properties.Value_unlocked(Constants::SERVICE_RANKING).first;
+            auto oldRankAny = d->coreInfo->properties.Value_unlocked(Constants::SERVICE_RANKING).first;
             if (!oldRankAny.Empty())
             {
                 // since the old ranking is extracted from existing service properties
                 // stored in the service registry, no need to type check before casting
                 old_rank = any_cast<int>(oldRankAny);
             }
-            d->properties = Properties(AnyMap(std::move(propsCopy)));
+            (d->coreInfo->properties) = Properties(AnyMap(std::move(propsCopy)));
```
pelliott-mathworks, line 170: Are these parentheses necessary?

framework/include/cppmicroservices/ServiceRegistrationBase.h
```diff
@@ -196,11 +196,11 @@ namespace cppmicroservices
          */
         ServiceRegistrationBase();
 
-        ServiceRegistrationBase(ServiceRegistrationBasePrivate* registrationPrivate);
+        ServiceRegistrationBase(std::shared_ptr<ServiceRegistrationBasePrivate> registrationPrivate);
 
         ServiceRegistrationBase(BundlePrivate* bundle, InterfaceMapConstPtr const& service, Properties&& props);
 
-        ServiceRegistrationBasePrivate* d { nullptr };
+        std::shared_ptr<ServiceRegistrationBasePrivate> d { nullptr };
```
pelliott-mathworks, line 203: I think the nullptr is unecessary. {} will do the same thing

framework/src/service/ServiceReferenceBase.cpp
```diff
@@ -89,36 +80,36 @@ namespace cppmicroservices
...
     }
 
     std::vector<Bundle>
     ServiceReferenceBase::GetUsingBundles() const
     {
         std::vector<Bundle> bundles;
-        auto l = d.load()->registration->Lock();
-        US_UNUSED(l);
-        for (auto& iter : d.load()->registration->dependents)
+        for (auto& iter : (std::atomic_load(&d)->coreInfo->dependents))
```
pelliott-mathworks, line 112: Can this be auto const& iter?

pelliott-mathworks (changes requested):
Well done!

framework/src/service/ServiceRegistrationBase.cpp
tcormackMW: No they aren't, thanks!

Updated in "Updated based on Patty's comments"(https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/7d7bdd60e332b54a5954fe8823ccc2212b5a46d3)

framework/include/cppmicroservices/ServiceRegistrationBase.h
tcormackMW: Yep, I have changed it to just "std::shared_ptr<ServiceRegistrationBasePrivate? d;" in "Updated based on Patty's comments"(https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/7d7bdd60e332b54a5954fe8823ccc2212b5a46d3)

pelliott-mathworks (approved):

jeffdiclemente (changes requested):

framework/src/bundle/BundleContext.cpp
```diff
@@ -274,8 +274,7 @@ namespace cppmicroservices
         {
             try
             {
-                sref.d.load()->UngetService(b.lock(), true);
-                //sref.d.load()->UngetService(b.lock(), true);
+                std::atomic_load(&sref.d)->UngetService(b.lock(), true);
```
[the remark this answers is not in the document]
tcormackMW: fixed in "changed to custom atomic load"

https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/7def5ac1deeaaf7748f2fd416edb991f4d9bba6d

framework/src/service/LockSet.cpp
```diff
@@ -0,0 +1,49 @@
...
+            regL = reg->Lock();
+        }
+        coreInfoL = coreInfo->Lock();
+    }
+#else
+    LockSet::LockSet(std::shared_ptr<ServiceRegistrationBasePrivate>,
+                     std::shared_ptr<ServiceRegistrationCoreInfo>)
+    {
+    }
+#endif
```
jeffdiclemente, line 42: I don't understand the purpose of the `LockSet` class. the `Lock()` function already uses mutexes if US_ENABLE_THREADING_SUPPORT is set and doesn't if the macro is unset. This class seems redundant.

framework/src/util/Utils.h
```diff
@@ -81,6 +81,16 @@ namespace cppmicroservices
...
 
+    template <typename T>
+    std::shared_ptr<T> atomic_load_tSafe(std::shared_ptr<T> ptr)
+    {
+#ifdef US_ENABLE_THREADING_SUPPORT
+        return std::atomic_load(&ptr);
+#else
+        return ptr;
+#endif
+    }
```
jeffdiclemente, line 92: this is already implemented by `cppmicroservices::detail::Atomic<T>` in `Threads.h`. I don't think you need another implementation.

framework/src/service/LockSet.cpp
tcormackMW: This is essentially a way of linking the locking of registrationBasePrivate and coreInfo. In the current implementation, the registrationBasePrivate is Locked whenever any data in it is accessed (namely properties, service, etc). That is no longer possible as that data, that originally was in the registrationBasePrivate, is now in CoreInfo. And, as the ServiceRegistrationBasePrivate dies independently of the coreInfo now, we cannot only rely on the locking of the registration and so we must also lock CoreInfo. This was written so that those two can be coupled without continually having to replicate this code all across the code base whenever we need to lock both. 

I am not sure this answered your question, but if it didn't let me know.

framework/src/service/LockSet.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#ifdef US_ENABLE_THREADING_SUPPORT
+    LockSet::LockSet(std::shared_ptr<ServiceRegistrationBasePrivate> reg,
+                     std::shared_ptr<ServiceRegistrationCoreInfo> coreInfo)
+    {
+        if (reg != nullptr)
+        {
+            regL = reg->Lock();
+        }
+        coreInfoL = coreInfo->Lock();
```
jeffdiclemente, line 35: `reg` is checked for `nullptr` but `coreInfoL` is not? should it be?

framework/src/service/ServiceRegistrationBase.cpp
```diff
@@ -320,22 +306,25 @@ namespace cppmicroservices
...
+            d->coreInfo->prototypeServiceInstances.clear();
+            d->coreInfo->bundleServiceInstance.clear();
+
             d->reference = nullptr;
-            d->unregistering = false;
+            d->coreInfo->unregistering = false;
         }
     }
 
+    std::shared_ptr<LockSet> ServiceRegistrationBase::GetLocks() const{
```
jeffdiclemente, line 323: I'd like to see this function renamed to something more descriptive. this function acquires locks an RAII type object, but the function name does not convey any of this to the user.

jeffdiclemente (changes requested):

framework/src/service/LockSet.cpp
tcormackMW: `reg` is checked because reg is not owned by `ServiceReferenceBasePrivate` and so it can die with `ServiceReferenceBasePrivate` still with a `weak_ptr` to it. So it has to be checked to confirm that it still exists. In contrast, `coreInfo` is owned by both the `ServiceReferenceBasePrivate` and the `ServiceRegistrationBasePrivate` so it is guaranteed to be alive. That is why this check was made, but it would also not cause problems to check both for consistency. The only caveat to generalizing this function, i.e. either lock can be passed in as the first or second argument is then you will get into deadlocks where one call is made with `LockSet(reg, coreInfo)` and the other will be made with `LockSet(coreInfo, reg)`.

framework/src/service/LockSet.cpp
```diff
@@ -0,0 +1,49 @@
+/*=============================================================================
```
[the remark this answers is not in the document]
tcormackMW: Ok will do.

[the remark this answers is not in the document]
tcormackMW: Moved in latest commit 

https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/5fc7802a40e3c35d2957b0562f56e73fb2744c23

framework/src/service/ServiceRegistrationBase.cpp
tcormackMW: Changed it to `LockRegistration`... This gave me more insight into the fact that this was a call to lock the registration object generally. I'm not sure if this is clearer given that you could also say this sounds like `registration.Lock()`, which is only part of it. If anyone has naming ideas, I'm definitely open to that

https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/5fc7802a40e3c35d2957b0562f56e73fb2744c23

framework/src/service/ServiceRegistrationCoreInfo.h
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * \ingroup MicroServices
+     */
+    class ServiceRegistrationCoreInfo : public detail::MultiThreaded<>
+    {
+
+      public:
+        ServiceRegistrationCoreInfo(BundlePrivate* bundle_, InterfaceMapConstPtr service, Properties&& props);
+        ~ServiceRegistrationCoreInfo();
```
jeffdiclemente, line 46: `= default;` can be done here instead of in the .cpp file

jeffdiclemente (changes requested):

framework/include/cppmicroservices/ServiceReferenceBase.h
```diff
@@ -232,13 +233,16 @@ namespace cppmicroservices
...
         void SetInterfaceId(std::string const& interfaceId);
 
         // This class is not thread-safe, but we support thread-safe
-        // copying and assignment.
-        std::atomic<ServiceReferenceBasePrivate*> d;
+        // copying and assignment
+        // This was changed to a std::shared_ptr and is accessed through
+        // Load(), but when this repository uses c++20,
+        // this will be changed to std::atomic<shared_ptr>
+        cppmicroservices::detail::Atomic<std::shared_ptr<ServiceReferenceBasePrivate>> d;
```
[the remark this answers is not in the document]
tcormackMW: I think this is not possible as shared_ptr are not supported for std::atomic until c++ 20 @jeffdiclemente

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -198,40 +200,43 @@ namespace cppmicroservices
...
 
         {
-            auto l = registration->Lock();
+            auto reg = registration.lock();
+            if (!reg)
+                return s;
+            auto l = LockRegistration();
             US_UNUSED(l);
-            if (!registration->available)
+            if (!coreInfo->available)
```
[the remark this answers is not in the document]
tcormackMW: Fixed in https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/446305785ce7eb1d76ffcbf05f3e5342602c3cf7#diff-9a7aa056e8a28a47886872400e1348255ec497ba87f60b6513fd98c2e3a1b1e7L212

framework/src/util/RegistrationLocks.h
```diff
@@ -0,0 +1,43 @@
...
+     * \ingroup MicroServices
+     */
+    class RegistrationLocks
+    {
+        public:
+            RegistrationLocks(std::shared_ptr<ServiceRegistrationBasePrivate> reg, std::shared_ptr<ServiceRegistrationCoreInfo> coreInfo);
+            
+            ~RegistrationLocks();
+        private:
+#ifdef US_ENABLE_THREADING_SUPPORT
```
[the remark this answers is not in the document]
tcormackMW: fixed in https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/446305785ce7eb1d76ffcbf05f3e5342602c3cf7

framework/src/util/RegistrationLocks.cpp
```diff
@@ -0,0 +1,49 @@
...
+        coreInfoL = coreInfo->Lock();
+    }
+#else
+    RegistrationLocks::RegistrationLocks(std::shared_ptr<ServiceRegistrationBasePrivate>,
+                     std::shared_ptr<ServiceRegistrationCoreInfo>)
+    {
+    }
+#endif
+
+    RegistrationLocks::~RegistrationLocks() = default;
```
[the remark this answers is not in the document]
tcormackMW: fixed in https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/446305785ce7eb1d76ffcbf05f3e5342602c3cf7

framework/src/service/ServiceRegistrationCoreInfo.h
tcormackMW: fixed in https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/446305785ce7eb1d76ffcbf05f3e5342602c3cf7

framework/include/cppmicroservices/ServiceRegistrationBase.h
```diff
@@ -196,11 +197,13 @@ namespace cppmicroservices
          */
         ServiceRegistrationBase();
 
-        ServiceRegistrationBase(ServiceRegistrationBasePrivate* registrationPrivate);
+        ServiceRegistrationBase(std::shared_ptr<ServiceRegistrationBasePrivate> registrationPrivate);
 
         ServiceRegistrationBase(BundlePrivate* bundle, InterfaceMapConstPtr const& service, Properties&& props);
 
-        ServiceRegistrationBasePrivate* d { nullptr };
+        std::shared_ptr<RegistrationLocks> LockRegistration() const;
```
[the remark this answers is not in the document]
tcormackMW: fixed in https://github.com/CppMicroServices/CppMicroServices/pull/841/commits/446305785ce7eb1d76ffcbf05f3e5342602c3cf7

framework/src/service/ServiceReferenceBase.cpp
```diff
@@ -28,56 +28,49 @@
...
+        d.Exchange(std::make_shared<ServiceReferenceBasePrivate>(std::weak_ptr<ServiceRegistrationBasePrivate>()));
         return *this;
     }
 
-    ServiceReferenceBase::~ServiceReferenceBase()
-    {
-        if (!--d.load()->ref)
-            delete d.load();
-    }
+    ServiceReferenceBase::~ServiceReferenceBase() {}
```
carneyweb, line 66: = default

?

framework/src/service/ServiceRegistrationBase.cpp
```diff
@@ -39,61 +40,50 @@ US_MSVC_DISABLE_WARNING(4503) // decorated name length exceeded, name was trunca
...
-            ++d->ref;
     }
 
     ServiceRegistrationBase::ServiceRegistrationBase(BundlePrivate* bundle,
                                                      InterfaceMapConstPtr const& service,
                                                      Properties&& props)
-        : d(new ServiceRegistrationBasePrivate(bundle, service, std::move(props)))
+        : d(std::make_shared<ServiceRegistrationBasePrivate>(bundle, service, std::move(props)))
     {
+        d->CreateReference();
```
carneyweb, line 60: What does this addition do?

framework/src/util/ServiceRegistrationLocks.h
```diff
@@ -0,0 +1,43 @@
...
+#include "ServiceRegistrationBasePrivate.h"
+#include "ServiceRegistrationCoreInfo.h"
+
+namespace cppmicroservices
+{
+
+    /**
+     * \ingroup MicroServices
+     */
+    class ServiceRegistrationLocks
```
carneyweb, line 29: A comment describing the purpose of this class would be useful.

carneyweb (changes requested):
Just need a comment describing the purpose and usage ofthe ServiceRegistrationLocks class.

framework/src/service/ServiceRegistrationBase.cpp
tcormackMW: To create the ServiceReference, you need to pass in a ServiceRegistrationBasePrivate shared_ptr, but that cannot be passed until the ServiceRegistrationBasePrivate object has been fully instantiated. So first we make the shared_ptr<ServiceRegistrationBasePrivate> and then we call createReference once that shared_ptr is finished being made to call shared_from_this when making the ServiceReferenceBase object for that ServiceRegistrationBasePrivate.

carneyweb: Please add that info to a comment  there.

framework/src/service/ServiceRegistrationCoreInfo.h
```diff
@@ -0,0 +1,100 @@
...
+     */
+    class ServiceRegistrationCoreInfo : public detail::MultiThreaded<>
+    {
+
+      public:
+        ServiceRegistrationCoreInfo(BundlePrivate* bundle, InterfaceMapConstPtr service, Properties&& props);
+        ~ServiceRegistrationCoreInfo() = default;
+
+        ServiceRegistrationCoreInfo(ServiceRegistrationCoreInfo const&) = delete;
+        ServiceRegistrationCoreInfo& operator=(ServiceRegistrationCoreInfo const&) = delete;
```
jeffdiclemente, line 49: a `= default;` destructor stops the compiler from generating implicit move operators. Per https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all, the move constructor and assignment need to be defined as `= delete` as well.

see https://en.cppreference.com/w/cpp/language/rule_of_three for more detailed information.

framework/src/bundle/BundleRegistry.cpp
```diff
@@ -249,8 +249,10 @@ namespace cppmicroservices
                 {
                     // Wait for the install thread to notify this thread that it is safe
                     // to install the current bundle
-                    auto& p = initialBundleInstallMap[location];
                     l.UnLock();
+#ifdef US_ENABLE_THREADING_SUPPORT
```
jeffdiclemente, line 253: the `#ifdef` here changes the behavior in the case where threading is enabled, `l.UnLock()` is in the wrong place.

jeffdiclemente (changes requested):

framework/src/bundle/BundleRegistry.cpp
jeffdiclemente: thinking about this more and digging into the problem, it looks like the gnu C++ library on the github build machines we use does not run into the linker problem we see on Debian 11:

```
CMakeFiles/CppMicroServices.dir/src/bundle/BundleRegistry.cpp.o: in function `std::cv_status std::condition_variable::__wait_until_impl<std::chrono::duration<long, std::ratio<1l, 1000000000l> > >(std::unique_lock<std::mutex>&, std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l> > > const&)':
/usr/include/c++/10/condition_variable:209: undefined reference to `pthread_cond_clockwait'
```

This PR has nothing to do with the abovementioned linker error, so I would prefer to treat this failure separately in another pull request. The fact that BundleRegistry.cpp uses condition_variables and mutexes means that it should be properly guarded when building without threading support, which is currently does not.

To be honest, I would prefer to eliminate the "no threading" build configuration entirely however that's a larger change with an unknown level of downstream impact that I don't currently have the bandwidth to address.

framework/src/service/ServiceRegistrationCoreInfo.h
```diff
@@ -0,0 +1,100 @@
...
+
+#include <atomic>
+
+namespace cppmicroservices
+{
+
+    /**
+     * \ingroup MicroServices
+     */
+    class ServiceRegistrationCoreInfo : public detail::MultiThreaded<>
```
jeffdiclemente, line 41: mark as `final` if no other classes should derive from it.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

framework/src/service/ServiceReferenceBase.cpp
tcormackMW: fixed

framework/src/util/ServiceRegistrationLocks.h
tcormackMW: This has been added to this class

framework/src/service/ServiceRegistrationCoreInfo.h
tcormackMW: This has been changed and fixed

framework/src/service/ServiceRegistrationCoreInfo.h
tcormackMW: This has been changed and fixed

framework/src/bundle/BundleRegistry.cpp
tcormackMW: The changes have been reverted

carneyweb (approved):

Issue #733, opened 2022-10-04
sporadic data race during framework shutdown


jeffdiclemente:
data race flagged by ThreadSanitizer build:
```
10: [ RUN      ] BundleTest.TestFrameworkAccessDuringFrameworkShutdown
10: ==================
10: WARNING: ThreadSanitizer: data race (pid=12064)
10:   Read of size 8 at 0x7b6400004b68 by thread T71:
10:     #0 cppmicroservices::CoreBundleContext::GetDataStorage[abi:cxx11](long) const <null> (libCppMicroServices.so.3.7.3+0x17d5a7)
10:     #1 cppmicroservices::BundlePrivate::BundlePrivate(cppmicroservices::CoreBundleContext*, std::shared_ptr<cppmicroservices::BundleArchive> const&) <null> (libCppMicroServices.so.3.7.3+0x156873)
10:     #2 std::__shared_ptr<cppmicroservices::BundlePrivate, (__gnu_cxx::_Lock_policy)2>::__shared_ptr<std::allocator<cppmicroservices::BundlePrivate>, cppmicroservices::CoreBundleContext*&, std::shared_ptr<cppmicroservices::BundleArchive> const&>(std::_Sp_alloc_shared_tag<std::allocator<cppmicroservices::BundlePrivate> >, cppmicroservices::CoreBundleContext*&, std::shared_ptr<cppmicroservices::BundleArchive> const&) <null> (libCppMicroServices.so.3.7.3+0x1661b0)
10:     #3 cppmicroservices::BundleRegistry::Install0(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<cppmicroservices::BundleResourceContainer> const&, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&, cppmicroservices::AnyMap const&) <null> (libCppMicroServices.so.3.7.3+0x15f51f)
10:     #4 cppmicroservices::BundleRegistry::Install(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, cppmicroservices::BundlePrivate*, cppmicroservices::AnyMap const&) <null> (libCppMicroServices.so.3.7.3+0x15e4ab)
[...]

PR #845, opened 2023-05-03
Remove CoreBundleContext.dataStorage clear

ShivamNegi:
As per issue #733 this commit removes the code for 'clear()' `dataStorage` to solve the data-race warning generated by ThreadSanitizer.

Analysis:
The Thread Sanitizer data-race warning is caused when two threads simultaneously try to access the `dataStorage` variable. Thread T71 (in the issue log) tries to perform a READ operation on the variable while thread T74 performs a write.

Solution:
As the `dataStorage` string variable is only set at the time of initialization of the framework and will not be changed during the lifecycle. There is no need to perform a `clear()` operation during `Shutdown`. Thus, this line is removed solving the possible race-condition.

### DIFF
--- a/framework/src/bundle/CoreBundleContext.cpp
+++ b/framework/src/bundle/CoreBundleContext.cpp
@@ -239,7 +239,6 @@ namespace cppmicroservices
         listeners.Clear();
         resolver.Clear();
 
-        dataStorage.clear();
         storage->Close();
     }
 


jeffdiclemente (approved):

jeffdiclemente:
Looks good. the thread sanitizer build found a number of other issues, unrelated to this change, which I will log as issues to be fixed later.

thread sanitizer build log: https://github.com/CppMicroServices/CppMicroServices/actions/runs/4872584091/jobs/8690974464?pr=845

jeffdiclemente:
make sure to update the [CHANGELOG](https://github.com/CppMicroServices/CppMicroServices/blob/development/CHANGELOG.rst) and add this fix. thanks.

Issue #771, opened 2022-12-15
Don't use std::move on constant variables.


jeffdiclemente:
Tracking issue for:
- [ ] https://github.com/CppMicroServices/CppMicroServices/security/code-scanning/108
- [ ] https://github.com/CppMicroServices/CppMicroServices/security/code-scanning/9
- [ ] https://github.com/CppMicroServices/CppMicroServices/security/code-scanning/8
- [ ] https://github.com/CppMicroServices/CppMicroServices/security/code-scanning/7

PR #848, opened 2023-05-04
Removing std::move use with const object [ci skip]

thrtouati:
#771 
This warning indicates that the use of std::move is not consistent with how it is intended to be used. When called on a const object, std::move does not perform a move operation but instead returns a copy of the object, which is likely not the developer's intent.

To fix this, I simply removed its use from the two files where the warning appeared:
- ConfigurationListener.hpp
- ComponentConfigurationImpl.hpp
I run tests locally (in the usDeclarativeServicesTest module) that use this part of the code, and the tests seem to be working as before. 

I also used [ci skip] since the changes are minor

### DIFF
--- a/compendium/CM/include/cppmicroservices/cm/ConfigurationListener.hpp
+++ b/compendium/CM/include/cppmicroservices/cm/ConfigurationListener.hpp
@@ -62,10 +62,10 @@ namespace cppmicroservices
                                    const ConfigurationEventType type,
                                    const std::string factoryPid,
                                    const std::string pid)
-                    : configAdmin(std::move(configAdmin))
+                    : configAdmin(configAdmin)
                     , type(type)
-                    , factoryPid(std::move(factoryPid))
-                    , pid(std::move(pid))
+                    , factoryPid(factoryPid)
+                    , pid(pid)
                 {
                 }
 
--- a/compendium/CM/include/cppmicroservices/cm/ConfigurationListener.hpp
+++ b/compendium/CM/include/cppmicroservices/cm/ConfigurationListener.hpp
@@ -58,14 +58,14 @@ namespace cppmicroservices
             class ConfigurationEvent
             {
               public:
-                ConfigurationEvent(ServiceReference<ConfigurationAdmin> const configAdmin,
+                ConfigurationEvent(ServiceReference<ConfigurationAdmin> configAdmin,
                                    const ConfigurationEventType type,
-                                   const std::string factoryPid,
-                                   const std::string pid)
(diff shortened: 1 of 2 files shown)


carneyweb (approved):

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.hpp
```diff
@@ -56,7 +56,7 @@ namespace cppmicroservices
         {
             ListenerToken(std::string pid, const ListenerTokenId tokenId)
                 : pid(std::move(pid))
-                , tokenId(std::move(tokenId))
```
jeffdiclemente, line 59: should the parameter be non-const? is it better to move ` tokenId`?

compendium/CM/include/cppmicroservices/cm/ConfigurationListener.hpp
```diff
@@ -62,10 +62,10 @@ namespace cppmicroservices
                                    const ConfigurationEventType type,
                                    const std::string factoryPid,
                                    const std::string pid)
-                    : configAdmin(std::move(configAdmin))
+                    : configAdmin(configAdmin)
                     , type(type)
-                    , factoryPid(std::move(factoryPid))
-                    , pid(std::move(pid))
+                    , factoryPid(factoryPid)
+                    , pid(pid)
```
jeffdiclemente, line 68: is there benefit to having the parameter's be non-const and moving them?

pelliott-mathworks (approved):

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.hpp
BrianWeed: Since ListenerTokenId is only a uint64, it's trivial to copy, so no move is necessary.
For more complex types, however, it would be better to remove the const and keep the move.

compendium/CM/include/cppmicroservices/cm/ConfigurationListener.hpp
BrianWeed: For these, we should remove the const, instead of the move

BrianWeed: Expensive to copy sync-params should be non-const by-value, and then moved

thrtouati:
I updated the change as per the comments, let me know if you have any additional feedback.

jeffdiclemente (approved):

PR #863, opened 2023-05-19
Fixes #852: adds a [[nodiscard]] to BundleContext::RegisterService

### LINKED ISSUE #852: Add [[nodiscard]] tag to `BundleContext::RegisterService`
Add a [[nodiscad]] attribute to the `BundleContext::RegisterService` function call. This will force users to hold on to their `ServiceRegistrationU` object encouraging them to use scoping rules to control the registration lifespan of services in future releases.

tcormackMW:
This change edits the BundleContext class to force the user to hold on to the `ServiceRegistrationU` object returned by `RegisterService`. This aims to change expected behavior for handling objects to, in the future, allow scoping rules to help determine the registered lifetime of services within the framework.

### DIFF
--- a/framework/src/bundle/BundleContext.cpp
+++ b/framework/src/bundle/BundleContext.cpp
@@ -203,7 +203,7 @@ namespace cppmicroservices
         return bus;
     }
 
-    [[nodiscard]] ServiceRegistrationU
+    ServiceRegistrationU
     BundleContext::RegisterService(InterfaceMapConstPtr const& service, ServiceProperties const& properties)
     {
         if (!d)
--- a/framework/src/bundle/BundleContext.cpp
+++ b/framework/src/bundle/BundleContext.cpp
@@ -203,7 +203,7 @@ namespace cppmicroservices
         return bus;
     }
 
-    ServiceRegistrationU
+    [[nodiscard]] ServiceRegistrationU
     BundleContext::RegisterService(InterfaceMapConstPtr const& service, ServiceProperties const& properties)
     {
         if (!d)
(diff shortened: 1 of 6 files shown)


framework/test/bench/ServiceRegistryTest.cpp
```diff
@@ -147,11 +149,12 @@ BENCHMARK_DEFINE_F(ServiceRegistryFixture, FindServices)
     auto regCount = state.range(0);
     auto interfaceCount = state.range(1);
     auto interfaceMap = MakeInterfaceMapWithNInterfaces(interfaceCount);
+    std::vector<ServiceRegistrationU> regs;
 
     for (auto i = regCount; i > 0; --i)
     {
         InterfaceMapPtr iMapCopy(std::make_shared<InterfaceMap>(*interfaceMap));
-        fc.RegisterService(iMapCopy);
+        regs.push_back(fc.RegisterService(iMapCopy));
```
jeffdiclemente, line 157: use `emplace_back` instead of `push_back` for better performance (allows constructing the element in-place).

framework/src/bundle/BundleContext.cpp
```diff
@@ -203,7 +203,7 @@ namespace cppmicroservices
         return bus;
     }
 
-    ServiceRegistrationU
+    [[nodiscard]] ServiceRegistrationU
```
jeffdiclemente, line 206: I think `[[nodiscard]]` can be removed here, it only needs to be in the function declaration.

jeffdiclemente (changes requested):

PR #874, opened 2023-06-27
#Issue873: Redundant calls to .Load()

tcormackMW:
Within ServiceReferenceBase, there are redundant calls to `.Load()` which are not required given that the class is not guarantied to be thread safe except for copying and assignment. These calls are causing significant performance losses.

### DIFF
--- a/framework/src/service/ServiceReferenceBase.cpp
+++ b/framework/src/service/ServiceReferenceBase.cpp
@@ -68,9 +68,10 @@ namespace cppmicroservices
     Any
     ServiceReferenceBase::GetProperty(std::string const& key) const
     {
-        auto l = d.Load()->coreInfo->properties.Lock();
+        auto temp = d.Load();
+        auto l = temp->coreInfo->properties.Lock();
         US_UNUSED(l);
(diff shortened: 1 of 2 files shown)


framework/src/service/ServiceReferenceBase.cpp
```diff
@@ -68,9 +68,10 @@ namespace cppmicroservices
     Any
     ServiceReferenceBase::GetProperty(std::string const& key) const
     {
-        auto l = d.Load()->coreInfo->properties.Lock();
+        auto temp = d.Load();
```
jeffdiclemente, line 71: please use a more descriptive name other than `temp`/`tmp`.

jeffdiclemente (changes requested):

framework/src/service/ServiceReferenceBase.cpp
tcormackMW: Yes. I will do that. Oversight on my part

jeffdiclemente (approved):

carneyweb (approved):

PR #875, opened 2023-06-28
Fix #872

### LINKED ISSUE #872: Static Member Variable in ServiceHooksTest causes failure on gtest_repeat
Static member variable `std::vector<int> TestServiceListenerHook::ordering` causes failures when repeating test cases. 

The failing test cases are:

`[  FAILED  ] 4 tests, listed below:`
`[  FAILED  ] ServiceHooksTest.TestListenerHook`
`[  FAILED  ] ServiceHooksTest.TestFindHook`
`[  FAILED  ] ServiceHooksTest.TestEventListenerHook`
`[  FAILED  ] MultipleListenersTest.testMultipleListeners`

Command to run is:

`./usFrameworkTests --gtest_filter="ServiceHooksTest.*" --gtest_repeat=2`
`./usFrameworkTests --gtest_filter="MultipleListenersTest.testMultipleListeners*" --gtest_repeat=2`

tcormackMW:
Used the test destructor to clear the static/namespace variables. These are required for the ServiceHooksTest because multiple listener instances must access same object, and for the MultipleListenersTest, callbacks defined in the namespace access the count variable. 

In each case, a destructor run after each test case of a certain class will clear/reset each of these variables.


framework/test/gtest/MultipleListenersTest.cpp
```diff
@@ -246,7 +246,7 @@ namespace
       public:
         MultipleListenersTest() : framework(FrameworkFactory().NewFramework()) {}
 
-        ~MultipleListenersTest() override = default;
+        ~MultipleListenersTest() override { count = 0; }
```
carneyweb, line 249: Curious as to the reason for this change.

carneyweb (commented):
Just curious about the removal of "= default".

framework/test/gtest/MultipleListenersTest.cpp
tcormackMW: @carneyweb count was a variable declared in the namespace that, when run using `gtest_repeat` was not getting cleared/reset after each run and so we were erroring because it was a running sum as opposed to a sum being reset each time. For that reason, I am using the destructor for the test to reset that count variable for each iteration of the test

carneyweb: I'd suggest a comment that says exactly that :-)

carneyweb (approved):

framework/test/gtest/ServiceHooksTest.cpp
```diff
@@ -381,6 +385,8 @@ TEST_F(ServiceHooksTest, TestListenerHook)
 
     listenerHookReg2.Unregister();
     listenerHookReg1.Unregister();
+
+    // serviceListenerHook1->ordering.clear();
```
jeffdiclemente, line 389: this line should be commented to explain why its commented out, deleted or uncommented.

jeffdiclemente (changes requested):
Clearing the `count` in the destructor is one way to solve the problem of not being able to run `gtest_repeat`. Another way would be to make the `ordering` member variables no longer `static`. I would prefer the latter solution as there doesn't appear to be a need to retain the ordering data across multiple test point runs.

framework/test/gtest/ServiceHooksTest.cpp
tcormackMW: That was an oversight, I was testing that as a solution and forgot to remove it.

tcormackMW: @jeffdiclemente In terms of the static-ness of the variable, I think the reason is that so each instance of the listener accesses the same variable. If we change it to be non static we'd have to then somehow link the instances to the same object. We could use a shared_ptr to an object so they both look at the same thing, but this would require a bit of changing constructors etc to facilitate. I think that is a good solution as it would remove the static, but I'm not sure if it fits into the goals you have.

framework/test/gtest/MultipleListenersTest.cpp
tcormackMW, line 1: @jeffdiclemente I am not totally sure how to resolve this instance of globalness. In the context of ordering, it was clear -allow the classes to be instantiated with a shared object somehow. In this case, we have callbacks being made from the framework to functions defined in the namespace. These functions are all accessing the same variable `count`. That namespace is not being reset between repetitions and so no destructors are being called. 

Given that each of these callbacks need to be able to access the same variable, which must be defined in the namespace (unless we could potentially move those class definitions to within functions, though this seems flawed) I am unsure how to create a variable which will be bound to the lifetime of the test case rather than the suite. 

I am not sure if you have suggestions on this issue.

framework/test/gtest/ServiceHooksTest.cpp
```diff
@@ -190,14 +198,12 @@ namespace
             ASSERT_EQ(context, bundleCtx);
 
             references.clear();
-            ordering.push_back(id);
+            ordering->push_back(id);
         }
 
-        static std::vector<int> ordering;
+        std::shared_ptr<std::vector<int>> ordering;
```
jeffdiclemente, line 204: for better encapsulation, make this `private` and add an accessor function.

jeffdiclemente (changes requested):

framework/test/gtest/ServiceHooksTest.cpp
```diff
@@ -237,20 +261,22 @@ namespace
...
             {
                 listenerInfos.erase(*iter);
-                ordering.push_back(id * 10);
+                ordering_->push_back(id * 10);
             }
             lastRemoved = listeners.back();
         }
 
-        static std::vector<int> ordering;
+        std::vector<int>
```
jeffdiclemente, line 269: return a `std::shared_ptr<std::vector<int>>>` instead. this breaks the purpose of having a `std::shared_ptr<std::vector<int>>>`

framework/test/gtest/ServiceHooksTest.cpp
```diff
@@ -190,14 +204,16 @@ namespace
             ASSERT_EQ(context, bundleCtx);
 
             references.clear();
-            ordering.push_back(id);
+            ordering_->push_back(id);
         }
 
-        static std::vector<int> ordering;
+        std::vector<int>
```
jeffdiclemente, line 210: return a `std::shared_ptr<std::vector<int>>>` instead. this breaks the purpose of having a `std::shared_ptr<std::vector<int>>>`

jeffdiclemente (changes requested):

framework/test/gtest/ServiceHooksTest.cpp
tcormackMW: I'm happy to make that change. In this case, the ordering is a `std::shared_ptr<std::vector<int>>` owned by the hooks. In the case of `GetOrdering()`, we are just comparing the value of the `std::vector` to some expected ordering. It has to be shared in order to be owned and edited by multiple hooks. 

I am curious why the calling code needs, or wants, to have a `shared_ptr<std::vector<int>>` to this ordering rather than just owning a `std::shared_ptr<ServiceHooksObject>` to the object that owns `shared_ptr<vector<int>> ordering_`. As long as the calling code has a `std::shared_ptr<ServiceHooksObject>`, that object and therefore the `ordering_` will not die

jeffdiclemente: I'm not sure what you mean by `calling code` in this case. Taking a step back, these tests are verifying the order of events delivered to multiple `ServiceEventListenerHook` and `ServiceFindHook` objects. Since `TestServiceFindHook` and `TestServiceEventListenerHook` derive from abstract base classes, could you use [gmock to check for a specific sequence of method calls](https://github.com/google/googletest/blob/main/docs/gmock_for_dummies.md#ordered-vs-unordered-calls-orderedcalls)?

tcormackMW: @jeffdiclemente While I agree that you could make mocks, there is also a lot of other things going on in the `Forward()`, `Find()` etc functions other than just verifying order of calls -- see [here](https://github.com/CppMicroServices/CppMicroServices/blob/e430df5df499098d56cf2aab3a1cee2d8be9dc7b/framework/test/gtest/ServiceHooksTest.cpp#L89C9-L90C10) for example. 

I read through the page on mocks and looked around and I think you are not allowed to program behavior for these mock functions. Given that, we would have to rearrange these tests significantly in some way (though I am not sure how off the top of my head) to verify what is currently verified.

Additionally we could verify that calls are made to `Find()`, `Added()`, `Removed()` etc. in the correct order, but the variable currently being placed into the vector is not an input but rather a private variable of the mock class. Given this, you don't have access to the `id` from the signature of the function call to use the syntax `EXPECT_CALL(mock, Removed(1));` and so the order would purely be which function is called, not which function with which input. 

Let me know if I am missing something here because ideally I agree, mocks would be the way forward.

jeffdiclemente: I agree, there is a lot of different behaviors being tested in the tests. I suggest pulling out the ordering related tests into separate test points and using mocks for those tests.

you can add custom behavior to your mocks. There are a few methods, based on specific needs detailed [here](http://google.github.io/googletest/gmock_cook_book.html#DelegatingToFake). Also, take a look at [Mocks.hpp](https://github.com/CppMicroServices/CppMicroServices/blob/e430df5df499098d56cf2aab3a1cee2d8be9dc7b/compendium/DeclarativeServices/test/gtest/Mocks.hpp#L330) for some examples of doing this.

As it relates to the variable being placed into the `ordering` vector, from what I can see the variable is used only as a means to check for the correct ordering. If using mocks allows to check for correct ordering of callback calls, is the variable placed in the `ordering` vector even needed? Is the `ordering` vector itself even needed?

tcormackMW: Darn, I missed this implementation ability in my looking around. Yes this seems like a good idea. The ordering variable is definitely not needed as long as we can verify that the correct instance of these functions is being called (which again is dependent on member variables, ie `id`, rather than arguments to the function call). I will look into a way to verify this behavior and go from there but separating out these tests feels like a good idea.

tcormackMW: You're right this should work no problem. I'll work on this before our meeting later this morning

tmp.txt
```diff
@@ -0,0 +1,95 @@
+Note: Google Test filter = ServiceHooksTest.*
```
jeffdiclemente, line 1: was this file unintentionally committed?

framework/test/gtest/ServiceHooksTest.cpp
```diff
@@ -305,6 +292,54 @@ namespace
 
 } // end unnamed namespace
 
+class MockServiceEventListenerHook : public ServiceEventListenerHook
+{
+  public:
+    MockServiceEventListenerHook(int id, BundleContext const& context)
+        : fake_(TestServiceEventListenerHook(id, context))
```
jeffdiclemente, line 299: I don't think this is correct. The `id` and `fake_` members are not needed for the mock and for testing the ordering of callbacks.

framework/test/gtest/ServiceHooksTest.cpp
```diff
@@ -305,6 +292,54 @@ namespace
...
+                (override));
+
+  private:
+    TestServiceFindHook fake_;
+};
+
+class MockServiceListenerHook : public ServiceListenerHook
+{
+  public:
+    MockServiceListenerHook(BundleContext const& context) : fake_(TestServiceListenerHook(context)) {}
```
jeffdiclemente, line 334: I don't think this is correct. The `fake_` member is not needed for the mock and for testing the ordering of callbacks.

jeffdiclemente (changes requested):

framework/test/gtest/ServiceHooksTest.cpp
tcormackMW: I guess you're right. We no longer need the `fake_` variable so we don't need to worry about id. I was thinking we still needed to have the fake_ variable so we needed the id, because the `TestServiceEventListener` requires the id for more than just ordering so it needed to remain in its constructor and therefore the constructor of the mock as well, but it doesn't need to stay in the mock if we don't have the `fake_`.

framework/test/gtest/ServiceHooksTest.cpp
tcormackMW: See below

framework/test/gtest/ServiceHooksTest.cpp
```diff
@@ -602,4 +567,146 @@ TEST_F(ServiceHooksTest, TestListenerHookFailure)
...
+    TestServiceListener serviceListener2;
+    context.AddServiceListener(&serviceListener1, &TestServiceListener::ServiceChanged);
+    context.AddServiceListener(&serviceListener2,
+                               &TestServiceListener::ServiceChanged,
+                               LDAPProp(Constants::OBJECTCLASS) == "bla");
+
+    auto serviceEventListenerHook1 = std::make_shared<MockServiceEventListenerHook>();
+    auto serviceEventListenerHook2 = std::make_shared<MockServiceEventListenerHook>();
+
+#ifdef US_BUILD_SHARED_LIBS
```
jeffdiclemente, line 678: is the `#ifdef` needed here? its not used in the original test point.

framework/test/gtest/ServiceHooksTest.cpp
```diff
@@ -602,4 +567,146 @@ TEST_F(ServiceHooksTest, TestListenerHookFailure)
...
+    listenerHookReg2.Unregister();
+    listenerHookReg1.Unregister();
+}
+
+TEST_F(ServiceHooksTest, TestFindHookCallbackOrdering)
+{
+    auto serviceFindHook1 = std::make_shared<MockServiceFindHook>();
+    auto serviceFindHook2 = std::make_shared<MockServiceFindHook>();
+
+#ifdef US_BUILD_SHARED_LIBS
```
jeffdiclemente, line 630: is the `#ifdef` needed here? its not used in the original test point.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #876, opened 2023-06-29
Fixes #718

### LINKED ISSUE #718: BundleContextTest.NoSegfaultWithServiceFactory is sporadically failing on all platforms
The `BundleContextTest.NoSegfaultWithServiceFactory` test is sporadically segfaulting on all platforms. It appears that the check-then-act fixes were not sufficient for preventing the segfault.

tcormackMW:
Removed assertion because, depending on interleaving, the framework may or may not throw. Either is fine and expected. 

Issue with segfaults occurred because if the framework threw, the `.join()` was not reached and on destruction of the spawned thread, it called terminate as a result of no `.join()` or `.detach()` being called. 

Solution: try catch around `GetService()` with no assertion as to behavior.

### DIFF
--- a/framework/test/gtest/BundleContextTest.cpp
+++ b/framework/test/gtest/BundleContextTest.cpp
@@ -183,6 +183,8 @@ TEST(BundleContextTest, NoSegfaultWithServiceFactory)
     // .join() and when the spawned thread is destructed, it will call terminate because it was not joined.
     try
     {
+        // Framework may throw as the registration is no longer valid (depending on thread ordering) but we care that no
+        // thread segfaults
         context.GetService(svcGetServiceThrowsRefs[0]);
     }
     catch (...)
--- a/framework/test/gtest/BundleContextTest.cpp
+++ b/framework/test/gtest/BundleContextTest.cpp
@@ -172,17 +172,15 @@ TEST(BundleContextTest, NoSegfaultWithServiceFactory)
     ASSERT_EQ(svcGetServiceThrowsRefs.size(), 1);
     ASSERT_EQ("1", svcGetServiceThrowsRefs[0].GetProperty(std::string("getservice_exception")).ToString());
 
-    auto fut = std::async(std::launch::async,
-                          [&framework]()
-                          {
-                              framework.Stop();
-                              framework.WaitForStop(std::chrono::milliseconds::zero());
-                          });
-    auto temp = svcGetServiceThrowsRefs[0];
-    US_UNUSED(temp);
+    std::thread thread(
+        [&framework]()
+        {
+            framework.Stop();
+            framework.WaitForStop(std::chrono::milliseconds::zero());
+        });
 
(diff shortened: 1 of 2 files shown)


framework/test/gtest/BundleContextTest.cpp
```diff
@@ -179,8 +179,17 @@ TEST(BundleContextTest, NoSegfaultWithServiceFactory)
...
+    // .join() and when the spawned thread is destructed, it will call terminate because it was not joined.
+    try
+    {
+        // Framework may throw as the registration is no longer valid (depending on thread ordering) but we care that no
+        // thread segfaults
+        context.GetService(svcGetServiceThrowsRefs[0]);
+    }
+    catch (...)
+    {
+    }
```
jeffdiclemente, line 192: removing the `ASSERT_EQ` negates the purpose of the test (side note: I think an `ASSERT_NO_THROW` should be added to this test). This tests that an exception is never thrown. Look into the reason why an exception is thrown when it is not expected.

jeffdiclemente (changes requested):

framework/test/gtest/BundleContextTest.cpp
tcormackMW: @jeffdiclemente When I spoke with Alex yesterday, we decided that because the framework can be stopped before the call to GetService (depending on scheduling), it is entirely possible that GetService will cause an error to be thrown. In this case the error is thrown from this [line ](https://github.com/CppMicroServices/CppMicroServices/blob/47628ddc8840d306ddbdd1b3c8d1e211068de646/framework/src/bundle/BundleContext.cpp#L323C1-L324C1)

The check, `!reference` brings us to this boolean operator [override ](https://github.com/CppMicroServices/CppMicroServices/blob/47628ddc8840d306ddbdd1b3c8d1e211068de646/framework/src/service/ServiceReferenceBase.cpp#L57C25-L57C25) which calls `GetBundle` and eventually enters this [line](https://github.com/CppMicroServices/CppMicroServices/blob/47628ddc8840d306ddbdd1b3c8d1e211068de646/framework/src/service/ServiceReferenceBase.cpp#L97C1-L97C13).

The reason that `reg` in `GetBundle` can be invalid is that if the framework has stopped, the `registrationBasePrivate` that `reg` points to has died. This is potentially different behavior than previously because in the new shared_ptr implementation, a `ServiceReferenceBasePrivate` no longer owns a `ServiceRegistrationBasePrivate`. 

Long story short, depending on threading, the reference may or may not have a valid `ServiceRegistrationBasePrivate` and so calling `GetReference` may or may not fail, but it will not segfault and cause the program to crash. I think Alex and my understanding was that this test ensures that we do not segfault, but similar to the [test](https://github.com/CppMicroServices/CppMicroServices/blob/47628ddc8840d306ddbdd1b3c8d1e211068de646/framework/test/gtest/BundleContextTest.cpp#L153) above this one, a throw is not outside of expected behavior but please let me know if this is not the case.

jeffdiclemente (approved):

PR #878, opened 2023-07-11
Fix #846

### LINKED ISSUE #710: listenerToken data race in ServiceTrackerPrivate
The ServiceTrackerPrivate holds a listener token. In ServiceTracker::Close(), the listener token is passed through a std::move() outside of the synchronized region for the ServiceTrackerPrivate, creating a data race when Open() and Close() calls are made concurrently.

This race is warned by TSan, but is currently suppressed.

Moving the RemoveListener call in ServiceTracker::Close() will likely resolve the race, but the possibility of re-entrancy should be examined before implementing this solution.

**To Reproduce:**
- Remove ServiceTracker lines from tsan_suppressions.txt
- Run usFrameworkTests (specifically ServiceTrackerTest.ServiceTrackerConcurrentOpenClose) in TSan

### LINKED ISSUE #846: multiple thread sanitizer issues in usFrameworkTests
These errors are from the test log: https://github.com/CppMicroServices/CppMicroServices/actions/runs/4872584091/jobs/8690974464?pr=845

These issues are reproducible by building the CppMicroServices code base with `US_ENABLE_TSAN` enabled on Linux and running usFrameworkTests repeatedly in a loop.

tcormackMW:
Some changes to code and suppressions for tsan as a result of a TSAN bug with waiting on condition variables


jeffdiclemente (changes requested):
I would prefer these changes be split out into 3 separate PRs instead for easier reviews:
1. The ServiceTracker.tpp code format change (its hard to spot the functional change hidden amongst all the formatting changes).
2. Fix for #846 
3. Fix for #710

framework/test/bundles/libC1/TestBundleC1.cpp
```diff
@@ -138,6 +132,8 @@ namespace cppmicroservices
                         count--;
                     }
                 }
+                // stop spin
```
jeffdiclemente, line 135: its more helpful to other developers to example the "why" in a comment here instead of describing the "what". I recommend adding a short, succinct note about why a `sleep_for` in a loop is being used.

tsan_suppressions.txt
```diff
@@ -2,5 +2,33 @@
 race:cppmicroservices::ServiceListeners::RemoveServiceListener
 race:cppmicroservices::ServiceTracker*
 
+# BundleRegistry races and double lock false-positives
```
jeffdiclemente, line 5: a link to the github issue, https://github.com/google/sanitizers/issues/1259, would be helpful here.

jeffdiclemente (changes requested):

framework/test/bundles/libC1/TestBundleC1.cpp
```diff
@@ -132,7 +136,8 @@ namespace cppmicroservices
                         count--;
                     }
                 }
-                // stop spin
+                // Rather than a condition variable, we are looping until regs is non-empty.
+                // This sleep stops the code from looping too quickly and causing slow downs.
```
jeffdiclemente, line 140: Noting that we are choosing to sleep (which is more inefficient) instead of using a `condition_variable` because of a false positive coming from tsan when using `wait_for` would be useful.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #883, opened 2023-07-12
Fix #710 ServiceTracker TSAN warnings

### LINKED ISSUE #710: listenerToken data race in ServiceTrackerPrivate

tcormackMW:
Fixing TSAN warnings in `ServiceTracker.hpp`. Some were valid, specifically the race when accessing the variable `d->listenerToken`, but others were TSAN bugs related to waiting on condition variables, and so those were silenced


jeffdiclemente (approved):

framework/include/cppmicroservices/detail/ServiceTracker.hpp
```diff
@@ -139,8 +139,16 @@ void ServiceTracker<S, T>::Open()
 template<class S, class T>
 void ServiceTracker<S, T>::Close()
 {
+  // Copy required to prevent data race in call to context.RemoveListener()
+  ListenerToken tokenCpy;
+  {
+    auto l = d->Lock();
+    US_UNUSED(l);
+    std::swap(d->listenerToken, tokenCpy);
```
carneyweb, line 147: The comment and this line of code don't match up. std::swap does exactly that, swaps the state of the two objects, effectively doing this:

    auto tmp = a;
    a = b;
    b = tmp;

So the state of d->listenerToken above is reset to the default state for the ListenerToken class.

carneyweb (changes requested):

carneyweb (approved):
Fix it, then ship it.

framework/include/cppmicroservices/detail/ServiceTracker.hpp
jeffdiclemente: good catch. yes, the comment is not telling the truth

framework/include/cppmicroservices/detail/ServiceTracker.hpp
```diff
@@ -151,9 +151,18 @@ namespace cppmicroservices
     void
     ServiceTracker<S, T>::Close()
     {
+
+        /* Swap default ListenerToken with d-listenerToken required to prevent data race in call to
```
jeffdiclemente, line 155: nit-pick: typo in `d-listenerToken`

The comment is a little confusing, specifically the use of the word `default`. Did you mean `default constructored`?  Also, this comment doesn't explain the "why", which is more imporatnt.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #900, opened 2023-08-10
701 - Fixes #852: adds a [[nodiscard]] to BundleContext::RegisterService (#863)

### LINKED ISSUE #852: Add [[nodiscard]] tag to `BundleContext::RegisterService`

insi-eb:
Cherry-picked #863 / 6ca5c26f849d7e4de8c5c263fbf1e8ed9bbaa830 with adaptations for C++14:
* use [[nodiscard]] only when __has_cpp_attribute is available and __has_cpp_attribute(nodiscard) is valid. Alternatively use [[gnu::warn_unused_result]] if nodiscard is not available, but gnu::warn_unused_result is. AFAICS __has_cpp_attribute was introduced with gcc-5 and is at least available since clang-9, so it should work as intended on most recent compilers and not break the build on ones where neither nodiscard nor gnu::warn_unused_result (that attribute also seems to be available in clang according to the clang documentation) is available.
* revert default C++ standard in doc/src/CMakeLists.txt back to C++14

### DIFF
--- a/compendium/DeclarativeServices/test/gtest/TestServiceComponentRuntimeImpl.cpp
+++ b/compendium/DeclarativeServices/test/gtest/TestServiceComponentRuntimeImpl.cpp
@@ -317,7 +317,8 @@ namespace cppmicroservices
             auto iMap = std::make_shared<InterfaceMap>();
             auto obj = std::make_shared<double>();
             iMap->insert(std::make_pair("double", std::static_pointer_cast<void>(obj)));
-            fc.RegisterService(iMap);
+            auto reg = fc.RegisterService(iMap);
+            US_UNUSED(reg);
 
             auto sRef = fc.GetServiceReference("double");
             EXPECT_TRUE(static_cast<bool>(sRef));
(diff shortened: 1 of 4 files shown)


insi-eb:
hm ... some clang versions seem to complain about seeing nodiscard even though they report it as available:

30: /home/runner/work/CppMicroServices/CppMicroServices/build/install_test/usr/local/include/cppmicroservices3/cppmicroservices/BundleContext.h:310:11: error: use of the 'nodiscard' attribute is a C++17 extension [-Werror,-Wc++17-extensions]
30:         [[nodiscard]]
30:           ^

insi-eb:
> as newer gcc and clang versions seem to provide both nodiscard and gnu::warn_unused_results, I now changed this to only use nodiscard when using C++17 or newer and fall back to gnu::warn_unused_results otherwise - I don't see how else to do that without making this even more complicated.

I could move this whole #ifdef chain to its own header and there then try to find something that works as intended for all major compilers (adding `_Check_return_` for MSVC).

jeffdiclemente (approved):

Issue #901, opened 2023-08-10
Bundle Validation called for every service activation


ShivamNegi:
In the current implementation of Bundle Validation the code for validation is being called from 

https://github.com/CppMicroServices/CppMicroServices/blob/f9fd6587f73ea270014e482d6560c3c70c02d4e7/compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp#L489

Which calls Bundle Validation for each Activation of the bundle service, i.e. if there are four services in a bundle, Bundle Validation will be called four times for the bundle. 

The desired  working is that Bundle Validation is only called for the first loading of the Bundle and not at service activation.

jeffdiclemente:
fixed in PR #921

PR #914, opened 2023-09-06
Fix Redundant Bundle Validation checks

ShivamNegi:
Issue #901 

Added a set `bundleValidated` to keep track of the bundles being validated. This will ensure that upon any subsequent calls for the bundle service the validation will not be performed again.

For the case when bundle state changes to `uninstall` the `bundleValidated` markers is not being removed as the bundle is never unloaded from the process memory. So, for any subsequent installs of the bundle need not be validated again.


ShivamNegi:
The [BundleValidationFunctionException](https://github.com/CppMicroServices/CppMicroServices/blob/4fded15af69c1b4ffd53c51d8eb857d4842d08dc/compendium/DeclarativeServices/test/gtest/TestBundleValidation.cpp#L271) test is failing as it expects to perform validation on a bundle which has already been validated and marked in the code as validated in the previous test. The test expects to get an exception when the same bundle is validated by overriding the bundle validation function. But, due to the current optimization the validation is never called a on a bundle which has already been validated.

I think the test case needs to be modified to perform the test on a fresh bundle which has not been validated.

jeffdiclemente:
this is a duplicate of PR #921, closing this one.

PR #915, opened 2023-09-11
Fix #913 ServiceTracker deadlock

### LINKED ISSUE #913: ServiceTracker deadlocks if framework is terminated first
context: using a `cppmicroservices::ServiceTracker` to watch for a service and then relying on either the service becoming resolved or closing the service tracker to unblock the thread.

If the `framework` is stopped while a `ServiceTracker` is calling `WaitForService` with no timeout, the `ServiceTracker` becomes entirely unusable and deadlocks the program. `Close()` can't be called on the tracker any more because accessing the bundle context throws, it will never resolve because the `framework` is gone.

```
Framework framework = FrameworkFactory().NewFramework();
framework.Start();

ServiceTracker<MyInterfaceOne> tracker(framework.GetBundleContext());
tracker.Open();
auto f = std::async(
    [&framework]
    {
        // technically there's a race here, so the async should ensure WaitForService is started prior to
        // terminating the framework.
        std::this_thread::sleep_for(std::chrono::seconds(1));

        framework.Stop();
        framework.WaitForStop(std::chrono::milliseconds::zero());
    });

auto const service = tracker.WaitForService();
// unreachable code since it's impossible to unblock WaitForService now
```

tcormackMW:
This fixes the deadlock in #913 caused by a waitForService call deadlocking as a result of the bundle being deleted while the tracker is waiting.

As bundles do not have a way to access the tracker on destruction to wake it up, this implementation utilizes a time limit for the wait condition to ensure that the bundle is periodically checked.

### DIFF
--- a/.github/workflows/performance_windows.yml
+++ b/.github/workflows/performance_windows.yml
@@ -3,8 +3,6 @@ name: PerformanceWindows
 on:
   push:
     branches: [ development ]
-  pull_request:
-    branches: [ development ]
   workflow_dispatch:
 
 permissions:
@@ -56,17 +54,17 @@ jobs:
     - name: Run benchmark and store result in json format
       run: ${{ env.BUILD_DIR }}\bin\Release\usFrameworkBenchTests.exe  --benchmark_format=json | tee ${{ env.BUILD_DIR }}\bin\Release\benchmark_result.json
 
-    # - name: Deploy results
-    #   uses: benchmark-action/github-action-benchmark@v1.16.1
-    #   with:
-    #     name: CppMicroServices Benchmark
-    #     tool: 'googlecpp'
-    #     output-file-path: ${{ env.BUILD_DIR }}\bin\Release\benchmark_result.json
-    #     github-token: ${{ secrets.GITHUB_TOKEN }}
-    #     auto-push: true
-    #     # Show alert with commit comment on detecting possible performance regression
-    #     alert-threshold: '20%'
-    #     comment-on-alert: true
-    #     fail-on-alert: false
+    - name: Deploy results
+      uses: benchmark-action/github-action-benchmark@v1.16.1
+      with:
+        name: CppMicroServices Benchmark
+        tool: 'googlecpp'
+        output-file-path: ${{ env.BUILD_DIR }}\bin\Release\benchmark_result.json
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        auto-push: true
+        # Show alert with commit comment on detecting possible performance regression
+        alert-threshold: '20%'
+        comment-on-alert: true
+        fail-on-alert: false
         
         
--- a/.github/workflows/performance_windows.yml
+++ b/.github/workflows/performance_windows.yml
@@ -3,6 +3,8 @@ name: PerformanceWindows
 on:
   push:
     branches: [ development ]
+  pull_request:
+    branches: [ development ]
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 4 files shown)


.github/workflows/performance_windows.yml
```diff
@@ -3,6 +3,8 @@ name: PerformanceWindows
 on:
   push:
     branches: [ development ]
+  pull_request:
```
tcormackMW, line 6: this is done just for performance testing pre-merge...; this will be reverted for the actual merge

framework/include/cppmicroservices/detail/ServiceTracker.hpp
```diff
@@ -263,7 +263,42 @@ namespace cppmicroservices
                 auto l = t->Lock();
                 if (t->Size_unlocked() == 0)
                 {
-                    t->WaitFor(l, rel_time, [&t] { return t->Size_unlocked() > 0 || t->closed; });
+                    BundleContext a = d->context;
+                    if (!a)
+                    {
+                        throw std::invalid_argument("The bundle context cannot be null.");
```
jeffdiclemente, line 269: A few comments on this line:
1. codecov is indicating that there is no code coverage for this line. I recommend adding a test that exercises this code path.
2. This should be a [logic_error](https://en.cppreference.com/w/cpp/error/logic_error) instead of `invalid_argument`. This failure happens when some other thread prematurely stops the bundle, which I would argue is a logic error.
3. The public doxygen API documentation for `WaitForService` needs to be updated to tell users about this exception.

jeffdiclemente (changes requested):

framework/include/cppmicroservices/detail/ServiceTracker.hpp
tcormackMW: 1. I agree, I have added a test... That test will be changed in future commits when we either get rid of the `GetLogSink` or change that it throws because right now we are catching a runtime error from that instead of the logic_error that we would like it to be. 
2. I agree
3. ServiceTracker.h has been updated

framework/include/cppmicroservices/detail/ServiceTracker.hpp
```diff
@@ -263,7 +263,42 @@ namespace cppmicroservices
...
+                    }
+
+                    if (rel_time == std::chrono::milliseconds::zero())
+                    {
+                        // need to cycle on definite end time to check for invalid bundle
+                        while (true)
```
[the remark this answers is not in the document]
tcormackMW: I agree. Has been resolved

framework/include/cppmicroservices/detail/ServiceTracker.hpp
```diff
@@ -252,7 +252,34 @@ namespace cppmicroservices
...
-                    t->WaitFor(l, rel_time, [&t] { return t->Size_unlocked() > 0 || t->closed; });
+                    BundleContext a = d->context;
+                    if (!a)
+                    {
+                        throw std::logic_error("The bundle context cannot be null.");
+                    }
+
+                    if (rel_time == std::chrono::milliseconds::zero())
+                    {
+                        // need to cycle on definite end time to check for invalid bundle
```
jeffdiclemente, line 263: it would be helpful to explain why this is checking for an invalid bundle.

jeffdiclemente (changes requested):
codecov is showing that some code isn't covered by the test, specifically the failure case. I recommend adding tests to cover the failure cases.

tcormackMW:
@jeffdiclemente For some reason CodeCov is not identifying that [l258](https://github.com/CppMicroServices/CppMicroServices/blob/87748fe2afab4af81564f3163ab16d593938e0af/framework/include/cppmicroservices/detail/ServiceTracker.hpp#L258C1-L259C1) is covered by test case [ServiceTrackerTests.TestServiceTrackerInvalidBundle](https://github.com/CppMicroServices/CppMicroServices/blob/87748fe2afab4af81564f3163ab16d593938e0af/framework/test/gtest/ServiceTrackerTest.cpp#L712C1-L713C1). To verify, I just walked through code and it does in fact hit that line. I think this could have to do with the fact that this is a new test and so it is not picking it up? Not sure, but that test does hit that line.

jeffdiclemente (approved):

PR #921, opened 2023-09-12
Fix Redundant Bundle Validation checks

ShivamNegi:
Making independent TestBundles for all the test cases in `TestBundleValidation.cpp`


compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
```diff
@@ -488,18 +489,24 @@ namespace cppmicroservices
         InstanceContextPair
         ComponentConfigurationImpl::CreateAndActivateComponentInstanceHelper(cppmicroservices::Bundle const& bundle)
         {
+            static Guarded<std::unordered_set<std::string>> bundleValidated;
```
jeffdiclemente, line 492: this is somewhat redundant to `static Guarded<std::map<std::string, void*>> bundleBinaries;` used in `GetComponentCreatorDeletors`. Do we need both maps or can this be consolidated?

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
```diff
@@ -488,18 +489,24 @@ namespace cppmicroservices
...
+            try {
+                if (bundleValidated.lock()->count(bundleLoc) == 0u) {
+                    if (!func.Empty()
+                        && !any_cast<std::function<bool(cppmicroservices::Bundle const&)>>(func)(this->bundle))
+                    {
+                        std::string errMsg("Bundle at location ");
+                        errMsg += bundleLoc;
+                        errMsg += " failed bundle validation.";
+                        throw SecurityException{ std::move(errMsg), this->bundle };
+                    }
```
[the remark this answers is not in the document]
ShivamNegi: Yes, this makes a lot of sense. Design wise it will be close to the actual bundle loading and the same data structure can be used for validation it.

compendium/DeclarativeServices/test/gtest/TestBundleValidation.cpp
```diff
@@ -358,11 +360,53 @@ TEST(TestBundleValidation, BundleValidationFunctionException)
...
+    auto compDesc = dsRuntimeService->GetComponentDescriptionDTO(*bundleIter, "sample::ServiceComponentBV3_1");
     ASSERT_FALSE(dsRuntimeService->IsComponentEnabled(compDesc));
     ASSERT_TRUE(receivedSecondBundleValidationErrorEvent);
 
     f.GetBundleContext().RemoveListener(std::move(token));
     f.Stop();
     f.WaitForStop(std::chrono::milliseconds::zero());
 }
+
+TEST(TestBundleValidation, BundleValidationMultipleSrc)
```
jeffdiclemente, line 372: What is this test testing that isn't already tested in pre-existing test cases? I would expect to see uninstalling of the bundle and re-installing and starting to make sure that bundle validation isn't happening again.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/gtest/TestBundleValidation.cpp
ShivamNegi: This test installs a bundle with two services. Making sure that a bundle with two services is not validated twice.

I see your point I can add another test something like `BundleValidationReinstall` which tests the validation logic on re-installation.

ShivamNegi:
Added a new testcase, `BundleValidationReinstall`. This test case works by reinstalling the bundle `TestBundleDSTBV5` to check if the reinstall skips validation. I have changed the `validationFunc` to fail on reinstalling the same bundle so when the test tries to reinstall the test expects `no error`. Its like a double negative which ensures that the validation is not happening.

I have taken this approach as the framework configurations cannot be modified once created to check for validation skips like in the `BundleValidationFailure` test case.

ShivamNegi:
Linking PR to Issue #901 

Brief on what it does as it is:
Provide a fix for redundant Bundle Validation by keeping track of validated bundles. Updating the test procedures to work with the updated functionality.

carneyweb (approved):
Ship it!

PR #922, opened 2023-09-13
Fix #920 serviceTracker segfault on concurrent `tracker.Close()` and `framework.Stop()`

### LINKED ISSUE #920: Sporadic failure in ServiceTracker on framework shutdown
There is a sporadic issue where it seems like `GetLogSink()` returns `nullptr` when the framework is being torn down and then is dereferenced. The error originates in `ServiceTracker.hpp` (line 190):

```
  DIAG_LOG(*d->context.GetLogSink()) << "ServiceTracker<S,TTT>::close[cached cleared]:" << d->filter;
```

Either the context is not valid at this point (it is being checked before for validity) or `GetLogSink()` is referencing a deleted object. It seems to be a race condition while closing the framework.

tcormackMW:


carneyweb (approved):

tcormackMW:
@jeffdiclemente Why is `d` valid when `CheckValid()` fails?

`BundleContext.d` is set to nullptr in the destruction of a framework inside the thread spawned by `Stop()`. This occurs [here](https://github.com/CppMicroServices/CppMicroServices/blob/4fded15af69c1b4ffd53c51d8eb857d4842d08dc/framework/src/util/FrameworkPrivate.cpp#L94).

But, as we found out, Bundles do not have access to their trackers. i.e. the tracker's `BundleContext` is not invalidated when the framework's `BundleContext` is swapped out of the framework. And the `ServiceTracker` holds a `shared_ptr<ServiceTrackerPrivate> which has a `BundleContext` which has a `std::shared_ptr<BundleContextPrivate>`, thereby keeping that object alive even if invalid. 

So even though the base object, the `BundleContextPrivate` is invalid, it is still alive. And so I think if `ServiceTracker`s are the ones that are calling `GetLogSink()`, I don't see a scenario where the `d` (a std::shared_ptr<BundleContextPrivate>` will ever be invalid in that function call. 

We are deleting the `GetLogSink()` anyway, but now I understand why that other if statement is not catching anything.

framework/src/bundle/BundleResourceContainer.cpp
```diff
@@ -223,8 +223,6 @@ namespace cppmicroservices
         }
         catch (std::exception const& ex)
```
jeffdiclemente, line 224: `ex` can be removed since its not used. I'm surprised that the linux/macOs builds didn't fail with an "unused variable" warning.

framework/src/bundle/BundleResource.cpp
```diff
@@ -289,12 +289,6 @@ namespace cppmicroservices
         }
 
         auto data = d->archive->GetResourceContainer()->GetData(d->stat.index);
-        if (!data)
-        {
-            auto sink = GetBundleContext().GetLogSink();
-            DIAG_LOG(*sink) << "Error uncompressing resource data for " << this->GetResourcePath() << " from "
-                            << d->archive->GetBundleLocation();
-        }
```
jeffdiclemente, line 292: remove `return data;` and replace `auto data = ` with `return`.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #929, opened 2023-09-18
Fix #926 Custom CppMicroService Boost Namespace

### LINKED ISSUE #926: Allow for custom Boost library usage
Accept a different boost library when building from source to allow multiple versions of Boost while avoiding boost symbol collisions.

tcormackMW:
For 4 files (outside of boost library) that were changed, symbols were verified as below

`nm example.cpp.o | c++filt`

The 4 files were: 

```
compendium/DeclarativeServices/src/CMakeFiles/DeclarativeServicesObjs.dir/SCRAsyncWorkService.cpp.o
compendium/DeclarativeServices/test/gtest/CMakeFiles/usDeclarativeServicesTests.dir/TestAsyncWorkService.cpp.o
compendium/ConfigurationAdmin/src/CMakeFiles/ConfigurationAdminObjs.dir/CMAsyncWorkService.cpp.o
compendium/ConfigurationAdmin/test/CMakeFiles/usConfigurationAdminTests.dir/TestAsyncWorkService.cpp.o
```

The output of this command on these files showed only symbols within the `cppmsboost` namespace

Additionally all tests passed locally


jeffdiclemente:
@tcormackMW thanks for updating all the boost files! Can you please verify that each of the 3 shared libraries (i.e. the core framework, DS and ConfigAdmin) don't contain symbols using the `boost` namespace. thanks.

tcormackMW:
@jeffdiclemente Additionally verified 

```
./lib/libCppMicroServicesd.so
./lib/libDeclarativeServicesd.so
./lib/libConfigurationAdmind.so
```

using command `nm -D ./lib/libConfigurationAdmind.so` and found that all references to `boost` were using the `cppmsboost` namespace

PR #939, opened 2023-10-03
Fix #938 Boost namespace Versioning and remove absl dependency

### LINKED ISSUE #938: Modify cppmicroservices Cmake build infrastructure to allow versioning of included libraries

tcormackMW:
This reverts the change from `namespace boost` to `namespace cppmsboost` as we have now changed Cmake infrastructure to allow for custom specification of Boost Library versioning


jeffdiclemente:
> This reverts the change from `namespace boost` to `namespace cppmsboost` as we have now changed Cmake infrastructure to allow for custom specification of Boost Library versioning

@tcormackMW  I would prefer not to revert the boost namespace change. This allows default building of cppmicroservices (i.e. no custom boost being used) to guarantee there are no namespace collisions with boost and other libraries using boost.

tcormackMW:
@jeffdiclemente Performance testing done locally. From what is clear through the noise, the runs that show high regression change every run -- i.e. inconsistent performance losses. The tests that go much faster are somewhat consistently faster. 

I think that these changes should not introduce regressions and I am guessing we will gain back some speed from these changes.

jeffdiclemente (changes requested):
With absl being removed, its usage should be removed from framework\CMakeLists.txt

jeffdiclemente (approved):

PR #945, opened 2023-10-16
CMake Updates for Boost Versioning Fix

tcormackMW:
This creates updates for the Boost Versioning fix that was merged recently in PR #939 
That PR did not include changes to CMake files in the subfolders (ConfigurationAdmin and DeclarativeServices) that were necessary to support the feature change

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,6 +171,7 @@ include(usFunctionGenerateBundleInit)
 include(usMacroCreateBundle)
 include(usFunctionCreateTestBundle)
 include(usFunctionCreateDSTestBundle)
+include(usFunctionBoostPath)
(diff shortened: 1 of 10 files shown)


jeffdiclemente (changes requested):
all of these files have the same code pattern:
```
  set(_boost_library)
  if (US_USE_SYSTEM_BOOST)
      set(_boost_library ${BOOST_INCLUDEDIR})
  else()
      set(_boost_library ${CppMicroServices_SOURCE_DIR}/third_party/boost/include)
  endif()
```

can this pattern be placed in some re-usable `.cmake` file for any CMakeLists.txt file?

jeffdiclemente (approved):

Issue #491, opened 2020-08-03
GetBundleContext doesn't work for Declarative Services bundles


jeffdiclemente:
[GetBundleContext](http://docs.cppmicroservices.org/en/stable/framework/doc/api/util/GetBundleContext.html) doesn't work if the bundle is a Declarative Services bundle. This is because this function relies on the "set bundle context" extern C function which is only generated when the CPPMICROSERVICES_INITIALIZE_BUNDLE macro is specified (which isn't typically used for Declarative Services bundles).

PR #947, opened 2023-10-19
Fix GetBundleContext when using Declarative Services

bwestcot:
Resolves issue #491 
Described in https://github.com/CppMicroServices/rfcs/pull/20

Changes included in this MR:
- Three new tests verifying bundle context consistency in DS
- Three new bundles corresponding to these tests
- `CPPMICROSERVICES_INITIALIZE_BUNDLE` is now added by the DS code generation tool
- CMake changes to prevent duplication of `CPPMICROSERVICES_INITIALIZE_BUNDLE`
- `CPPMICROSERVICES_INITIALIZE_BUNDLE` now uses `BundleContext`s instead of `BundleContextPrivate`s and function signature was moved, enabling its use outside of the CF
- Moved `BundleUtils` from CF to `util/`, simplified it, and renamed to the more specific `BundleHandles`
- Updated CF bundle loader to use `BundleHandles`
- Added bundle context initialization to the DS loader, using the now-public `BundleHandles` functions

Qualification done:
- Sterile build on Windows + Visual Studio produces no errors (other platforms not tested)
- All existing tests pass (no apparent regressions)
- One new test passed both before and after changes (verifying problem understanding)
- Other two new tests used to fail and now pass (verifying solution)


jeffdiclemente (changes requested):
I think this design can be simpler since the `BundleContext` object is a ref-counted type.
Consider that there are 3 use cases for how a `BundleContext` object is set for the bundle:
1. the bundle does not use DS and has an activator
2. the bundle does not have an activator and uses DS (immediate or lazily loaded)
3. the bundle uses DS and has an activator (lazily loading has no relevance because the activator requires immediate library loading by design).

### case #1
nothing special needs to happen; the existing code is enough

### case #2
the existing code in `BundlePrivate.cpp` will not set the bundle context (as it already does). Instead DS will set the bundle context in the same manner as BundlePrivate does (i.e. finding the extern C APIs and calling them, passing in the BundleContext object). DS will also clean up the `BundleContext` in this case since it knows when any bundle is being stopped (see `SCRActivator.cpp`). Because the `BundleContext` object is a ref-counted object, as its copied around, the ref count should be managed correctly and the object destroyed automatically.

### case #3
This case is essentially case #1 with added smarts in DS not to attempt to set the `BundleContext` object (because the core framework already did). DS can try to get the `BundleContext` object via the extern C API and if it succeeds, DS can do nothing and assume that the core framework has already set the `BundleContext` object. The `BundleContext` object  will be cleaned up when the last reference to it goes out of scope.

framework/include/cppmicroservices/BundleInitialization.h
```diff
@@ -57,19 +61,26 @@ namespace cppmicroservices
...
-                                                                                                                     \
-        extern "C" US_ABI_EXPORT void US_SET_CTX_FUNC(US_BUNDLE_NAME)(cppmicroservices::BundleContextPrivate * ctx); \
-        extern "C" US_ABI_EXPORT void US_SET_CTX_FUNC(US_BUNDLE_NAME)(cppmicroservices::BundleContextPrivate * ctx)  \
-        {                                                                                                            \
-            US_CTX_INS(US_BUNDLE_NAME).store(ctx);                                                                   \
+#    define CPPMICROSERVICES_INITIALIZE_BUNDLE                                                               \
+        struct {                                                                                             \
+            std::mutex mut;                                                                                  \
+            cppmicroservices::BundleContext* ctx = nullptr;                                                  \
+        } US_CTX_INS(US_BUNDLE_NAME);                                                                        \
```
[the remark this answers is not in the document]
bwestcot: This is needed so that the two accesses on line 81 plus the access on line 82 all occur in one atomic transaction. Otherwise, for example, the `ctx` could be changed between the `if` and the `delete` on 81, which would result in deleting the wrong object.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -534,28 +534,37 @@ namespace cppmicroservices
...
+                util::GetSymbol(createActivatorHook, libHandle, create_activator_func, &create_activator_err);
 
                 std::string destroy_activator_func = US_STR(US_DESTROY_ACTIVATOR_PREFIX) + symbolicName;
-                BundleUtils::GetSymbol(destroyActivatorHook, libHandle, destroy_activator_func);
+                std::string destroy_activator_err;
+                util::GetSymbol(destroyActivatorHook, libHandle, destroy_activator_func, &destroy_activator_err);
 
                 if (!createActivatorHook)
                 {
+                    coreCtx->logger->Log(logservice::SeverityLevel::LOG_ERROR, create_activator_err);
```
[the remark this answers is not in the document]
bwestcot: These calls were already present in the original code, they were just hidden inside `GetSymbol`. In order to move that to `util`, references to the internal logger had to be removed from it. To maintain existing behavior, I added equivalent logger calls where `GetSymbol` is used.

framework/include/cppmicroservices/BundleInitialization.h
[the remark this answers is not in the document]
jeffdiclemente: That does make sense. I recommend changing the design to remove the need for the mutex (and the need to call `delete`) entirely. `extern "C" US_ABI_EXPORT void US_SET_CTX_FUNC(US_BUNDLE_NAME)` is not called concurrently for the same bundle from within the CppMicroServices code base as far as I know.

bwestcot:
Build errors seem to boil down to:
- Commit https://github.com/CppMicroServices/CppMicroServices/commit/1760c69408f8e697c3b6788d2e6cbf7e570e44a6, which adds several new test bundles. These use DS, but of course don't have the `USES_DS` flag added by this change. Will need to reverse-merge and fix.
- Some string-comparison tests on the results of the DS code generator seem to be failing. These were probably also added in commits that aren't in my fork.
- Incomplete typedef for `BundleContext` in `BundleInitialization.h`, making `delete` operation UB. Should be fixable by adding an include, or by taking advantage of `BudleContext` being reference counted.
- GCC doesn't like the explicit copy constructor for `BundleContext`, but IIRC, I added it because MSVC didn't seem to like not having it. I'm probably doing something weird that's making the compilers confused.
- Some other errors that appear to be indirectly caused by the above through build failure.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -534,28 +534,37 @@ namespace cppmicroservices
...
                     throw std::runtime_error("Bundle #" + util::ToString(id) + " (location=" + location
                                              + ") activator constructor not found");
                 }
                 if (!destroyActivatorHook)
                 {
+                    coreCtx->logger->Log(logservice::SeverityLevel::LOG_ERROR, destroy_activator_err);
```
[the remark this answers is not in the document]
jeffdiclemente: I was mistaken, these log calls are fine to leave in. I was thinking about our legacy logger, `detail::LogSink`

framework/src/bundle/BundlePrivate.cpp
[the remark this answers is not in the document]
jeffdiclemente: I was mistaken, these log calls are fine to leave in. I was thinking about our legacy logger, `detail::LogSink`

framework/include/cppmicroservices/BundleContext.h
```diff
@@ -123,6 +123,11 @@ namespace cppmicroservices
          */
         BundleContext();
 
+        /**
+         * Allocates a new %BundleContext object that is a copy of an existing one.
```
carneyweb, line 127: typo?

framework/include/cppmicroservices/BundleInitialization.h
```diff
@@ -20,22 +20,34 @@
...
-#    include "cppmicroservices/GlobalConfig.h"
-
-#    include <atomic>
+#include <functional>
 
 namespace cppmicroservices
 {
+    class BundleContext;
     class BundleContextPrivate;
+    // THIS IS A TRANSFER OF OWNERSHIP. The pointed object will be deleted when the bundle is
```
carneyweb, line 34: I find this comment to be confusing. This actually does nothing except declare a type which is a std::function, which can be coded to do anything. Whether or not it transfers ownership depends entirely on how a particular std::function of that type is coded.

bwestcot: These essentially act as function signatures for the functions inside `CPPMICROSERVICES_INITIALIZE_BUNDLE`. We can't use actual function signatures, since the implementations are in a runtime-selected bundle and therefore can't be resolved at link time. I will add a brief note to explain this. Since interface-level doc generally goes with the function signature, I decided to put it here.

carneyweb (commented):
Ship it!

framework/include/cppmicroservices/BundleContext.h
```diff
@@ -123,6 +123,11 @@ namespace cppmicroservices
...
+         */
+        BundleContext* NewBundleContext();
```
jeffdiclemente, line 129: I'm concerned about inappropriate use of this method by users. This method has a very specific purpose and it should never be used by cppmicroservices users. Can we guard this somehow or put up some indication to users that they should never call this themselves?

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/BundleLoader.cpp
```diff
@@ -132,10 +134,27 @@ namespace cppmicroservices
...
                                     + " (location=" + bundleLoc + ")");
+
+                    std::string set_bundle_context_func = US_STR(US_SET_PUBLIC_CTX_PREFIX) + fromBundle.GetSymbolicName();
+                    SetBundleContextPublicHook hook;
+                    std::string set_bundle_context_err;
+                    util::GetSymbol(hook, sh.GetHandle(), set_bundle_context_func, set_bundle_context_err);
+                    if (hook)
+                    {
+                        hook(ctx.NewBundleContext());
+                    }
```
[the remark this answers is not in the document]
bwestcot: What do you mean "removing it"? The `hook` variable is the `SetBundleContext` function pointer. These lines are the actual fixing of the bug. The variable was named `hook` before, but if you think it's a bad name I can rename it.

framework/include/cppmicroservices/BundleInitialization.h
```diff
@@ -20,22 +20,34 @@
 
 =============================================================================*/
 
-#ifndef US_BUNDLE_NAME
-#    error Missing US_BUNDLE_NAME preprocessor define
-#endif
```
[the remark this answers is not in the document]
bwestcot: The BundleContextHook type definitions needed to be moved to a common place outside the CF for DS to know how to call SetBundleContext. This header seemed like an appropriate place, but including this header in DS and the CF caused this error to get thrown. I reworked the preprocessor statements to the appropriate parts of the header are always used. I can move the type definitions to a new header file though if you think this check is important.

compendium/DeclarativeServices/src/manager/BundleLoader.cpp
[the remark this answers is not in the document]
jeffdiclemente: I mean don't use the word hook anywhere in the new code. The term "hook" in cppms generally refers to a bit of user-defined non-cppms code that cppms runs. This isn't the case for setting/getting the `BundleContext`.

framework/include/cppmicroservices/BundleContext.h
jeffdiclemente: A clear comment of intent will have to do for now.

compendium/tools/SCRCodeGen/test/ReferenceAutogenFiles.hpp
```diff
@@ -57,11 +57,14 @@ extern "C" US_ABI_EXPORT void DeleteInstance_DSSpellCheck_SpellCheckImpl(scd::Co
 #include <vector>
 #include <cppmicroservices/ServiceInterface.h>
 #include "cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp"
+#include "cppmicroservices/BundleInitialization.h"
 #include "SpellCheckerImpl.hpp"
 
 namespace sc = cppmicroservices::service::component;
 namespace scd = cppmicroservices::service::component::detail;
 
+CPPMICROSERVICES_INITIALIZE_BUNDLE
```
jeffdiclemente, line 66: on further thought, having SCRCodeGen put in the `CPPMICROSERVICS_INITIALIZE_BUNDLE` macro automatically is unnecessary. Bundle authors already have to add `CPPMICROSERVICS_INITIALIZE_BUNDLE` to the bundle's source code and its reasonable to expect bundle authors to continue to do so for DS bundles. 

Reverting the SCRCodeGen change also simplifies the CMake build scripts; now there is no distinction needed on whether a DS bundle is being built and there are no concerns about duplicate symbols coming from multiple uses of `CPPMICROSERVICS_INITIALIZE_BUNDLE`.

jeffdiclemente (changes requested):

compendium/tools/SCRCodeGen/test/ReferenceAutogenFiles.hpp
bwestcot: While it is, strictly speaking, unnecessary to add it to the code generator, it is also unnecessary to make every bundle author add it manually. If the purpose of DS is to eliminate boilerplate, and this is boilerplate, then I think it should be done by DS. It feels like a technical debt that bundle authors still need to write this, and this project seems like a perfect time to address it.

While the volume of changed lines is large, that is due to repetition - the change itself is actually quite small. If you are concerned about the number of changed lines, I can invert the flag, so the default is DS and only a few bundles need to specify that they *don't* use DS. But that feels a bit counterintuitive and doesn't seem to solve any problem.

jeffdiclemente: I agree that `CPPMICROSERVICS_INITIALIZE_BUNDLE` is boilerplate and I like your thinking of eliminating technical debt as you make changes, however the cost of keeping this as boilerplate is low. DS mainly eliminates the need for a user to implement a `BundleActivator` class. This class is the biggest potential source of bugs generated by users and eliminating its use through DS has significant value. On the other hand,  `CPPMICROSERVICS_INITIALIZE_BUNDLE` is a single line of code where the potential bug comes from the user not adding it to their source code. Users who are using CppMicroServices are already adding `CPPMICROSERVICS_INITIALIZE_BUNDLE` and the goal of this change is not to eliminate more boilerplate code.

I am less concerned about the # of changes to the CMakeLists.txt files and more concerned about the fact that developers have to remember to add another parameter to `usFunctionCreateTestBundleWithResources` when creating DS test bundles. If the logic was inverted, the same concern remains; adding another parameter to remember. I would prefer to see another CMake function created specifically for creating DS test bundles instead of adding a parameter onto an existing one.

I see less upside to auto-generating `CPPMICROSERVICS_INITIALIZE_BUNDLE` than not auto-generating it. Consider an existing user of CppMicroServices who already has build infrastructure to build their CppMicroServices modules. Its likely they will have to make changes to their infrastructure to make sure there isn't a re-definition coming from `CPPMICROSERVICS_INITIALIZE_BUNDLE` with this change. Whereas if this doesn't exist, no infrastructure change is necessary.  I am valuing not forcing potential build infrastructure changes upon users over removing a piece of boilerplate code.

compendium/test_bundles/TestBundleDSTOI22/src/Activator.cpp
```diff
@@ -0,0 +1,51 @@
...
+    class Activator : public cppmicroservices::BundleActivator
+    {
+      public:
+        Activator() = default;
+        ~Activator() = default;
+
+        void
+        Start(cppmicroservices::BundleContext context)
+        {
+            activatorProvidedCtx = context;
```
jeffdiclemente, line 42: I'm not sure there's much value in saving the bundle context here to check later. We already guarantee that the `BundleContext` accessed through `ComponentContext` and passed to `ServiceComponent22::Activate(std::shared_ptr<ComponentContext> const& ctx)` is the same.

not storing the `BundleContext` also simplifies the test.

jeffdiclemente (changes requested):

framework/src/CMakeLists.txt
```diff
@@ -1,6 +1,3 @@
-#-----------------------------------------------------------------------------
-# Source files
-#-----------------------------------------------------------------------------
```
jeffdiclemente, line 3: minor; this removal seems unnecessary. can this change (deletion) be undone?

framework/src/bundle/BundleUtils.h
```diff
@@ -20,8 +20,8 @@
 
 =============================================================================*/
 
-#ifndef CPPMICROSERVICES_BUNDLEUTILS_H
-#define CPPMICROSERVICES_BUNDLEUTILS_H
```
jeffdiclemente, line 24: revert the #define name change. the convention used is CPPMICROSERVICES_<HEADER FILENAME>_H

jeffdiclemente (changes requested):
looking great! a couple minor fixes are needed and then this looks good to merge!

compendium/test_bundles/TestBundleDSTOI22/src/Activator.cpp
bwestcot: Not checking this would remove a few lines, but it wouldn't really reduce complexity at all. Considering that, I think it makes sense to test all points of contact with the BundleContext at the same time.

If this is something we are already testing for, perhaps it would make sense to merge the two tests?

jeffdiclemente (approved):

PR #949, opened 2023-10-24
ServiceComponent build tests cmake configuration

tcormackMW:
Update ServiceComponent build tests to use the cmake configuration of the original build

### DIFF
--- a/compendium/ServiceComponent/test/CMakeLists.txt
+++ b/compendium/ServiceComponent/test/CMakeLists.txt
@@ -137,7 +137,7 @@ foreach(exe_name ${compile_only_exe_names})
 
     #For compile-only tests we add a test with Cmake command to build the target only as we expect a build failure
     add_test(NAME ${exe_name}
-    COMMAND ${CMAKE_COMMAND} --build . --target ${exe_name}
+    COMMAND ${CMAKE_COMMAND} --build . --config ${CMAKE_BUILD_TYPE} --target ${exe_name}
     WORKING_DIRECTORY ${CppMicroServices_BINARY_DIR}
     )
 


jeffdiclemente (approved):

jeffdiclemente (changes requested):
looks like builds are failing, I see that the config option is not set:

```
2023-10-24T14:06:35.9246210Z 13: Test command: C:\Strawberry\c\bin\cmake.exe "--build" "." "--config" "--target" "usServiceComponentRefOrderTest"
2023-10-24T14:06:35.9247350Z 13: Working Directory: Z:/build/config_0
2023-10-24T14:06:35.9248156Z 13: Test timeout computed to be: 600
2023-10-24T14:06:35.9504543Z 13: CMake Error: Invalid value used with --config
2023-10-24T14:06:35.9505956Z 13: Usage: cmake --build <dir>             [options] [-- [native-options]]
2023-10-24T14:06:35.9506937Z 13:        cmake --build --preset <preset> [options] [-- [native-options]]
```

tcormackMW:
@jeffdiclemente Updated code. There was an `AND` statement in the cmake enforcing that, too define `CMAKE_BUILD_TYPE` , both `CMAKE_BUILD_TYPE AND CMAKE_CONFIGURATION_TYPES` must be null... So in our case, `CMAKE_CONFIGURATION_TYPES` was not null and `CMAKE_BUILD_TYPE` was, but it was not getting set.

tcormackMW:
@jeffdiclemente I looked into the difference between CMAKE_CONFIGURATION_TYPES and CMAKE_BUILD_TYPE...

CMAKE_CONFIGURATION_TYPES just 'Specifies the available build types', see [this](https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIGURATION_TYPES.html)...

I'm not sure of the original intention of this line but it doesn't make much sense (to me) why we would insist that the argument containing the possible build types, `CMAKE_CONFIGURATION_TYPES`, is null in order to set the CMAKE_BUILD_TYPE by default.

jeffdiclemente (approved):

PR #951, opened 2023-10-26
Introducing IPO to the CMake build process. 

tcormackMW:
Performance verified locally: does not seem to be causing regressions consistently... subject to consistency of local machine.


CMakeLists.txt
```diff
@@ -254,6 +255,7 @@ us_cache_var(US_BUILD_TESTING OFF BOOL "Build tests")
 us_cache_var(US_BUILD_EXAMPLES OFF BOOL "Build example projects")
 us_cache_var(US_USE_SYSTEM_GTEST OFF BOOL "Build using an external GTest installation" ADVANCED)
 us_cache_var(US_USE_SYSTEM_BOOST OFF BOOL "Build using an external Boost installation" ADVANCED)
+us_cache_var(US_USE_LTO ON BOOL "Use link time optimization if supported" ADVANCED)
```
jeffdiclemente, line 258: this comment refers to "link time optimization" when most of the other code refers to IPO or "interprocedural optimization". I recommend using consistent terms throughout the code.

jeffdiclemente (changes requested):
Is it necessary to call `set_property(TARGET ${us_configurationadmin_test_exe_name} PROPERTY INTERPROCEDURAL_OPTIMIZATION ${CMAKE_ENABLE_LTO})` in each CMakeLists.txt file? Can [CMAKE_INTERPROCEDURAL_OPTIMIZATION](https://cmake.org/cmake/help/latest/variable/CMAKE_INTERPROCEDURAL_OPTIMIZATION.html#variable:CMAKE_INTERPROCEDURAL_OPTIMIZATION) be used to set this property once and have it used for all targets?

CMakeLists.txt
tcormackMW: right you are

PR #964, opened 2023-11-21
Remove CCActiveState deadlock suppression

jeffdiclemente:
local qualification can't reproduce the tsan reported deadlock. Removing the deadlock suppression to see if it still happens.


tcormackMW (approved):
Looks good to me!

PR #966, opened 2023-11-28
Fix Miniz API export

tcormackMW:
Adds a 'fvisibile=hidden' flag to c-compiler line to stop c from exporting miniz APIs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -562,16 +562,12 @@ else()
     usFunctionCheckCompilerFlags("-fvisibility-ms-compat" _have_visibility)
   else()
     usFunctionCheckCompilerFlags("-fvisibility=hidden -fvisibility-inlines-hidden" _have_visibility)
-    usFunctionCheckCompilerFlags("-fvisibility=hidden" _have_visibility_c)
   endif()
 
   if(_have_visibility)
     set(US_HAVE_VISIBILITY_ATTRIBUTE 1)
     set(US_CXX_FLAGS "${US_CXX_FLAGS} ${_have_visibility}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -562,13 +562,18 @@ else()
...
 
   if(_have_visibility)
     set(US_HAVE_VISIBILITY_ATTRIBUTE 1)
     set(US_CXX_FLAGS "${US_CXX_FLAGS} ${_have_visibility}")
   endif()
 
+  if(_have_visibility_c)
+    set(US_C_FLAGS "${US_C_FLAGS} -fvisibility=hidden")
+  endif()
+
```
jeffdiclemente, line 576: the first call to `usFunctionCheckCompilerFlags` already checks for `-fvisibility=hidden`, there isn't a need to check for it again. You can move 
```
set(US_C_FLAGS "${US_C_FLAGS} -fvisibility=hidden")
```
into the `if(_have_visibility)`.

jeffdiclemente (changes requested):

CMakeLists.txt
tcormackMW: Will do. I had done it this way because in usFunctionCheckCompilerFlags there is also some functionality that replaces elements of the flag and cleans it up if need be. So I wanted to put in the exact output of the function so that if there is a character that is not allowed, it would be cleaned up. In this case, I guess we can just put in the value directly. 

As it happens, I made a mistake anyway and didn't use the `_have_visibility_c` variable as I should've when setting the US_C_FLAGS but that is besides the point

PR #977, opened 2023-12-20
Add dynamic targeting functionality for factory services to DS.

pelliott-mathworks:
Add dynamic targeting functionality for factory services to DS.
- move CreateFactoryComponent from ConfigurationNotifier to new class ComponentFactoryImpl
- Process targets from configuration objects in CreateFactoryComponent
- Add one test. (more to be added later). Signed-off by: pelliott@mathworks.com


compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -113,7 +115,7 @@ namespace cppmicroservices
             }
         }
         bool
-        ConfigurationNotifier::AnyListenersForPid(std::string const& pid) noexcept
+        ConfigurationNotifier::AnyListenersForPid(std::string const& pid, std::shared_ptr<cppmicroservices::AnyMap> properties) noexcept
```
jeffdiclemente, line 118: `properties` should be passed into this function as a `const ref AnyMap`. I don't see a need to wrap it in a shared_ptr.

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+            ::test::InstallAndStartConfigAdmin(context);
+            auto svcRef = context.GetServiceReference<cppmicroservices::service::cm::ConfigurationAdmin>();
+            ASSERT_TRUE(svcRef);
+            configAdminSvc = context.GetService(svcRef);
+            ASSERT_TRUE(configAdminSvc);
+        }
+        template <typename ServiceInterfaceT>
+        std::shared_ptr<ServiceInterfaceT>
+        getFactoryService(std::string factoryPid, std::string factoryInstance, cppmicroservices::AnyMap& props)
```
jeffdiclemente, line 56: `props` should be `const`

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include "TestInterfaces/Interfaces.hpp"
+#include "cppmicroservices/LDAPFilter.h"
+#include "cppmicroservices/LDAPProp.h"
+#include "cppmicroservices/servicecomponent/ComponentConstants.hpp"
+#include "gtest/gtest.h"
+
+namespace test
+{
+
+    class tFactoryTarget : public tGenericDSSuite
```
jeffdiclemente, line 33: derive from `tGenericDSAndCASuite` instead so you don't have to re-write `SetUp` and `TearDown`

jeffdiclemente (changes requested):
I thought that dynamic updating of the target property was being added for all services, not just factories. Is this change supposed to dynamically update the `target` property of service component?

There should be more unit tests.

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
pelliott-mathworks: tGenericDSAndCASuite is part of the usConfigurationAdminTests project (TestFixtures.hpp). I didn't think we were allowed to use include files from another project.

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
pelliott-mathworks: Will fix

pelliott-mathworks:
Our requirements document only has use cases for factory services. I'm not aware of any use cases for dynamic targets for non-factory services. I'm not sure it makes sense. 
Having a singleton service dependent on a factory service is not allowed (Section 3.4 of requirements document).
If a singleton is dependent on another singleton the targets in the manifest.json should suffice so we wouldn't need dynamic targets?

jeffdiclemente:
> Our requirements document only has use cases for factory services. I'm not aware of any use cases for dynamic targets for non-factory services. I'm not sure it makes sense. Having a singleton service dependent on a factory service is not allowed (Section 3.4 of requirements document). If a singleton is dependent on another singleton the targets in the manifest.json should suffice so we wouldn't need dynamic targets?

Yes, however I thought that the functionality was to allow the `target` property of a service component's references to be dynamically updated by config admin at runtime, which doesn't care about the type of service (factory, singleton, etc...).

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -1071,13 +1071,16 @@ namespace cppmicroservices
         ConfigurationAdminImpl::RandomInstanceName()
```
carneyweb, line 1071: Should this function be "static"?? I don't see any references to instance data.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -1071,13 +1071,16 @@ namespace cppmicroservices
...
         {
             static auto const characters = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
-            std::uniform_int_distribution<> dist(0, 61);
-            std::string randomString(6, '\0');
-            for (auto& c : randomString)
             {
-                c = characters[dist(randomGenerator)];
+                std::lock_guard<std::mutex> lk { randomInstanceMutex };
```
carneyweb, line 1075: It appears as though this is a re-entrant call. Is a mutex necessary?

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.hpp
```diff
@@ -245,6 +245,7 @@ namespace cppmicroservices
 
             // Used to generate a random instance name for CreateFactoryConfiguration
             std::string RandomInstanceName();
```
carneyweb, line 247: static?

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.hpp
```diff
@@ -245,6 +245,7 @@ namespace cppmicroservices
...
+            std::mutex randomInstanceMutex;
```
carneyweb, line 248: Is this mutex necessary?

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -43,17 +43,16 @@ namespace cppmicroservices
...
-            , logger(std::move(logger))
-            , asyncWorkService(asyncWorkSvc)
-            , extensionRegistry(extensionReg)
-        {
-            if (!bundleContext || !(this->logger) || (!this->asyncWorkService) || (!this->extensionRegistry))
+            , logger(logger)
+            , componentFactory(std::make_shared<ComponentFactoryImpl>(context, logger, asyncWorkSvc, extensionReg))
+ 
+          {
+            if (!context || !(this->logger) || !asyncWorkSvc || !extensionReg || !componentFactory)
```
carneyweb, line 50: This check is already done in ComponentFactoryImpl::ComponentFactoryImpl. It's redundant here.

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
```diff
@@ -234,7 +234,9 @@ namespace cppmicroservices
                         {
                             for (auto const& config : configs)
                             {
-                                configNotifier->CreateFactoryComponent(config->GetPid(), mgr);
+                                auto properties = config->GetProperties();
+                                auto componentFactory = configNotifier->GetComponentFactory();
+                                componentFactory->CreateFactoryComponent(config->GetPid(), mgr, properties );
```
carneyweb, line 239: Like "config->GetPid()", for consistency, "properties" should be passed as "config->GetProperties()" instead of declaring a local variable.

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.hpp
```diff
@@ -96,14 +97,13 @@ namespace cppmicroservices
 
             void UnregisterListener(std::string const& pid, const cppmicroservices::ListenerTokenId token) noexcept;
 
-            bool AnyListenersForPid(std::string const& pid) noexcept;
+            bool AnyListenersForPid(std::string const& pid, cppmicroservices::AnyMap const& properties) noexcept;
```
carneyweb, line 100: Should this function be "const"? I don't see any instance data being modified in the implementation.

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
```diff
@@ -0,0 +1,157 @@
...
+                }
+            }
+            newMetadata->configurationPids.emplace_back(pid);
+
+            // Look for dynamic targets in the references.
+            // A dynamic target will appear in the properties for the configuration object
+            // with the interface name as the key and the target as the value. 
+            for (auto& ref : newMetadata->refsMetadata)
+            {
+               auto interface = ref.interfaceName;
```
tcormackMW, line 90: formatting here, one to few spaces on line 90

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
pelliott-mathworks: We've been following the "standard" in DS that each constructor checks its own input parameters.

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.hpp
pelliott-mathworks: It doesn't compile if I add the const. I think it's the listenersMap.lock function call.

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.hpp
```diff
@@ -0,0 +1,77 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+
+ =============================================================================*/
+
+#ifndef __CPPMICROSERVICES_SCRIMPL_COMPONENTFACTORYIMPL_HPP__
+#define __CPPMICROSERVICES_SCRIMPL_COMPONENTFACTORYIMPL_HPP__
+
+#include "../SCRLogger.hpp"
+#include "ConcurrencyUtil.hpp"
```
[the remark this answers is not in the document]
pelliott-mathworks: Fixed

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
pelliott-mathworks: Added a tGenericDSAndCASuite which is derived from tGenericDSSuite to TestFixture.hpp

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+        void
+        ComponentFactoryImpl::CreateFactoryComponent(std::string const& pid,
+                                                     std::shared_ptr<ComponentConfigurationImpl>& mgr,
+                                                     cppmicroservices::AnyMap const& properties)
+        {
+            // Create the virtual metadata for the factory instance. 
+            // Start with the metadata from the factory 
+            auto oldMetadata = mgr->GetMetadata();
+            auto newMetadata = std::make_shared<ComponentMetadata>(*oldMetadata);
```
tcormackMW, line 68: maybe don't need the additional variable oldMetadata? instead could do:

`auto newMetadata = std::make_shared<ComponentMetadata>(*(mgr->GetMetadata()));` -- not sure what the 'correct' coding practice is here

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
```diff
@@ -0,0 +1,582 @@
...
+        serviceBReg.Unregister();
+        testBundle.Stop();
+     }
+    /* tFactoryTarget.multipleTargetsExistBefore.
+     * ServiceA, ServiceB and ServiceC are all factory instances.
+     * ServiceA~1 is dependent on ServiceB~123 and ServiceC~123
+     * ServiceB~123 and ServiceC~123 are registered before the bundle containing ServiceA is
+     * started.
+     */
+
```
tcormackMW, line 184: extra space

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
```diff
@@ -0,0 +1,582 @@
...
+        auto serviceBReg = registerSvc<test::ServiceBInt, MockServiceBImpl>("ServiceB~123", "ServiceBId", "ServiceB~123");
+   
+        auto serviceA = tracker->WaitForService(std::chrono::milliseconds(500));
+        ASSERT_TRUE(serviceA) << "ServiceA not registered";
+  
+        // Clean up
+        tracker->Close();
+        serviceBReg.Unregister();
+        testBundle.Stop();
+     }
```
tcormackMW, line 177: no return after 177

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
```diff
@@ -0,0 +1,582 @@
...
+        props["test::ServiceBInt"] = std::string("(ServiceBId=ServiceB~123)");
+        props["test::ServiceCInt"] = std::string("(ServiceCId=ServiceC~123)");
+        auto service = getFactoryService<test::ServiceAInt>("ServiceA", "1", props);
+        ASSERT_TRUE(service);
+
+        // Clean up
+        serviceBReg.Unregister();
+        serviceCReg.Unregister();
+        testBundle.Stop();
+    }
```
tcormackMW, line 213: need a return and deleted return to align with proper test

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
```diff
@@ -0,0 +1,582 @@
...
+        auto serviceCReg = registerSvc<test::ServiceCInt, MockServiceCImpl>("ServiceC~123", "ServiceCId", "ServiceC~123");
+  
+        auto serviceA = tracker->WaitForService(std::chrono::milliseconds(500));
+        ASSERT_TRUE(serviceA) << "ServiceA not registered";
+
+        // Clean up
+        tracker->Close();
+        serviceBReg.Unregister();
+        testBundle.Stop();
+    }
```
tcormackMW, line 257: carriage return

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
carneyweb: OK. I'd argue that in this case since those values are not stored in the ConfigurationNotifier but are just passed on to the ComponentFactoryImpl constructor which is what stores them and uses them that that class should be the one to decide what valid arguments are for itself.

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
pelliott-mathworks: You are right. The extra variable is unecessary. Will fix.

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
pelliott-mathworks: Hmmm. I don't think there should be a return here. 177 marks the end of the TEST_F(tFactoryTarget, dependencyExistsAfter test. It wouldn't compile with a return there.

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
pelliott-mathworks: Will fix

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
pelliott-mathworks: see previous comment

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
pelliott-mathworks: Not sure what you mean.

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
tcormackMW: sorry, I meant carriage return. That is my fault. Just need a new line.

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
tcormackMW: New line is what I meant. sorry for confusion.

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -127,6 +126,14 @@ namespace cppmicroservices
                 auto iter = listenersMapHandle->find(pid);
                 if (iter != listenersMapHandle->end())
                 {
+                    // Make sure the properties does not contain any dynamic targets
+                    auto tokenMapPtr = iter->second;
+                    for (auto const& tokenEntry : (*tokenMapPtr))
+                    {
+                        auto listener = tokenEntry.second;
+                        VerifyProperties(properties, listener.mgr);
```
jeffdiclemente, line 134: this method only logs and doesn't really verify. I recommend renaming the method. Also, I question the usefulness of this method; is all this processing necessary to maybe log an error?

compendium/tools/SCRCodeGen/ManifestParserImpl.cpp
```diff
@@ -70,13 +70,14 @@ ManifestParserImplV1::ParseAndGetComponentInfos(Json::Value const& scr) const
             componentInfo.configurationPolicy
                 = JsonValueValidator(jsonComponent, "configuration-policy", policyChoices).GetString();
         }
+        std::unordered_map<std::string, std::string> duplicatePids;
```
jeffdiclemente, line 73: move the construction closer to where its used.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -260,6 +203,30 @@ namespace cppmicroservices
...
-
+        std::shared_ptr<ComponentFactoryImpl>  ConfigurationNotifier::GetComponentFactory() {
+            return componentFactory;
+        }
+        void
+        ConfigurationNotifier::VerifyProperties(cppmicroservices::AnyMap const& properties,
+                                                std::shared_ptr<ComponentConfigurationImpl> mgr) const noexcept
+        {
+            // Look for dynamic targets in the references.
+            // A dynamic target will appear in the properties for the configuration object
```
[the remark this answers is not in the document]
pelliott-mathworks: yes. So instead of the key being the interface name it should probably be the interfacename.target?

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -260,6 +203,30 @@ namespace cppmicroservices
...
+            // with the interface name as the key and the target as the value.
+            auto metadata = mgr->GetMetadata();
```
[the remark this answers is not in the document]
pelliott-mathworks: Will fix

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -260,6 +203,30 @@ namespace cppmicroservices
...
+            for (auto& ref : metadata->refsMetadata)
```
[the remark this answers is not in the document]
pelliott-mathworks: Will fix

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -260,6 +203,30 @@ namespace cppmicroservices
...
+            {
+                auto interface = ref.interfaceName;
+                auto iter = properties.find(interface);
```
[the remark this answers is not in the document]
pelliott-mathworks: Will fix

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -260,6 +203,30 @@ namespace cppmicroservices
...
+                if (iter != properties.end())
```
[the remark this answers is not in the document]
pelliott-mathworks: This code is checking to see if a dynamic target is in the configuration properties. It doesn't come through the properties as part of a ReferenceMetadata structure. It comes through as a key value pair where the key is the interface name (or interfacename.target)

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
pelliott-mathworks: Will fix

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
pelliott-mathworks: Will fix

compendium/DeclarativeServices/test/gtest/TestFactoryTarget.cpp
pelliott-mathworks: Will fix

compendium/tools/SCRCodeGen/ManifestParserImpl.cpp
pelliott-mathworks: This is used in the if statement that follows this declaration but I can't move it inside the if statement because it is used later on line 108. It has to be defined outside the if block

compendium/DeclarativeServices/src/ConfigurationListenerImpl.cpp
```diff
@@ -56,29 +56,24 @@ namespace cppmicroservices
...
                     {
                         logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
-                                    "configurationEvent error. ConfigurationAdmin service "
-                                    "reference is no longer valid");
+                            "configurationEvent error. ConfigurationAdmin service "
+                            "reference is no longer valid");
                         return;
                     }
-                    auto properties = cppmicroservices::AnyMap(AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS);
+                    cppmicroservices::AnyMap properties;
```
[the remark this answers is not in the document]
jeffdiclemente: right, I didn't see that. nevermind then.

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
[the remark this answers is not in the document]
jeffdiclemente: check the OSGi spec, I think there are two ways to specify a target?

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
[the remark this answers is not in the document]
jeffdiclemente: what's the purpose of the `ReferenceMetadata::target` member then? I thought that this member held the value of the 'target' property value.

compendium/tools/SCRCodeGen/ManifestParserImpl.cpp
jeffdiclemente: I missed that. ok, nevermind then.

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.hpp
```diff
@@ -269,7 +268,7 @@ namespace cppmicroservices
                 cppmicroservices::service::cm::ManagedServiceFactory,
                 TrackedServiceWrapper<cppmicroservices::service::cm::ManagedServiceFactory>>
                 managedServiceFactoryTracker;
-            std::mt19937 randomGenerator;
+            thread_local static std::mt19937 randomGenerator;
```
jeffdiclemente, line 271: this doesn't seem right. Why not move `randomGenerator` from being a member variable to a `static thread_local` scoped variable within `ConfigurationAdminImpl::RandomInstanceName()`?

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -165,78 +172,14 @@ namespace cppmicroservices
                     return false;
                 }
             } // release listenersMapHandle lock
+
             for (auto & mgr : mgrs)
             {
-                CreateFactoryComponent(pid, mgr);
+            componentFactory->CreateFactoryComponent(pid, mgr, properties);
```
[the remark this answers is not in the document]
jeffdiclemente: still needs to be fixed.

compendium/DeclarativeServices/test/gtest/TestFixture.hpp
```diff
@@ -314,7 +317,30 @@ namespace test
...
+            configAdmin = context.GetService(svcRef);
+            ASSERT_TRUE(configAdmin);
+        }
+        void
+        TearDown() override
+        {
+            if (configAdmin)
+            {
+                configAdmin.reset();
+            };
```
jeffdiclemente, line 338: Is is necessary to call `reset()`? destroying the `tGenericDSAndCASuite` object does the same thing.

compendium/DeclarativeServices/test/gtest/TestFixture.hpp
```diff
@@ -253,6 +253,9 @@ namespace test
         void
         TearDown() override
         {
+            if (dsRuntimeService) {
+                dsRuntimeService.reset();
+            }
```
jeffdiclemente, line 258: Is is necessary to call `reset()`? destroying the `tGenericDSSuite` object does the same thing.

compendium/test_bundles/TestBundleDSFAC1/src/ServiceImpl.cpp
```diff
@@ -0,0 +1,30 @@
...
+{
+
+    ServiceAImpl::ServiceAImpl(std::shared_ptr<cppmicroservices::AnyMap> const& props,
+                                               const std::shared_ptr<test::ServiceBInt> interface1)
+        : properties(props)
+        {
+        if (nullptr != interface1)
+        {
+          
+        }
```
jeffdiclemente, line 14: this code does nothing. is the if statement supposed to do something? If not, please delete this.

compendium/tools/SCRCodeGen/ManifestParserImpl.cpp
```diff
@@ -103,6 +104,13 @@ ManifestParserImplV1::ParseAndGetComponentInfos(Json::Value const& scr) const
         {
             auto factoryComponentID
                 = JsonValueValidator(jsonComponent, "factory", Json::ValueType::stringValue).GetString();
+            if (!factoryComponentID.empty()) {
+                if (duplicatePids.size() != 1)
```
jeffdiclemente, line 108: reduce nested if statements by combining the two if statements into one.

compendium/test_bundles/TestBundleDSFAC2/src/ServiceImpl.hpp
```diff
@@ -0,0 +1,32 @@
...
+                      const std::shared_ptr<test::ServiceCInt> interface2);
+        void Modified(std::shared_ptr<ComponentContext> const& context,
+                      std::shared_ptr<cppmicroservices::AnyMap> const& configuration);
+        cppmicroservices::AnyMap GetProperties() override;
+ 
+        ~ServiceAImpl2() = default;
+
+      private:
+        std::mutex propertiesLock;
+        std::shared_ptr<cppmicroservices::AnyMap> properties;
```
jeffdiclemente, line 26: there's no reason to make this a `std::shared_ptr`. make this an `AnyMap` value type.

compendium/test_bundles/TestBundleDSFAC2/src/ServiceImpl.cpp
```diff
@@ -0,0 +1,29 @@
...
+                            std::shared_ptr<cppmicroservices::AnyMap> const& configuration)
+    {
+        std::lock_guard<std::mutex> lock(propertiesLock);
+        properties = configuration;
+    }
+    cppmicroservices::AnyMap
+    ServiceAImpl2::GetProperties()
+    {
+        std::lock_guard<std::mutex> lock(propertiesLock);
+        return *properties;
```
jeffdiclemente, line 26: this should not be done, even in test code, it breaks the `std::shared_ptr` ownership model. I recommend changing `properties` to a value type `AnyMap` and then returning `properties` either by value or ref.

compendium/test_bundles/TestBundleDSFAC1/src/ServiceImpl.cpp
```diff
@@ -0,0 +1,30 @@
...
+                                   std::shared_ptr<cppmicroservices::AnyMap> const& configuration)
+    {
+        std::lock_guard<std::mutex> lock(propertiesLock);
+        properties = configuration;
+    }
+    cppmicroservices::AnyMap
+    ServiceAImpl::GetProperties()
+    {
+        std::lock_guard<std::mutex> lock(propertiesLock);
+        return *properties;
```
jeffdiclemente, line 27: this should not be done, even in test code, it breaks the `std::shared_ptr` ownership model. I recommend changing `properties` to a value type `AnyMap` and then returning `properties` either by value or ref.

jeffdiclemente (changes requested):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.hpp
pelliott-mathworks: Will fix

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.cpp
```diff
@@ -1078,6 +1078,7 @@ namespace cppmicroservices
                 c = characters[dist(randomGenerator)];
             }
             return randomString;
+
```
carneyweb, line 1081: remove additional empty line?

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
```diff
@@ -0,0 +1,143 @@
...
+                    {
+                        logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
+                                    "CreateFactoryComponent failed because of invalid target ldap filter"
+                                        + newMetadata->name + " target= " + ref.target);
+                        throw std::invalid_argument(e.what());
+                    }
+                 }
+            }
+
+            auto bundle = mgr->GetBundle();
```
carneyweb, line 102: Should these be declared inside the try block where they are used?

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -260,6 +203,31 @@ namespace cppmicroservices
...
             }
         }
-
+        std::shared_ptr<ComponentFactoryImpl>  ConfigurationNotifier::GetComponentFactory() {
+            return componentFactory;
+        }
+        void
+        ConfigurationNotifier::LogInvalidDynamicTargetInProperties(
+            cppmicroservices::AnyMap const& properties,
+                                                std::shared_ptr<ComponentConfigurationImpl> mgr) const noexcept
```
carneyweb, line 212: indentation.

carneyweb (approved):

compendium/DeclarativeServices/test/gtest/TestFixture.hpp
pelliott-mathworks: Will fix

compendium/test_bundles/TestBundleDSFAC1/src/ServiceImpl.cpp
pelliott-mathworks: Will fix

compendium/test_bundles/TestBundleDSFAC1/src/ServiceImpl.cpp
pelliott-mathworks: Will fix

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
```diff
@@ -0,0 +1,143 @@
...
+                    {
+                        LDAPFilter filter(ref.target);
+                    }
+                    catch (std::exception const& e)
+                    {
+                        logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
+                                    "CreateFactoryComponent failed because of invalid target ldap filter"
+                                        + newMetadata->name + " target= " + ref.target);
+                        throw std::invalid_argument(e.what());
+                    }
```
jeffdiclemente, line 98: why not let the LDAPFilter constructor throw to the caller instead of catching and throwing (which is a performance hit)? The caller can log the exception.

compendium/test_bundles/TestBundleDSFAC2/src/ServiceImpl.hpp
pelliott-mathworks: Will fix

compendium/test_bundles/TestBundleDSFAC2/src/ServiceImpl.cpp
pelliott-mathworks: Will fix

compendium/tools/SCRCodeGen/ManifestParserImpl.cpp
pelliott-mathworks: Will fix

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -165,78 +172,14 @@ namespace cppmicroservices
                     return false;
                 }
             } // release listenersMapHandle lock
+
             for (auto & mgr : mgrs)
             {
-                CreateFactoryComponent(pid, mgr);
+                componentFactory->CreateFactoryComponent(pid, mgr, properties);
```
jeffdiclemente, line 178: the way that `CreateFactoryComponent` now handles exceptions means that this loop could break early and not run `CreateFactoryComponent` on all `mgr` in `mgrs`, which is a change in behavior from before. Its not clear to me how exceptions and failure are supposed to be handled. can you clarify what is supposed to happen in the case of an exception being thrown?

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
```diff
@@ -234,7 +234,10 @@ namespace cppmicroservices
                         {
                             for (auto const& config : configs)
                             {
-                                configNotifier->CreateFactoryComponent(config->GetPid(), mgr);
+                                auto componentFactory = configNotifier->GetComponentFactory();
+                                componentFactory->CreateFactoryComponent(config->GetPid(),
```
jeffdiclemente, line 238: Its not clear to me how exceptions and failure are supposed to be handled. can you clarify what is supposed to happen in the case of an exception being thrown? e.g. are all `config` in `configs` supposed to have a factory component created if one or more cause an exception?

jeffdiclemente (changes requested):

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.hpp
```diff
@@ -268,8 +268,7 @@ namespace cppmicroservices
                 cppmicroservices::service::cm::ManagedServiceFactory,
                 TrackedServiceWrapper<cppmicroservices::service::cm::ManagedServiceFactory>>
                 managedServiceFactoryTracker;
-            thread_local static std::mt19937 randomGenerator;
-            cppmicroservices::ServiceTracker<cppmicroservices::service::cm::ConfigurationListener>
+             cppmicroservices::ServiceTracker<cppmicroservices::service::cm::ConfigurationListener>
```
jeffdiclemente, line 271: formatting.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
pelliott-mathworks: Any exceptions thrown in ComponentConfigurationImpl::Initialize are caught and logged by CMEnabledState::CreateConfigurations. This is true for factory components but it's also true for normal components that have been created with a manifest.json file rather than virtual metadata. OSGI has a bunch of places where it talks about errors happening and the result is that the error should be logged.

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
pelliott-mathworks: If the LDAPFilter is invalid, a std::exception will be thrown and the caller will log whatever message is included in the std::exception. I thought that it would be better to catch and log this here because we can include a more helpful message in the log. I think this is an error that will happen a lot to our users(it happened to me a few times) and warrants a bit more help.

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
pelliott-mathworks: None of those can throw exceptions so not sure it matters. They are just returning a member variable to the caller.

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -127,6 +126,14 @@ namespace cppmicroservices
                 auto iter = listenersMapHandle->find(pid);
```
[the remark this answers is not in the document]
pelliott-mathworks: Doesn't compile if it's const

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
pelliott-mathworks: Hmmm. I don't think it's a change in behavior. If one factory component creation failed, they all failed. If you don't think this is the right behavior I can put a try /catch here so that the other components will still be created if one fails.

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
pelliott-mathworks: Will fix

compendium/ConfigurationAdmin/src/ConfigurationAdminImpl.hpp
pelliott-mathworks: Will fix

Burgch:
> > Our requirements document only has use cases for factory services. I'm not aware of any use cases for dynamic targets for non-factory services. I'm not sure it makes sense. Having a singleton service dependent on a factory service is not allowed (Section 3.4 of requirements document). If a singleton is dependent on another singleton the targets in the manifest.json should suffice so we wouldn't need dynamic targets?
> 
> Yes, however I thought that the functionality was to allow the `target` property of a service component's references to be dynamically updated by config admin at runtime, which doesn't care about the type of service (factory, singleton, etc...).

I can provide use cases where it would be useful to be able to update the target of a singleton service via Config Admin at runtime. One is where we have multiple services which satisfy a reference, but the "best" one (by `service.ranking`) is not the one we want to use. Being able to update the target would allow make it possible to ensure the correct service is chosen.

I think it would be very valuable to have this feature for all configurable service components, instead of only factories. Very happy for factories to be prioritised and released first though, as they are the majority of the use cases I believe!

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
```diff
@@ -127,6 +126,14 @@ namespace cppmicroservices
...
                 if (iter != listenersMapHandle->end())
                 {
+                    // Make sure the properties does not contain any dynamic targets
+                    auto const tokenMapPtr = iter->second;
+                    for (auto const& tokenEntry : (*tokenMapPtr))
+                    {
+                        auto listener = tokenEntry.second;
+                        LogInvalidDynamicTargetInProperties(properties, listener.mgr);               
+                    }
```
jeffdiclemente, line 135: I don't see the value of logging this and I don't understand the restriction of allowing dynamic targets only for factory components. I think this is over constraining and should be removed.

compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp
jeffdiclemente: I don't see how creation of one factory component causing none of the other unrelated factory components to be created is the right behavior. am I missing something?

jeffdiclemente (changes requested):
there are merge conflicts on the development branch that need to be addressed in this PR as well.

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
jeffdiclemente: what is the message thrown from LDAPFilter?

pelliott-mathworks:
Merging latest changes from development into this pull request

jeffdiclemente (approved):

PR #979, opened 2023-12-29
Ensure that bundle.Start() after the framework has stopped throws

tcormackMW:
According to the OSGi [spec](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.api.html#org.osgi.framework.Bundle.start-int-), we must throw if the framework has been stopped when bundle.Start() is called.

### DIFF
--- a/framework/src/bundle/BundlePrivate.cpp
+++ b/framework/src/bundle/BundlePrivate.cpp
@@ -420,8 +420,8 @@ namespace cppmicroservices
     {
         auto l = this->Lock();
         US_UNUSED(l);
-        auto frameworkStopped = coreCtx->GetFrameworkStopped();
-        if (frameworkStopped->stopped)
+        auto frameworkBlock = coreCtx->GetFrameworkStopped();
+        if (frameworkBlock->frameworkHasStopped)
         {
             throw std::runtime_error("Bundle " + symbolicName + " (location=" + location
                                      + ") belongs to a stopped framework");
--- a/framework/src/bundle/BundlePrivate.cpp
+++ b/framework/src/bundle/BundlePrivate.cpp
@@ -74,7 +74,7 @@ namespace cppmicroservices
 
             if (state == Bundle::STATE_UNINSTALLED)
             {
-                throw std::logic_error("Bundle #" + util::ToString(id) + " (location=" + location + ") is uninstalled");
+                throw std::logic_error("Bundle " + symbolicName + " (location=" + location + ") is uninstalled");
             }
 
             if ((options & Bundle::STOP_TRANSIENT) == 0)
@@ -136,7 +136,7 @@ namespace cppmicroservices
             catch (...)
             {
                 res = std::make_exception_ptr(
-                    std::runtime_error("Bundle #" + util::ToString(id) + " (location=" + location
+                    std::runtime_error("Bundle " + symbolicName + " (location=" + location
                                        + "), BundleActivator::Stop() failed: " + util::GetLastExceptionStr()));
             }
 
@@ -165,8 +165,8 @@ namespace cppmicroservices
                 }
                 if (!cause.empty())
                 {
-                    res = std::make_exception_ptr(std::runtime_error(
-                        "Bundle #" + util::ToString(id) + " (location=" + location + ") stop failed: " + cause));
+                    res = std::make_exception_ptr(std::runtime_error("Bundle " + symbolicName + " (location=" + location
+                                                                     + ") stop failed: " + cause));
                 }
             }
             bactivator = nullptr;
@@ -263,7 +263,7 @@ namespace cppmicroservices
                     return;
                 }
             }
-            [[fallthrough]];
+                [[fallthrough]];
             case Bundle::STATE_RESOLVED:
             {
                 state = Bundle::STATE_STARTING;
@@ -286,10 +286,10 @@ namespace cppmicroservices
                 // This happens if call start from inside the BundleActivator.stop
                 // method.
                 // Don't allow it.
-                throw std::runtime_error("Bundle #" + util::ToString(id) + " (location=" + location
+                throw std::runtime_error("Bundle " + symbolicName + " (location=" + location
                                          + "), start called from BundleActivator::Stop");
             case Bundle::STATE_UNINSTALLED:
-                throw std::logic_error("Bundle #" + util::ToString(id) + " (location=" + location
+                throw std::logic_error("Bundle " + symbolicName + " (location=" + location
                                        + ") is in UNINSTALLED state");
         }
     }
@@ -304,7 +304,7 @@ namespace cppmicroservices
             switch (static_cast<Bundle::State>(state.load()))
             {
                 case Bundle::STATE_UNINSTALLED:
-                    throw std::logic_error("Bundle #" + util::ToString(id) + " (location=" + location
+                    throw std::logic_error("Bundle " + symbolicName + " (location=" + location
                                            + ") is in BUNDLE_UNINSTALLED state");
                 case Bundle::STATE_STARTING: // Lazy start
                 case Bundle::STATE_ACTIVE:
@@ -337,7 +337,7 @@ namespace cppmicroservices
                         }
                     }
                 }
-                [[fallthrough]];
+                    [[fallthrough]];
                 case Bundle::STATE_RESOLVED:
                 case Bundle::STATE_INSTALLED:
                 {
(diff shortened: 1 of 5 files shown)


framework/src/bundle/CoreBundleContext.cpp
```diff
@@ -250,4 +251,16 @@ namespace cppmicroservices
...
+    void CoreBundleContext::StopFramework() {
+        this->self.Lock(), stopped = true;
+    }
+
+    void CoreBundleContext::StartFramework() {
+        this->self.Lock(), stopped = false;
+    }
+
+    bool CoreBundleContext::FrameworkStopped() const {
+        return this->self.Lock(), stopped;
```
jeffdiclemente, line 264: can these functions be re-named to something more accurate and descriptive? The functions only set a flag which indicates the state of the framework.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -420,6 +420,10 @@ namespace cppmicroservices
     {
         auto l = this->Lock();
         US_UNUSED(l);
+        if(coreCtx->FrameworkStopped()){
+            throw std::runtime_error("Bundle #" + util::ToString(id) + " (location=" + location
```
jeffdiclemente, line 424: I find that logging the bundle's symbolic name is more useful than the bundle id.

framework/src/bundle/CoreBundleContext.h
```diff
@@ -197,6 +202,7 @@ namespace cppmicroservices
         // The core context is exclusively constructed by the FrameworkFactory class
         friend class FrameworkFactory;
 
+        bool stopped;
```
jeffdiclemente, line 205: can this be a `atomic_bool`? Given how this variable is used, it also removes the need to lock a mutex in `Stop/StartFramework` and `FrameworkStopped`.

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -420,6 +420,10 @@ namespace cppmicroservices
     {
         auto l = this->Lock();
         US_UNUSED(l);
+        if(coreCtx->FrameworkStopped()){
```
jeffdiclemente, line 423: there is still a race when checking the state of the framework since the return value of `stopped` could change after this function returns but before the if statement is evaluated.

jeffdiclemente (changes requested):

framework/src/bundle/CoreBundleContext.cpp
tcormackMW: Fixed

framework/src/bundle/CoreBundleContext.h
tcormackMW: Fixed

framework/src/bundle/BundlePrivate.cpp
tcormackMW: fixed. was following other throw's within this file, but have updated to use symbolic name. May be worth converting all throws to use one or the other at some point

jeffdiclemente: I agree its worth converting other throw messages. If its not too much effort, can you convert the other throw messages in this file? Thanks.

tcormackMW:
No performance regressions found using local benchmarking

framework/src/bundle/CoreBundleContext.h
```diff
@@ -56,10 +56,22 @@ as specified in the OSGi R4.2 specifications.
...
 #include <map>
 #include <ostream>
+#include <shared_mutex>
 #include <string>
 
 namespace cppmicroservices
 {
+    using shared_lock = std::shared_mutex;
+    using WriteLock = std::unique_lock<shared_lock>;
+    using ReadLock = std::shared_lock<shared_lock>;
```
jeffdiclemente, line 66: the use of the word `lock` for a mutex and for a C++ `*_lock` type is confusing. I recommend removing `using shared_lock = std::shared_mutex;` and reference `std::shared_mutex`. The `using` statement is unnecessary.

framework/src/bundle/CoreBundleContext.h
```diff
@@ -187,6 +199,11 @@ namespace cppmicroservices
 
         void Uninit1();
 
+        WriteLock SetFrameworkStopped();
+
+        WriteLock SetFrameworkStarted();
+
+        std::unique_ptr<BlockFrameworkShutdown> GetFrameworkStopped() const;
```
jeffdiclemente, line 206: `GetFrameworkStopped` is not a descriptive name for what this method is meant to do. This method blocks the `stopped` flag from being modified until `BlockFrameworkShutdown` is destroyed.

framework/src/bundle/CoreBundleContext.h
```diff
@@ -187,6 +199,11 @@ namespace cppmicroservices
 
         void Uninit1();
 
+        WriteLock SetFrameworkStopped();
+
+        WriteLock SetFrameworkStarted();
```
jeffdiclemente, line 204: these are better names. I think there is room for improvement on the naming. Also, as a developer who may have to add calls to these methods in the cppms code base, I don't know how to use these methods. Some comments on what these methods do needs to added or the method name made more descriptive.

jeffdiclemente (changes requested):

framework/src/bundle/CoreBundleContext.h
tcormackMW: I agree. Fixed

framework/src/bundle/CoreBundleContext.h
tcormackMW: I agree

framework/src/bundle/CoreBundleContext.h
tcormackMW: Changed the names... I think they are better, but probably not quite right yet

framework/src/bundle/CoreBundleContext.h
```diff
@@ -187,6 +198,31 @@ namespace cppmicroservices
...
+         */
+        WriteLock BlockDuringFrameworkShutdown();
+
+        /**
+         * Called when framework startup has begun.
+         * This blocks (while returned object is held):
+         *     - other calls to start or stop the framework
+         *     - calls to start bundles
+         */
+        WriteLock BlockDuringFrameworkStartup();
```
jeffdiclemente, line 215: since these methods are setting state and returning a way to guard against other threads modifying the framework state, what if they are called:
```
SetFrameworkStateToStoppedAndBlockUntilComplete
SetFrameworkStateToStartedAndBlockUntilComplete
```
alternatively, you could pass in a bool to one method named:
```
SetFrameworkStateAndBlockUntilComplete
```

yes, these methods are wordy, but I think its worth it. Also, this solution surfaces a more widespread issue with how we start and stop the Framework and ensure that concurrent public Framework API calls function without crashing. I would eventually like to see a state machine type solution, similar to what is done in Declarative Services. however that would take more time and effort and this is a reasonable solution for now.

jeffdiclemente (changes requested):

framework/src/bundle/CoreBundleContext.h
```diff
@@ -197,6 +233,13 @@ namespace cppmicroservices
         // The core context is exclusively constructed by the FrameworkFactory class
         friend class FrameworkFactory;
 
+        // Mutex required to be held when changing stopped.
+        // ReadLock or WriteLock construction is done using this mutex.
+        mutable std::shared_mutex stoppedLock;
+
+        // Flag for whether the Framework has been stopped. See mutex stoppedLock
+        bool stopped;
```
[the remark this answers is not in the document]
tcormackMW: I see that the bundle queries its own state. The crux here is the communication between any given `BundlePrivate` and the `FrameworkPrivate`. I think that the `BundlePrivate::state` variable is only owned and visible to that bundle and therefore is not going to be able to link with changes made in the `Frameworkprivate`. The `CoreBundleContext` is the variable that I found that linked those two independent objects. Let me know if this is not your understanding because yes it would be ideal to not have to introduce new methods etc here.

framework/src/bundle/CoreBundleContext.h
```diff
@@ -187,6 +198,23 @@ namespace cppmicroservices
...
+        WriteLock SetFrameworkStateAndBlockUntilComplete(bool desiredState);
+
+        /**
+         * Called when bundle startup is occuring
+         * This blocks (while returned object is held):
+         *    - calls to start or stop the framework
+         * And allows:
+         *    - concurrent calls for bundle startup on other bundles
+         */
+        std::unique_ptr<FrameworkShutdownBlocker> BlockFrameworkShutdown() const;
```
tcormackMW, line 216: `GetFrameworkStateAndBlock`

tcormackMW: fixed

jeffdiclemente (approved):

PR #983, opened 2024-01-04
Ensure safe concurrent destruction of bundles and framework stopping

tcormackMW:
This addresses issues encountered when concurrently destroying bundles and stopping the framework. The solution was to allow a service to be already unregistered when the state was switching from active in the ComponentConfiguration.

### DIFF
--- a/compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
+++ b/compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
@@ -105,7 +105,16 @@ namespace cppmicroservices
                     // the latch counts down to 0, thereby allowing all Activate, Rebind and Modified
                     // activities to complete.
                     currentState->WaitForTransitionTask(); // wait for the previous transition to finish
-                    mgr.UnregisterService();
+                    try
+                    {
+                        mgr.UnregisterService();
+                    }
+                    catch (...)
+                    {
+                        mgr.GetLogger()->Log(cppmicroservices::logservice::SeverityLevel::LOG_WARNING,
+                                             "Failed to unregister the service.",
+                                             std::current_exception());
+                    }
                     mgr.DestroyComponentInstances();
                     transitionAction.set_value();
                 }
--- a/compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
+++ b/compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
@@ -128,7 +128,7 @@ namespace cppmicroservices
                                                                                        mockRegistry,
                                                                                        fakeLogger,
                                                                                        notifier);
-                 EXPECT_EQ(fakeCompConfig->GetConfigState(), ComponentState::UNSATISFIED_REFERENCE);
+                EXPECT_EQ(fakeCompConfig->GetConfigState(), ComponentState::UNSATISFIED_REFERENCE);
                 EXPECT_EQ(fakeCompConfig->regManager, nullptr);
                 EXPECT_EQ(fakeCompConfig->referenceManagers.size(), static_cast<size_t>(0));
             });
@@ -147,9 +147,9 @@ namespace cppmicroservices
             auto notifier = std::make_shared<ConfigurationNotifier>(GetFramework().GetBundleContext(),
                                                                     fakeLogger,
                                                                     asyncWorkService,
-                                                                    extRegistry); 
+                                                                    extRegistry);
             std::set<unsigned long> idSet;
-            const size_t iterCount = 10;
+            size_t const iterCount = 10;
             for (size_t i = 0; i < iterCount; ++i)
             {
                 EXPECT_NO_THROW({
@@ -201,7 +201,7 @@ namespace cppmicroservices
                                                                                    mockRegistry,
                                                                                    fakeLogger,
                                                                                    notifier);
-           EXPECT_CALL(*fakeCompConfig, GetFactory()).Times(1).WillOnce(testing::Return(mockFactory));
+            EXPECT_CALL(*fakeCompConfig, GetFactory()).Times(1).WillOnce(testing::Return(mockFactory));
             // add the mock reference managers to the config object
             fakeCompConfig->referenceManagers.insert(std::make_pair("ref1", refMgr1));
             fakeCompConfig->referenceManagers.insert(std::make_pair("ref2", refMgr2));
@@ -225,7 +225,7 @@ namespace cppmicroservices
             auto mockMetadata = std::make_shared<metadata::ComponentMetadata>();
             auto mockRegistry = std::make_shared<MockComponentRegistry>();
             auto fakeLogger = std::make_shared<FakeLogger>();
- 
+
             auto logger = std::make_shared<SCRLogger>(GetFramework().GetBundleContext());
             auto asyncWorkService
                 = std::make_shared<cppmicroservices::scrimpl::SCRAsyncWorkService>(GetFramework().GetBundleContext(),
@@ -270,7 +270,7 @@ namespace cppmicroservices
             auto mockMetadata = std::make_shared<metadata::ComponentMetadata>();
             auto mockRegistry = std::make_shared<MockComponentRegistry>();
             auto fakeLogger = std::make_shared<FakeLogger>();
- 
+
             // Test that a call to Register with a component containing both a service
             // and a reference to the same service interface will not cause a state change.
             scrimpl::metadata::ReferenceMetadata refMetadata {};
@@ -473,7 +473,7 @@ namespace cppmicroservices
             auto notifier = std::make_shared<ConfigurationNotifier>(GetFramework().GetBundleContext(),
                                                                     fakeLogger,
                                                                     asyncWorkService,
-                                                                    extRegistry); 
+                                                                    extRegistry);
             // Test for exception from user code
             auto fakeCompConfig = std::make_shared<MockComponentConfigurationImpl>(mockMetadata,
                                                                                    GetFramework(),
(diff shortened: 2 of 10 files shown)


jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
```diff
@@ -936,6 +936,64 @@ namespace cppmicroservices
...
+                    });
+                std::thread frameworkT = std::thread(
+                    [&framework]()
+                    {
+                        framework.Stop();
+                        framework.WaitForStop(std::chrono::milliseconds::zero());
+                    });
+
+                bundleT.join();
+                frameworkT.join();
```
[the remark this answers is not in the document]
tcormackMW: I looked at this. This function (as far as I see) is running the same function x number of times concurrently, whereas we just want to run two functions concurrently with eachother. I suppose we could run the `bundle.Stop()` functionality concurrently so that all bundles are stopping at the same time as eachother, but I am not sure that is necessary.

compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
```diff
@@ -936,6 +936,64 @@ namespace cppmicroservices
...
+            auto framework = cppmicroservices::FrameworkFactory().NewFramework();
+            framework.Start();
+            ASSERT_TRUE(framework);
+
+            auto context = framework.GetBundleContext();
+            ASSERT_TRUE(context);
+
+            test::InstallAndStartDS(context);
+
+            std::vector<std::string> bundlesToInstall
```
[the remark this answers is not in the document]
tcormackMW: https://github.com/CppMicroServices/CppMicroServices/pull/983/files#:~:text=InstallAndStartDS(context)%3B-,std%3A%3Avector%3Ccppmicroservices%3A%3ABundle%3E%20installedBundles,%7D%3B,-for%20(auto

deleted that vector, see other comment for more about const

compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
```diff
@@ -936,6 +936,64 @@ namespace cppmicroservices
...
+
+            interfaces.push_back(context.GetServiceReference<test::DSGraph01>());
+            interfaces.push_back(context.GetServiceReference<test::DSGraph02>());
+            interfaces.push_back(context.GetServiceReference<test::DSGraph03>());
+            interfaces.push_back(context.GetServiceReference<test::DSGraph04>());
+            interfaces.push_back(context.GetServiceReference<test::DSGraph05>());
+            interfaces.push_back(context.GetServiceReference<test::DSGraph06>());
+            interfaces.push_back(context.GetServiceReference<test::DSGraph07>());
+
+            for (size_t i = 0; i < interfaces.size(); ++i)
```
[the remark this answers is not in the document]
tcormackMW: https://github.com/CppMicroServices/CppMicroServices/pull/983/files#:~:text=for%20(auto%20const%26%20sref%20%3A%20interfaces)

fixed

compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
```diff
@@ -936,6 +936,64 @@ namespace cppmicroservices
...
+            ASSERT_TRUE(framework);
+
+            auto context = framework.GetBundleContext();
+            ASSERT_TRUE(context);
+
+            test::InstallAndStartDS(context);
+
+            std::vector<std::string> bundlesToInstall
+                = { "DSGraph01", "DSGraph02", "DSGraph03", "DSGraph04", "DSGraph05", "DSGraph06", "DSGraph07" };
+            std::vector<cppmicroservices::ServiceReferenceU> interfaces;
```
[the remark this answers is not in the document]
tcormackMW: something like [this](https://github.com/CppMicroServices/CppMicroServices/pull/983/files#:~:text=InstallAndStartDS(context)%3B-,std%3A%3Avector%3Ccppmicroservices%3A%3ABundle%3E%20installedBundles,%7D%3B,-for%20(auto)? 

I had to leave `installedBundles` as non const in order to call `bundle.Stop()` later on.

tcormackMW:
Local benchmarking do not sure performance regression @jeffdiclemente

jeffdiclemente (changes requested):

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -825,8 +825,20 @@ namespace cppmicroservices
...
+            // wrap in try-catch to catch failures if service is already unregistered
+            // if service is unregistered, all work in UngetService is already done by Unregister() previously
+            try
+            {
+                auto ref = i->GetReference(std::string());
+                ref.d.Load()->UngetService(this->shared_from_this(), false);
+            }
+            catch (...)
+            {
+                coreCtx->logger->Log(logservice::SeverityLevel::LOG_INFO,
```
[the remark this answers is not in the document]
tcormackMW: fixed. Other PR (for concurrent bundle start and framework stop) fixes all instances of referencing bundleid rather than symbolic name in this file, so I will just update this message rather than all in this PR.

framework/src/service/ServiceRegistrationBase.cpp
```diff
@@ -209,28 +209,34 @@ namespace cppmicroservices
...
+            return;
+        }
+
+        // d->coreInfo->unregistering is now true
+
+        // if unavailable
         if (!d->coreInfo->available)
         {
+            // reset isUnregistering to true
+            isUnregistering = true;
```
jeffdiclemente, line 226: I don't understand the logic of setting `isUnregistering` to `true`. If `d->coreInfo->available` is `false`, doesn't that mean that there is no unregistering taking place?

framework/src/service/ServiceRegistrationBase.cpp
```diff
@@ -326,7 +332,10 @@ namespace cppmicroservices
             d->coreInfo->bundleServiceInstance.clear();
 
             d->reference = nullptr;
-            d->coreInfo->unregistering = false;
+
+            // reset d->coreInfo->unregistering to false
+            isUnregistering = true;
```
jeffdiclemente, line 337: I don't understand the logic of setting `isUnregistering` to true. At this point, isn't unregistering finished and shouldn't `isUnregistering` be `false`?

Is this solely so that `atomic_compare_exchange_strong` can be called? If so why not just say `d->coreInfo->unregistering = false;`?

framework/src/service/ServiceRegistrationBase.cpp
tcormackMW: if `d->coreInfo->available` is `false` then that means that unregistration has already happened. This means we should not unregister again at this time. Therefore, we need to exit this function using a throw. That being said, before that I am setting `d->coreInfo->unregistering` to false because we are no longer in this function unregistering. To do that, I first set `isUnregistering` to true as that is the expected value in `atomic_compare_exchange_strong`.

framework/src/service/ServiceRegistrationBase.cpp
tcormackMW: This is solely to call the atomic_compare_Exchange_strong... But I didn't like that in order to start unregistering we have to use an atomic compare, but to exit unregistration we can just set the variable... Unless that is just so that we can compare and set and setting a boolean is an atomic operation... which I dont believe it is

jeffdiclemente: `d->coreInfo->unregistering` is a `std::atomic<bool>` and the `operator=` of `atomic<bool>` is the same as [`atomic<T>::store`](https://en.cppreference.com/w/cpp/atomic/atomic/store) - https://stackoverflow.com/questions/18850752/must-i-call-atomic-load-store-explicitly

framework/src/service/ServiceRegistrationBase.cpp
jeffdiclemente: if the unregistration already happened, then `isUnregistering` should be `false`, right? since the unregistration already happened and this code running on the thread is not performing any unregistration.

is an atomic compare and exchange necessary? or is setting the atomic `d->coreInfo->unregistering` variable enough?

framework/src/service/ServiceRegistrationBase.cpp
tcormackMW: ah ok. Yep that makes sense. Sorry, was looking at wrong variable... but yes, then I agree

jeffdiclemente (approved):

compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
```diff
@@ -105,7 +105,16 @@ namespace cppmicroservices
...
+                    try
+                    {
+                        mgr.UnregisterService();
+                    }
+                    catch (...)
+                    {
+                        mgr.GetLogger()->Log(cppmicroservices::logservice::SeverityLevel::LOG_WARNING,
+                                             "Failed to unregister the service.",
+                                             std::current_exception());
+                    }
```
Burgch, line 117: This makes sense to me - it means we'll definitely get to the next lines. Do we not want the same protection in `CCActiveState` in [this transition](https://github.com/CppMicroServices/CppMicroServices/blob/64860b15acd0dd17a234651e19a393d77de89f21/compendium/DeclarativeServices/src/manager/states/CCSatisfiedState.cpp#L47) ?

Burgch (commented):
Thanks again for this Toby - looks great!

compendium/DeclarativeServices/src/manager/states/CCActiveState.cpp
tcormackMW: Hmm good question. We didn't see this as a failure mode but that may just be a function of the test case. For that to happen we'd have to see this race while the component is satisfied but not active (concurrent shutdown and bundle stop of a satisfied but not active component). That seems like it could happen but is just less likely, probably why we haven't seen it as a failure mode. @jeffdiclemente what do you think about this

PR #985, opened 2024-01-12
DS/CA race on creation of Configuration and ComponentConfigurationImpl

tcormackMW:
This fixes a race when you have a concurrent `Configuration` creation and the creation of a `ComponentConfigurationImpl` object that depends on that configuration. 

What happens is...

in threadA: `compConfig->Initialize()` is called. This causes that config to register a listener for the configuration it depends on and call `configManager->Initialize();`. The `configManager` then checks for a configuration with the proper PID. In the race, this configuration has been created and updated on threadB (see threadB description). Therefore, it creates the properties and returns. The `configManager` has marked the config as satisfied. Therefore, when the `compConfig` checks whether the config is satisfied, it sees that it is and so the component is Registered and set to an active state. 

in threadB: `configAdminService->GetConfiguration` and `configuration->UpdateIfDifferent` are called to set the configuration as valid. In a typical situation, this would not yield a `Modifed` call on the `serviceComponent` as the update is the first update to occur. In this case, the `UpdateIfDifferent` call occurs and the changeCount of the configuration is marked as > 0. Then the `NotifyConfigurationUpdated` call is made which eventually propogates asynchronously to `componentConfigurationImpl::ConfigChangedState`.

The race occurs: Between the time when the changeCount is updated, and `componentConfigurationImpl::ConfigChangedState` is received, the componentConfiguration is `initialize`d, marks the state as active. Then when the `componentConfigurationImpl::ConfigChangedState` call is received, it checks `if (configWasSatisfied && configNowSatisfied...)`. Because it was satisfied on initialization and remains satisfied now, both checks are true even though the 'change' seen on initialization and the change it is receiving right now is the same change. It therefore thins this is another modification after initialization and calls modified. 

To fix this, we assosciate the `changeCount` of any configuration as an additional check to verify that two checks on the same config will not show a change if the `changeCount` remains the same. Essentially, it is a unique id for each new configuration.


compendium/DeclarativeServices/src/ConfigurationListenerImpl.cpp
```diff
@@ -80,17 +80,18 @@ namespace cppmicroservices
                     }
                     auto properties = cppmicroservices::AnyMap(AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS);
                     auto type = event.getType();
-
+                    auto changeCount_ = -1;
```
carneyweb, line 83: Identifier names that end in "_" are usually fields in a class, not local variables.

carneyweb (approved):

jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
```diff
@@ -981,6 +983,97 @@ namespace cppmicroservices
...
+            // Publish ConfigurationListener
+            auto configListener
+                = std::make_shared<cppmicroservices::service::cm::ConfigurationListenerImpl>(frameworkContext,
+                                                                                             logger,
+                                                                                             notifier);
+            auto configListenerReg
+                = frameworkContext.RegisterService<cppmicroservices::service::cm::ConfigurationListener>(
+                    std::move(configListener));
+
+#    if defined(US_BUILD_SHARED_LIBS)
```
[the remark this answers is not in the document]
tcormackMW: This was based on what is done in tServiceComponent test fixture -- but I have changed it to remove the "else"

compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
```diff
@@ -981,6 +983,97 @@ namespace cppmicroservices
...
+            for (auto& bundle : dsbundles)
+            {
+                try
+                {
+                    bundle.Start();
+                }
+                catch (std::exception& e)
+                {
+                    std::cerr << "    " << e.what();
+                }
```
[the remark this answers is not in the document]
tcormackMW: removed

compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
```diff
@@ -981,6 +983,97 @@ namespace cppmicroservices
...
+            }
+#    else
+            auto dsbundles = frameworkContext.GetBundles();
+            for (auto& bundle : dsbundles)
+            {
+                try
+                {
+                    bundle.Start();
+                }
+                catch (std::exception& e)
```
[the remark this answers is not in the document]
tcormackMW: removed

compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
```diff
@@ -981,6 +983,96 @@ namespace cppmicroservices
...
+                                                                                    framework,
+                                                                                    mockRegistry,
+                                                                                    fakeLogger,
+                                                                                    notifier);
+            auto mockCompContext = std::make_shared<MockComponentContextImpl>(compConfig);
+            auto mockCompInstance = std::make_shared<MockComponentInstance>();
+            compConfig->SetComponentInstancePair(InstanceContextPair(mockCompInstance, mockCompContext));
+
+            EXPECT_CALL(*mockCompInstance, Modified()).Times(0);
+            EXPECT_CALL(*mockCompInstance, DoesModifiedMethodExist()).WillRepeatedly(testing::Return(true));
```
jeffdiclemente, line 1054: There should be another test/testpoint that checks a component instances without a `Modified` method implementation also does the right thing (i.e. the service component is not destroyed and created again).

jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
tcormackMW: [Fixed](https://github.com/CppMicroServices/CppMicroServices/pull/985/commits/e5e53bbd1270621344d13763bc0439f69122c135 ). No new test needed

compendium/DeclarativeServices/src/manager/ConfigurationManager.hpp
```diff
@@ -94,12 +96,13 @@ namespace cppmicroservices
...
 
             std::shared_ptr<cppmicroservices::logservice::LogService> logger; ///< logger for this runtime
-            const std::shared_ptr<const metadata::ComponentMetadata> metadata;
+            std::shared_ptr<metadata::ComponentMetadata const> const metadata;
             cppmicroservices::BundleContext bundleContext; ///< context of the bundle which contains the component
             mutable std::mutex propertiesMutex; // mutex to protect the configProperties and mergedProperties members
             std::unordered_map<std::string, cppmicroservices::AnyMap>
                 configProperties; // properties for available configuration objects.
             cppmicroservices::AnyMap mergedProperties;
+            int changeCount;
```
jeffdiclemente, line 105: `changeCount` should be `unsigned long`; its never going to be a value less than 0.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/src/manager/ConfigurationManager.hpp
tcormackMW: fixed.

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
```diff
@@ -73,6 +74,7 @@ namespace cppmicroservices
                         auto config = configAdmin->ListConfigurations("(pid=" + pid + ")");
                         if (config.size() > 0)
                         {
+                            changeCount = config.front()->GetChangeCount();
```
jeffdiclemente, line 77: is it possible that this `changeCount` variable can be overwritten with a value that still causes the race to occur. Since this code is looping over all the configurations associated with the declared pids, it seems possible that a hypothetical config1 has a change count of 2 and config2 has a change count of 1, causing the `ConfigurationManager::changeCount` to be 1. This appears like it could either cause the same race to occur (i.e. Modified is called when it shouldn't be) or it could cause Modified **not** to be called when it should.

tcormackMW: Yes I would agree on first glance. The issue boiled down is that the Configuration Manager manages multiple pids and we have stored just one. So we need to store a set of changeCounts, one for each pid. @jeffdiclemente

tcormackMW: @jeffdiclemente Two new tests added. Both fail prior to updating ChangeCount structure to map<string, unsigned long> rather than unsigned long. Both pass 10000 runs after update.

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
```diff
@@ -36,12 +36,15 @@ namespace cppmicroservices
...
-            , changeCount(0)
         {
             if (!this->metadata || !this->bundleContext || !this->logger)
             {
                 throw std::invalid_argument("ConfigurationManagerImpl - Invalid arguments passed to constructor");
             }
+            for (auto const& pid : metadata->configurationPids)
+            {
+                changeCount[pid] = 0;
+            }
```
jeffdiclemente, line 47: is this necessary? doesn't the `unordered_map` guarantee zero initialized values if the key doesn't exist?

tcormackMW: yes you're correct, I verified this statement and it does return 0

jeffdiclemente (approved):

PR #988, opened 2024-01-25
Added GitHub action to lint new changes with clang-tidy

AntiEXE:
ClangTidyDiffReview action uses the clang-tidy-review action available on the marketplace.
This is aimed to lint the new changes being made to the repository. The clang-tidy warnings and errors are given as a comment to the PR. Making new commit to the same PR will give new comments.

The .clang-tidy file can be used to configure the clang-tidy checks for the action.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,44 @@
+---
+Checks:          '-*,bugprone-*,cppcoreguidelines-*,modernize-*,concurrency-*'
+WarningsAsErrors: ''
+HeaderFilterRegex: ''
+AnalyzeTemporaryDtors: false
+FormatStyle:     file
+CheckOptions:
+  - key:             cert-dcl16-c.NewSuffixes
+    value:           'L;LL;LU;LLU'
+  - key:             cert-oop54-cpp.WarnOnlyIfThisHasSuspiciousField
+    value:           '0'
+  - key:             cert-str34-c.DiagnoseSignedUnsignedCharComparisons
+    value:           '0'
+  - key:             cppcoreguidelines-explicit-virtual-functions.IgnoreDestructors
+    value:           '1'
+  - key:             cppcoreguidelines-non-private-member-variables-in-classes.IgnoreClassesWithAllMemberVariablesBeingPublic
+    value:           '1'
+  - key:             google-readability-braces-around-statements.ShortStatementLines
+    value:           '1'
+  - key:             google-readability-function-size.StatementThreshold
+    value:           '800'
+  - key:             google-readability-namespace-comments.ShortNamespaceLines
+    value:           '10'
+  - key:             google-readability-namespace-comments.SpacesBeforeComments
+    value:           '2'
+  - key:             llvm-else-after-return.WarnOnConditionVariables
+    value:           '0'
+  - key:             llvm-else-after-return.WarnOnUnfixable
+    value:           '0'
+  - key:             llvm-qualified-auto.AddConstToQualified
+    value:           '0'
+  - key:             modernize-loop-convert.MaxCopySize
+    value:           '16'
+  - key:             modernize-loop-convert.MinConfidence
+    value:           reasonable
+  - key:             modernize-loop-convert.NamingStyle
+    value:           CamelCase
+  - key:             modernize-pass-by-value.IncludeStyle
+    value:           llvm
+  - key:             modernize-replace-auto-ptr.IncludeStyle
+    value:           llvm
+  - key:             modernize-use-nullptr.NullMacros
+    value:           'NULL'
+...
--- a/.github/workflows/clang_tidy_diff_review.yml
+++ b/.github/workflows/clang_tidy_diff_review.yml
@@ -1,0 +1,37 @@
+name: ClangTidyDiffReview
+
+# You can be more specific, but it currently only works on pull requests
+on: [pull_request]
+
+jobs:
+  review:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+
+    # Optionally generate compile_commands.json
+
+    - name: Checkout submodules
+      run: git submodule update --init --recursive
+
+    - uses: ZedThree/clang-tidy-review@v0.14.0
+      id: review
+      with:
+        build_dir: build
+        config_file: ".clang-tidy"
+        max_comments: '100'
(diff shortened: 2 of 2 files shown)


tcormackMW (approved):

.clang-tidy
```diff
@@ -0,0 +1,44 @@
...
+  - key:             cert-str34-c.DiagnoseSignedUnsignedCharComparisons
+    value:           '0'
+  - key:             cppcoreguidelines-explicit-virtual-functions.IgnoreDestructors
+    value:           '1'
+  - key:             cppcoreguidelines-non-private-member-variables-in-classes.IgnoreClassesWithAllMemberVariablesBeingPublic
+    value:           '1'
+  - key:             google-readability-braces-around-statements.ShortStatementLines
+    value:           '1'
+  - key:             google-readability-function-size.StatementThreshold
+    value:           '800'
```
jeffdiclemente, line 21: this check isn't something I want to flag right now. let's remove it.

.clang-tidy
```diff
@@ -0,0 +1,44 @@
...
+  - key:             google-readability-namespace-comments.ShortNamespaceLines
+    value:           '10'
+  - key:             google-readability-namespace-comments.SpacesBeforeComments
+    value:           '2'
```
jeffdiclemente, line 25: this file doesn't specify google-* checks. do these check options do anything? shouldn't they be removed since the checks don't run?

.clang-tidy
```diff
@@ -0,0 +1,44 @@
...
+Checks:          '-*,bugprone-*,cppcoreguidelines-*,modernize-*,concurrency-*'
+WarningsAsErrors: ''
+HeaderFilterRegex: ''
+AnalyzeTemporaryDtors: false
+FormatStyle:     file
+CheckOptions:
+  - key:             cert-dcl16-c.NewSuffixes
+    value:           'L;LL;LU;LLU'
+  - key:             cert-oop54-cpp.WarnOnlyIfThisHasSuspiciousField
+    value:           '0'
```
jeffdiclemente, line 11: don't we want this to be the default value of `true`?

.clang-tidy
```diff
@@ -0,0 +1,44 @@
...
+  - key:             cert-str34-c.DiagnoseSignedUnsignedCharComparisons
+    value:           '0'
```
jeffdiclemente, line 13: don't we want the default value of `true`?

.clang-tidy
```diff
@@ -0,0 +1,44 @@
...
+  - key:             cppcoreguidelines-explicit-virtual-functions.IgnoreDestructors
+    value:           '1'
```
jeffdiclemente, line 15: we should use the default of `false`

.clang-tidy
```diff
@@ -0,0 +1,44 @@
...
+  - key:             google-readability-namespace-comments.ShortNamespaceLines
+    value:           '10'
+  - key:             google-readability-namespace-comments.SpacesBeforeComments
+    value:           '2'
+  - key:             llvm-else-after-return.WarnOnConditionVariables
+    value:           '0'
+  - key:             llvm-else-after-return.WarnOnUnfixable
+    value:           '0'
+  - key:             llvm-qualified-auto.AddConstToQualified
+    value:           '0'
```
jeffdiclemente, line 31: do these options needs to be specific? I don't see that we run any of these checks,

.clang-tidy
```diff
@@ -0,0 +1,44 @@
+---
+Checks:          '-*,bugprone-*,cppcoreguidelines-*,modernize-*,concurrency-*'
+WarningsAsErrors: ''
+HeaderFilterRegex: ''
+AnalyzeTemporaryDtors: false
+FormatStyle:     file
+CheckOptions:
+  - key:             cert-dcl16-c.NewSuffixes
```
jeffdiclemente, line 8: this options doesn't seem necessary since the check isn't run?

jeffdiclemente (changes requested):

vivekkushwaha39 (commented):
Seems like some checks are failing 
image

jeffdiclemente:
> Seems like some checks are failing image

I'm assuming this is failing because of this repos access permissions for PRs from outside forked repos. See https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

.github/workflows/clang_tidy_diff_review.yml
```diff
@@ -0,0 +1,37 @@
...
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+
+    # Optionally generate compile_commands.json
+
+    - name: Checkout submodules
+      run: git submodule update --init --recursive
+
+    - uses: ZedThree/clang-tidy-review@v0.14.0
```
jeffdiclemente, line 20: I suspect the [split workflow](https://github.com/marketplace/actions/clang-tidy-review#usage-in-fork-environments-split-workflow) should be implemented to handle PRs from forked repos.

jeffdiclemente (changes requested):

.clang-tidy
AntiEXE: Yes, these aren't being used however these are the default values. I will remove these.

.clang-tidy
AntiEXE: We aren't using the cert-* checks too.

.clang-tidy
AntiEXE: AnalyzeTemporaryDtors is an artifact from an older clang-tidy version, notably 6 and below. Yeah it was dumped while dumping the config. 
FormatStyle: use the .clang-format file when its set to file. So, I guess that's okay to keep.

.github/workflows/clang_tidy_diff_review.yml
AntiEXE: I have applied the changes fort split_workflow.

.clang-tidy
AntiEXE: No we need not specify those.

.clang-tidy
AntiEXE: Updated it to false.

.clang-tidy
AntiEXE: We aren't using the cert checks.

.clang-tidy
AntiEXE: Removed it.

jeffdiclemente (approved):

PR #996, opened 2024-02-14
Add ability to have deterministic bundle builds on all platforms

achristoforides:


jeffdiclemente (changes requested):
there needs to be tests to make sure the timestamps are what is expected.

also, what about file ordering within the zip? I would think that if clients add more files to their meta data ZIP files, this will also be a problem for reproducible builds. See https://reproducible-builds.org/docs/archives/

tcormackMW:
In an effort to keep the changelog up to date, please add a new entry describing this change in the changelog in a

third_party/patches/miniz_3.0.2.patch
```diff
@@ -103,7 +111,41 @@
...
++
++    // If US_DETERMINISTIC_MINIZ_ZIP_FILES is defined, set the time and date fields of the central
++    // directory header to a consistent value.
++#ifdef US_DETERMINISTIC_MINIZ_ZIP_FILES
++    MZ_WRITE_LE16(pDst + MZ_ZIP_CDH_FILE_TIME_OFS, 0);
++    MZ_WRITE_LE16(pDst + MZ_ZIP_CDH_FILE_DATE_OFS, 0);
++#else
+     MZ_WRITE_LE16(pDst + MZ_ZIP_CDH_FILE_TIME_OFS, dos_time);
+     MZ_WRITE_LE16(pDst + MZ_ZIP_CDH_FILE_DATE_OFS, dos_date);
++#endif
```
jeffdiclemente, line 144: this is third-party code, we can't pollute it with our cppms specific #defines.

tools/rc/ResourceCompiler.cpp
```diff
@@ -795,6 +796,11 @@ checkSanity(option::Parser& parse, option::Option* options)
 int
 main(int argc, char** argv)
 {
+    // Deterministic build things to set
```
jeffdiclemente, line 799: I would prefer a comment that explains why the local is set in such a way for deterministic builds. This is not a helpful comment.

.github/workflows/build_and_test_nix.yml
```diff
@@ -96,3 +96,32 @@ jobs:
...
+    name: Build and Test [${{matrix.os}},Configuration=0,Deterministic]
+    runs-on: ${{matrix.os}}
+    env:
+      BUILD_CONFIGURATION: 0
+      BUILD_WITH_DETERMINISTIC: 1
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, macOS-11, macos-12]
```
jeffdiclemente, line 110: are two macOS versions necessary? can we pick one?

CMakeLists.txt
```diff
@@ -533,6 +534,10 @@ if(MSVC)
   endif()
 
   set(US_CXX_FLAGS "/MP /WX /D${msvc_version_define} ${US_CXX_FLAGS}")
+
+  # UTF-8 support
```
jeffdiclemente, line 538: It would help to explain why UTF-8 support is used in this comment.

CMakeLists.txt
```diff
@@ -756,6 +778,16 @@ if(APPLE AND CMAKE_GENERATOR MATCHES "Xcode")
   set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} ${US_LINK_FLAGS}")
 endif()
 
+#-----------------------------------------------------------------------------
+# Deterministic builds environment variables
+#-----------------------------------------------------------------------------
+if(US_USE_DETERMINISTIC_BUNDLE_BUILDS)
+  # On windows, setlocale() is used to set these variables in the ResourceCompiler
```
jeffdiclemente, line 785: an explanation of why the locale matter for deterministic bundles would be helpful.

third_party/miniz.c
```diff
@@ -6091,8 +6090,16 @@ static mz_bool mz_zip_writer_create_local_dir_header(mz_zip_archive *pZip, mz_ui
...
+
+    // If US_DETERMINISTIC_MINIZ_ZIP_FILES is defined, set the time and date fields of the local
+    // directory header to a consistent value.
+#ifdef US_DETERMINISTIC_MINIZ_ZIP_FILES
+    MZ_WRITE_LE16(pDst + MZ_ZIP_LDH_FILE_TIME_OFS, 0);
+    MZ_WRITE_LE16(pDst + MZ_ZIP_LDH_FILE_DATE_OFS, 0);
+#else
     MZ_WRITE_LE16(pDst + MZ_ZIP_LDH_FILE_TIME_OFS, dos_time);
     MZ_WRITE_LE16(pDst + MZ_ZIP_LDH_FILE_DATE_OFS, dos_date);
+#endif
```
jeffdiclemente, line 6102: this is third-party code, we can't pollute it with our cppms specific #defines.

third_party/miniz.c
```diff
@@ -6113,8 +6120,16 @@ static mz_bool mz_zip_writer_create_central_dir_header(mz_zip_archive *pZip, mz_
...
+
+    // If US_DETERMINISTIC_MINIZ_ZIP_FILES is defined, set the time and date fields of the central
+    // directory header to a consistent value.
+#ifdef US_DETERMINISTIC_MINIZ_ZIP_FILES
+    MZ_WRITE_LE16(pDst + MZ_ZIP_CDH_FILE_TIME_OFS, 0);
+    MZ_WRITE_LE16(pDst + MZ_ZIP_CDH_FILE_DATE_OFS, 0);
+#else
     MZ_WRITE_LE16(pDst + MZ_ZIP_CDH_FILE_TIME_OFS, dos_time);
     MZ_WRITE_LE16(pDst + MZ_ZIP_CDH_FILE_DATE_OFS, dos_date);
+#endif
```
jeffdiclemente, line 6132: this is third-party code, we can't pollute it with our cppms specific #defines.

jeffdiclemente (changes requested):

.github/workflows/build_and_test_nix.yml
carneyweb: Sure. macos-12?

jeffdiclemente: the latest OS version sounds reasonable to me. Take a look at the github action documentation, there may be a moniker you can use for the latest macOS, like there is for Windows (windows-latest), so we don't have to keep changing the value in this file every time a new macOS version is supported by github actions.

maci64_link.sh
```diff
@@ -0,0 +1,13 @@
+#! /usr/bin/env bash
```
jeffdiclemente, line 1: nit: can the file be named something more descriptive? e.g. maci64_deterministic_link.sh ?

tools/rc/test/basic_test.sh
```diff
@@ -0,0 +1,24 @@
+#! /usr/bin/env bash
```
jeffdiclemente, line 1: nit: can the file be called something more descriptive? e.g., basic_deterministic_test.sh ?

tools/rc/test/basic_test.bat
```diff
@@ -0,0 +1,39 @@
+:: This test is only invoked if the CMAKE flag US_USE_DETERMINISTIC_BUNDLE_BUILDS is set.
```
jeffdiclemente, line 1: nit: can the file be called something more descriptive? e.g., basic_deterministic_test.sh ?

jeffdiclemente (approved):

PR #999, opened 2024-02-22
Update code-analysis to run on windows-2022

tcormackMW:
Newest version (1.13 -- we are currently on 1.12) does not fix this issue. Therefore, we are just going to run our analysis on windows-2022 for now

### DIFF
--- a/compendium/LogServiceImpl/CMakeLists.txt
+++ b/compendium/LogServiceImpl/CMakeLists.txt
@@ -41,7 +41,6 @@ usMacroCreateBundle(LogServiceImpl
   SOURCES ${_srcs} src/Activator.cpp
   BINARY_RESOURCES manifest.json
   )
-set(_SECURE_SCL ON)
 
 target_include_directories(LogService PRIVATE
   ${CMAKE_CURRENT_SOURCE_DIR}/src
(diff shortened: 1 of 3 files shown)


jeffdiclemente:
@tcormackMW, it seems like switching to `windows-2022` doesn't solve the problem.

compendium/LogServiceImpl/CMakeLists.txt
```diff
@@ -22,6 +22,9 @@ endif()
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/resources/manifest.json.in
 	       ${CMAKE_CURRENT_BINARY_DIR}/resources/manifest.json)
+if(MSVC)
+  add_compile_definitions(_SILENCE_STDEXT_ARR_ITERS_DEPRECATION_WARNING)
```
jeffdiclemente, line 26: a short explanation of why this is necessary would help in the future. eventually we should be able to remove this and whomever is in a position to do that should know what is required to remove this compile macro definition.

jeffdiclemente (changes requested):

compendium/LogServiceImpl/CMakeLists.txt
tcormackMW: comment added

PR #1007, opened 2024-04-30
Fix RemoveFromBoundServicesCache to handle invalid service objects

tcormackMW:
If an invalid ServiceObject is returned from `bc.GetServiceObjects`, calling GetService will return an invalid service. If we see this, we should not continue to removeFromBoundServicesCache.


jeffdiclemente (approved):

Burgch (approved):
Thanks!

PR #1009, opened 2024-05-28
Update LogService class in CppMicroServices

architapruthi:
The current CppMicroServices LogService API and implementation that we ship with CppMicroServices does not confirm to the latest LogService OSGi spec (https://docs.osgi.org/specification/osgi.core/8.0.0/service.log.html).

Hence, to support latest features and classes, we are updating LogService class, which confirms to OSGi spec 8.

More information can be found out in the following RFC:
https://github.com/CppMicroServices/rfcs/pull/21

Signed-off-by: The MathWorks, Inc. 
architap@mathworks.com


compendium/LogService/include/cppmicroservices/logservice/Logger.hpp
```diff
@@ -0,0 +1,110 @@
...
+        class Logger
+        {
+          public:
+
+            virtual ~Logger() = default;
+
+	    virtual void audit(const std::string& message) = 0;
+	    virtual void audit(const std::string& format, const std::string& arg) = 0;
+	    virtual void audit(const std::string& format, const std::string& arg1, const std::string& arg2) = 0;
+	    virtual void audit(std::string const& message, const std::exception_ptr ex) = 0;
```
kkumar45, line 61: const in parameters are inconsistent across different interfaces. please follow one style of all interfaces

compendium/LogService/include/cppmicroservices/logservice/Logger.hpp
```diff
@@ -0,0 +1,110 @@
...
+            virtual void warn(const std::string& format, const std::string& arg1, const std::string& arg2) = 0;
+	    virtual void warn(std::string const& message, const std::exception_ptr ex) = 0;
+            virtual void warn(std::string const& message, ServiceReferenceBase const& sr) = 0;
+            virtual void warn(std::string const& message, ServiceReferenceBase const& sr, const std::exception_ptr ex) = 0;
+
+	     /**
+             * Registers a sink to the logger for introspection of contents. This is not a publicly available
+             * function and should only be used for testing. This is NOT thread-safe.
+             */
+	    virtual void AddSink(spdlog::sink_ptr& sink) = 0;
```
kkumar45, line 104: Typically, a logger interface would focus solely on logging actions, not on managing logger sinks. Does it violates SRP?

compendium/LogServiceImpl/src/LoggerImpl.cpp
```diff
@@ -0,0 +1,351 @@
...
+            m_Logger->set_level(spdlog::level::trace);
+            m_Logger->trace(format, arg1, arg2);
+        }
+
+        void
+        LoggerImpl::audit(std::string const& message, const std::exception_ptr ex)
+        {
+            m_Logger->set_level(spdlog::level::trace);
+            std::string full_message = message;
+            full_message = message + GetExceptionMessage(ex);
```
kkumar45, line 87: this can be simplifies as,
`m_Logger->trace(message + GetExceptionMessage(ex));`
please check it at other places

compendium/LogServiceImpl/src/LoggerImpl.cpp
```diff
@@ -0,0 +1,351 @@
...
+{
+    namespace logservice
+    {
+        std::string
+        GetExceptionMessage(std::exception_ptr const& ex)
+        {
+            std::string message = "\nException logged: ";
+            if (ex)
+            {
+                std::ostringstream stream;
```
kkumar45, line 18: where is `stream` used here?

compendium/LogService/include/cppmicroservices/logservice/Logger.hpp
architapruthi: This function registers a sink to the logger for introspection of contents. This is not a publicly available
 function and is only used for testing. 
We are using this function only for testing purpose.

compendium/LogServiceImpl/src/LogServiceImpl.cpp
```diff
@@ -6,93 +6,187 @@
...
         {
-            std::string message = "\nException logged: ";
-            if (ex)
+            logger = getLogger(loggerName);    
+	}
+
+        void
+        LogServiceImpl::Log(SeverityLevel level, std::string const& message)
+        {
+            switch (level)
```
carneyweb, line 21: Lots of repeated code below. Consider an early exit at the top of the function:

auto currLogger = std::atomic_load(&logger);
if (!currLogger) {
    return;
}

Then all of the cases in the switch become one line.

compendium/LogServiceImpl/src/LogServiceImpl.cpp
```diff
@@ -6,93 +6,187 @@
...
+		      currLogger->trace(message);
+		    }
+		    break;
+		}
+            }
         }
 
         void
-        LogServiceImpl::Log(SeverityLevel level, std::string const& message)
+        LogServiceImpl::Log(SeverityLevel level, std::string const& message, const std::exception_ptr ex)
```
carneyweb, line 77: See above. Same repeated code.

compendium/LogServiceImpl/src/LogServiceImpl.cpp
```diff
@@ -6,93 +6,187 @@
...
+		    {
+                        currLogger->trace(message, ex);
+		    }
+                    break;
+                }
+            }
         }
 
         void
         LogServiceImpl::Log(ServiceReferenceBase const& sr, SeverityLevel level, std::string const& message)
```
carneyweb, line 135: And again...

compendium/LogServiceImpl/src/LogServiceImpl.cpp
```diff
@@ -101,15 +195,78 @@ namespace cppmicroservices
                             std::string const& message,
                             const std::exception_ptr ex)
         {
-            std::string full_message = message;
-            full_message = message + GetServiceReferenceInfo(sr) + GetExceptionMessage(ex);
-            LogServiceImpl::Log(level, full_message);
+            switch (level)
```
carneyweb, line 198: And again...

Another possible DRY optimization: make all the overloaded calls to Log call this one implementation with whatever the default arguments are for the missing arguments.

carneyweb (changes requested):

compendium/LogServiceImpl/src/LogServiceImpl.cpp
architapruthi: Thanks for the feedback. For the second suggestion, calling one Log implementation with default arguments is resulting in failure of existing LogService tests since the value in the 'message' comes out to be different when default input values come into picture. (For example, if we pass nullptr as default value in exception pointer, 'GetExceptionMessage()' function returns 'none' and as a result, this gets added in the message)

jeffdiclemente (changes requested):

compendium/LogService/include/cppmicroservices/logservice/Logger.hpp
```diff
@@ -0,0 +1,121 @@
...
+         * \ingroup MicroService
+         * \ingroup gr_logservice
+         *
+         * Provides methods for bundles to write messages to the log.
+         * Logger interface defines several methods for each of the defined LogLevels.
+         *
+         */
+
+	enum class LogLevel
+        {
```
tcormackMW, line 36: Enums are typically denoted by uppercase letters, maybe should be here as well

compendium/LogServiceImpl/src/Activator.cpp
```diff
@@ -9,10 +10,14 @@ namespace cppmicroservices
...
-            {
-                auto svc
-                    = std::make_shared<cppmicroservices::logservice::LogServiceImpl>("cppmicroservices::logservice");
+            { 
+		cppmicroservices::Bundle bundle = bc.GetBundle();
+		const std::string bsn = bundle.GetSymbolicName();
+		const std::string logger_name = "LogService." + bsn;
+		auto svc
+                    = std::make_shared<cppmicroservices::logservice::LogServiceImpl>(logger_name);
                 bc.RegisterService<cppmicroservices::logservice::LogService>(std::move(svc));
```
tcormackMW, line 19: moving from a moved from object?

std::move(svc);
std::move(svc);

compendium/LogServiceImpl/src/LogServiceImpl.cpp
```diff
@@ -6,110 +6,192 @@
...
+
+           switch (level)
+           {
+               case SeverityLevel::LOG_DEBUG:
+               {
+		   currLogger->debug(message);
+	           break;
+	       }
+	       case SeverityLevel::LOG_INFO:
+	       {
```
tcormackMW, line 37: I believe you should be able to remove the '{' and '}' around each case, i.e.:

               case SeverityLevel::LOG_DEBUG:
		   currLogger->debug(message);
	           break;
	       case SeverityLevel::LOG_INFO:
	       .......
	      
I believe our clang settings support this syntax, though I could be wrong

compendium/LogServiceImpl/test/TestLogService.cpp
```diff
@@ -163,13 +163,6 @@ TEST_F(LogServiceImplTests, InvalidLoggerUsage)
 {
     auto logger = GetLogger();
 
-    ASSERT_NO_THROW(logger->Log(static_cast<ls::SeverityLevel>(-1), "Test invalid negative severity level"));
-    EXPECT_FALSE(ContainsRegex(log_preamble + "Test invalid negative severity level"));
-
-    ASSERT_NO_THROW(logger->Log(static_cast<ls::SeverityLevel>(std::numeric_limits<unsigned int>::max()),
-                                "Test invalid maximum severity level"));
-    EXPECT_FALSE(ContainsRegex(log_preamble + "Test invalid maximum severity level"));
-
```
jeffdiclemente, line 172: what was the reason for removing these test points

compendium/LogService/include/cppmicroservices/logservice/LogService.hpp
```diff
@@ -69,11 +70,24 @@ namespace cppmicroservices
          *
          * @remarks This class is thread safe.
          */
-        class LogService
+        class LogService : public LoggerFactory
         {
           public:
             virtual ~LogService() = default;
 
+	    LogService() = default;
```
jeffdiclemente, line 78: this class is abstract and cannot be instantiated. There should not be a default constructor.

compendium/LogService/include/cppmicroservices/logservice/Logger.hpp
```diff
@@ -0,0 +1,121 @@
...
+            trace //This log level is used for large volume of output for tracing operations.   
+	};
+	
+        class Logger
+        {
+          public:
+
+            virtual ~Logger() = default;
+
+	    Logger() = default;
```
jeffdiclemente, line 56: this class is abstract and cannot be instantiated. There should not be a default constructor.

compendium/LogService/include/cppmicroservices/logservice/LoggerFactory.hpp
```diff
@@ -0,0 +1,61 @@
...
+namespace cppmicroservices::logservice
+{
+	class LoggerFactory
+	{
+	    public:
+
+	        inline static const std::string ROOT_LOGGER_NAME = "ROOT";
+	
+                virtual ~LoggerFactory() = default;
+		LoggerFactory() = default;
```
jeffdiclemente, line 41: this class is abstract and cannot be instantiated. There should not be a default constructor.

jeffdiclemente (changes requested):

compendium/LogServiceImpl/src/LogServiceImpl.cpp
architapruthi: I added braces here since using braces consistently for each case enhances code clarity, prevents bugs, and makes code more maintainable. This style aligns with principles of core guidelines which prioritize code readability and maintainability.

compendium/LogService/include/cppmicroservices/logservice/Logger.hpp
architapruthi: I removed this altogether since we are not using this as of now in our code. Maybe in future, when we take LoggerContext (An interface that allows the configuration of effective logging levels for a Bundle) into picture, we can add this.

compendium/LogServiceImpl/test/TestLogService.cpp
architapruthi: These are removed since earlier there was no default level for logging and these values of log level were considered invalid. With new OSGi spec, the default logging level is 'trace' and hence the messages are logged at that level.

architapruthi: Modified test instead of removing. Thanks.

framework/src/util/CFRLogger.cpp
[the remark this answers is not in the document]
architapruthi: I don't see any existing test for CFRLogger. Even existing API's in this file are not covered by any tests. I think there can be a separate task for this. What are your thoughts?

compendium/LogServiceImpl/test/TestLogService.cpp
```diff
@@ -60,6 +60,8 @@ class LogServiceImplTests : public ::testing::Test
         std::string text = oss.str();
         std::smatch m;
         bool found = std::regex_search(text, m, std::regex(regex));
+        std::cout << "TEXT STRING =" << text << std::endl;
+        std::cout << "REGEX = " << regex << std::endl;
```
jeffdiclemente, line 64: are these lines still needed? seems like these were for debugging and can be removed?

jeffdiclemente (changes requested):
codecov is flagging a number of lines were not covered by tests. Are there more tests that need to be written?

framework/src/util/CFRLogger.cpp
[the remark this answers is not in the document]
jeffdiclemente: you may need to write some new tests. first look in the `FrameworkTests` tests to see if any already exist.

compendium/LogService/include/cppmicroservices/logservice/Logger.hpp
```diff
@@ -0,0 +1,343 @@
...
+#endif
+#ifdef INFO
+#undef INFO
+#endif
+#ifdef DEBUG
+#undef DEBUG
+#endif
+#ifdef TRACE
+#undef TRACE
+#endif
```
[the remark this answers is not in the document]
architapruthi: Looks like 'ERROR' has some conflicting macro defined in some Windows header, which is resulting in errors in our code if we don't undefine 'ERROR' . Rest of the #undef 's are not required, and I have removed them.

[Ideally, an enum class should scope its enumerators appropriately, preventing name conflicts. However, certain macros in the Windows headers can cause unexpected behavior when their names are reused, even within the scope of an enum class.]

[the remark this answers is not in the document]
kkumar45: Ideally, its good practice to avoid use of ALL_CAPS for enumerators. See - https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#enum5-dont-use-all_caps-for-enumerators

framework/src/util/CFRLogger.cpp
```diff
@@ -30,6 +29,16 @@ namespace cppmicroservices
     {
         CFRLogger::CFRLogger() : serviceTracker(), logService(nullptr) {}
 
+	CFRLogger::CFRLogger(cppmicroservices::BundleContext context)
+            : cfrContext(context)
+            , serviceTracker(
```
jeffdiclemente, line 34: If `CFRLogger::Open()` is called after this constructor runs, then `serviceTracker` is destroyed and re-created as a new service tracker. This is a redundant operation.

jeffdiclemente (changes requested):

framework/src/util/CFRLogger.cpp
architapruthi: I added this constructor since without this, the tests were failing (With error: Actual function call count doesn't match EXPECT_CALL's). I am guessing that since we were not opening service tracker in the constructor, the calls to Log in the LogService were not being made.

I have also tried calling default constructor of CFRLogger and then Open() method on the object, but I am getting same error in the test.

Please let me know if there can be some other way to avoid doing this and run the tests. Thanks.

jeffdiclemente: @architapruthi, which test is failing?

framework/src/util/CFRLogger.cpp
```diff
@@ -119,6 +128,20 @@ namespace cppmicroservices
             }
         }
 
+	std::shared_ptr<cppmicroservices::logservice::Logger>
+        CFRLogger::getLogger(const std::string& name) const
+	{
+	     auto currLogger = std::atomic_load(&logService);
```
jeffdiclemente, line 134: a crash will happen if `currLogger` is `nullptr`. This needs to be handled gracefully.

framework/src/util/CFRLogger.cpp
```diff
@@ -119,6 +128,20 @@ namespace cppmicroservices
...
+	     return currLogger->getLogger(name);
+	}
+
+	std::shared_ptr<cppmicroservices::logservice::Logger>
+        CFRLogger::getLogger(const cppmicroservices::Bundle& bundle, const std::string& name) const
+        {
+             auto currLogger = std::atomic_load(&logService);
```
jeffdiclemente, line 141: a crash will happen if `currLogger` is `nullptr`. This needs to be handled gracefully.

compendium/DeclarativeServices/src/SCRLogger.cpp
```diff
@@ -126,5 +126,20 @@ namespace cppmicroservices
                 currLogger->Log(sr, level, message, ex);
             }
         }
+
+	std::shared_ptr<cppmicroservices::logservice::Logger>
+        SCRLogger::getLogger(const std::string& name) const
+	{
+	    auto currLogger = std::atomic_load(&logService);
```
jeffdiclemente, line 133: a crash will happen if `currLogger` is `nullptr`. This needs to be handled gracefully.

compendium/DeclarativeServices/src/SCRLogger.cpp
```diff
@@ -126,5 +126,20 @@ namespace cppmicroservices
...
+	    return currLogger->getLogger(name);
+	}
+
+	std::shared_ptr<cppmicroservices::logservice::Logger>
+        SCRLogger::getLogger(const cppmicroservices::Bundle& bundle, const std::string& name) const
+        {
+            auto currLogger = std::atomic_load(&logService);
```
jeffdiclemente, line 140: a crash will happen if `currLogger` is `nullptr`. This needs to be handled gracefully.

jeffdiclemente (changes requested):

framework/src/util/CFRLogger.cpp
architapruthi: The tests are running fine with the new constructor.
I will see if I can achieve running the tests without adding that. Thanks.

architapruthi: When I use the default constructor and Open() call, the cfrContext object (bundle context) is null only and hence the service tracker is not starting to start tracking the service and call the Log methods.

Hence, I have kept the new constructor for now, for testing purposes.

architapruthi: Made the changes as discussed.

framework/src/util/CFRLogger.cpp
```diff
@@ -30,6 +29,16 @@ namespace cppmicroservices
     {
         CFRLogger::CFRLogger() : serviceTracker(), logService(nullptr) {}
 
+	CFRLogger::CFRLogger(cppmicroservices::BundleContext context)
+            : cfrContext(context)
```
BrianWeed, line 33: ‘context’ could be moved here

compendium/LogServiceImpl/src/LogServiceImpl.cpp
```diff
@@ -6,110 +6,193 @@
...
-            LogServiceImpl::Log(level, full_message);
+            std::shared_ptr<LoggerFactory> lf = std::make_shared<LoggerFactoryImpl>();
+            return lf->getLogger(bundle, name);
         }
 
         void
         LogServiceImpl::AddSink(spdlog::sink_ptr& sink)
         {
-            m_Logger->sinks().push_back(sink);
+	    std::shared_ptr<LoggerImpl> logimpl = std::dynamic_pointer_cast<LoggerImpl>(logger);
```
jeffdiclemente, line 195: does this instance of accessing `logger` need to follow the same pattern of calling `std::atomic_load(&logger);` first before accessing the instance?

jeffdiclemente (approved):

compendium/LogServiceImpl/src/LogServiceImpl.cpp
```diff
@@ -6,110 +6,198 @@
...
+           if (!currLogger)
+           {
+               return;
+           }
+
+           switch (level)
+           {
+               case SeverityLevel::LOG_DEBUG:
+               {
+		   currLogger->debug(message);
```
tcormackMW, line 33: I think having this switch statement 3 times in 3 functions is perhaps not ideal. Rather what if we just had a 'impl' of the `Log` call that looked like this:

```
template<typename... Args>
void LogServiceImpl::logImpl(SeverityLevel level, Args&&... args) {
    auto currLogger = std::atomic_load(&logger);
    if (!currLogger) {
        return;
    }

    switch (level) {
        case SeverityLevel::LOG_DEBUG:
            currLogger->debug(std::forward<Args>(args)...);
            break;
        case SeverityLevel::LOG_INFO:
            currLogger->info(std::forward<Args>(args)...);
            break;
        case SeverityLevel::LOG_WARNING:
            currLogger->warn(std::forward<Args>(args)...);
            break;
        case SeverityLevel::LOG_ERROR:
            currLogger->error(std::forward<Args>(args)...);
            break;
        default:
            currLogger->trace(std::forward<Args>(args)...);
            break;
    }
}
```

**Note**: This would need to be in hpp as it is templated

And then each of the other log calls could simply forward their arguments:

```
void LogServiceImpl::Log(SeverityLevel level, std::string const& message) {
    logImpl(level, message);
}

void LogServiceImpl::Log(SeverityLevel level, std::string const& message, const std::exception_ptr ex) {
    logImpl(level, message, ex);
}

void LogServiceImpl::Log(ServiceReferenceBase const& sr, SeverityLevel level, std::string const& message) {
	    logImpl(level, message, sr);
}
```

architapruthi: Thanks for the feedback.
However, this change would necessitate making all the log methods in the Logger class (e.g., debug, error, info, etc.) template functions that can accept variadic arguments. This would mean that these templates, along with their implementations, would reside in the service interface, resulting in a loss of the loose coupling between service consumers and the service implementation.

tcormackMW: Hi Archita, thanks for the response. 

I am not sure this would require that all the log methods take variadic arguments. 

I cloned your branch and imjplemented my suggestion such that the LogServiceImpl class made all its calls like this:

```
    void
    LogServiceImpl::Log(SeverityLevel level, std::string const& message, std::exception_ptr const ex)
    {
        logImpl(level, message, ex);
    }
```

and in LogServiceImpl.hpp:

```
            template <typename... Args>
            void
            logImpl(SeverityLevel level, Args&&... args)
            {
                auto currLogger = std::atomic_load(&logger);
                if (!currLogger)
                {
                    return;
                }

                switch (level)
                {
                    case SeverityLevel::LOG_DEBUG:
                        currLogger->debug(std::forward<Args>(args)...);
                        break;
                    case SeverityLevel::LOG_INFO:
                        currLogger->info(std::forward<Args>(args)...);
                        break;
                    case SeverityLevel::LOG_WARNING:
                        currLogger->warn(std::forward<Args>(args)...);
                        break;
                    case SeverityLevel::LOG_ERROR:
                        currLogger->error(std::forward<Args>(args)...);
                        break;
                    default:
                        currLogger->trace(std::forward<Args>(args)...);
                        break;
                }
            }
```
And this compiled and passed all tests without any changes to the Logger class functions. 

Let me know if I am missing something here, but I think this should work.

jeffdiclemente (approved):

jeffdiclemente:
@carneyweb, it looks like merging this PR is blocked by your review. Can you please re-review and either approve or provide feedback? Thanks.

carneyweb (approved):

carneyweb:
Sorry for the delay. Looks good.

- Mike.

PR #1011, opened 2024-06-11
Add test to verify support of nested `AnyMaps` in initializer lists

tcormackMW:


framework/test/gtest/AnyMapTest.cpp
```diff
@@ -442,4 +442,24 @@ TEST(AnyMapTest, InitializerList)
...
+    };
+    // EXPECT_EQ(any_map::UNORDERED_MAP_CASEINSENSITIVE_KEYS, unordered_any_nested.GetType());
+    // EXPECT_EQ(4, unordered_any_nested.size());
+    // EXPECT_EQ(Any(7), unordered_any_nested.at("j"));
+    // EXPECT_EQ(Any(8), unordered_any_nested.at("k"));
+    // EXPECT_EQ(std::string("B"), any_cast<std::string>(unordered_any_nested.at("l")));
+    // EXPECT_EQ(AnyMap(nested), any_cast<AnyMap>(unordered_any_nested.at("m")));
+    // EXPECT_EQ(Any(9), any_cast<AnyMap>(unordered_any_nested.at("m")).at("n"));
+    // EXPECT_EQ(Any(10), any_cast<AnyMap>(unordered_any_nested.at("m")).at("o"));
+    // EXPECT_EQ(std::string("C"), any_cast<std::string>(any_cast<AnyMap>(unordered_any_nested.at("m")).at("p")));
```
jeffdiclemente, line 464: why are these commented out?

tcormackMW: we're seeing failures on windows in all tests that are executed after this test suite... Mike and I are guessing the stack is being corrupted in this test somehow, so we are trying to pinpoint what is causing this corruption

PR #1013, opened 2024-06-19
Functionality to print anymap to a compilable c++ format

tcormackMW:


framework/include/cppmicroservices/Any.h
```diff
@@ -885,6 +1004,64 @@ namespace cppmicroservices
...
+        newline_and_indent(os, increment, indent - increment);
+        os << "}}";
+        return os;
+    }
+
+    template <class K, class V>
+    std::ostream&
+    any_value_to_cpp(std::ostream& os, std::map<K, V> const& m, uint8_t const increment, int32_t const indent)
+    {
+        os << "AnyMap {{";
```
carneyweb, line 1041: Needs to have map type in the constructor.

framework/src/util/AnyMap.cpp
```diff
@@ -1197,6 +1197,49 @@ namespace cppmicroservices
...
 
+    template <>
+    std::ostream&
+    any_value_to_cpp(std::ostream& os, AnyMap const& m, uint8_t const increment, int32_t const indent)
+    {
+        auto const mapType = m.GetType();
+        std::string typeStr;
+        switch (mapType)
+        {
+            case cppmicroservices::any_map::map_type::ORDERED_MAP:
```
carneyweb, line 1208: Should the "typeStr" values include the fully qualified name? I.e. "cppmicroservices::any_map::map_type::ORDERED_MAP" etc.? Or would I have "using" statements in the surrounding generated code to simplify this output?

Similarly, should the types "AnyMap" and "AnyVector" be fully qualified?

carneyweb (commented):
Fix the one missing map_type (if I read the code right), and we can chat about whether the types and enums should be fully qualified or not.

framework/src/util/AnyMap.cpp
tcormackMW: They certainly can be fully qualified if you'd like, however I figured for the sake of being concise, I figured we would use Using statements

carneyweb (approved):

PR #1015, opened 2024-06-21
Fix Unix Valgrind Suppression

tcormackMW:


valgrind_suppressions.txt
```diff
@@ -417,3 +417,14 @@
...
 
+{
+   <usDeclarativeServicesTestsSuppression2>
+   Memcheck:Leak
+   match-leak-kinds: possible
+   fun:calloc
+   fun:calloc
+   fun:allocate_dtv
+   fun:_dl_allocate_tls
+   fun:allocate_stack
```
jeffdiclemente, line 428: I'm concerned this is too generic and might mask the detection of real leaks. Can this suppression be a little more specific (e.g. by adding test point related stack frames to the suppressions)?

tcormackMW: we can, this is actually an extension on a suppression that already exists in this file, however I can update this one to be more specific:

{
   <usDeclarativeServicesTestsSuppression0>
   Memcheck:Leak
   match-leak-kinds: possible
   fun:calloc
   fun:allocate_dtv
   fun:_dl_allocate_tls
   fun:allocate_stack
   ...
}

jeffdiclemente (commented):
With these suppressions, there are still valgrind errors in the linux builds.

jeffdiclemente (approved):

PR #1018, opened 2024-07-01
Test Case for Race Condition in Config Notifier

tcormackMW:
This test case does expose the bug fixed in #1016 for the ConfigNotifier. Essentially if we have a race between unregistering a listener for a config and then notifying of a config change, it can get past the `AnyListenersForPid` call made in `ConfigurationListenerImpl::ConfigurationEvent` and into `NotifyAllListeners` with an empty map. All this means is that the check, already implemented in #1016, needs to done. 

Of note, this is a *very sporadic* occurence even in this test case, on the order of 1/500 runs. I am not sure how common this is in real life examples, I have seen it on a much more common level i.e. 1/3 runs in the wild.

### DIFF
--- a/compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
+++ b/compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
@@ -1125,7 +1125,7 @@ namespace cppmicroservices
 
             auto bundleT = std::thread([&compConfig]() { compConfig->Initialize(); });
             auto frameworkT = std::thread(
-                [configAdminService=configAdminService]()
+                [configAdminService = configAdminService]()
                 {
                     auto configuration = configAdminService->GetConfiguration("sample::config");
                     auto fut = configuration->UpdateIfDifferent(std::unordered_map<std::string, cppmicroservices::Any> {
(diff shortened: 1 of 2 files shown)


compendium/DeclarativeServices/test/gtest/TestComponentConfigurationImpl.cpp
```diff
@@ -1344,6 +1287,87 @@ namespace cppmicroservices
...
+                    });
+                    fut.second.wait();
+                    configuration->UpdateIfDifferent(std::unordered_map<std::string, cppmicroservices::Any> {
+                        {"foo", true}
+                    });
+                });
+            auto bundleT = std::thread([&compConfig]() { compConfig->Stop(); });
+            auto bundleT1 = std::thread([&compConfig1]() { compConfig1->Stop(); });
+            auto bundleT2 = std::thread([&compConfig2]() { compConfig2->Stop(); });
+            auto bundleT3 = std::thread([&compConfig3]() { compConfig3->Stop(); });
```
jeffdiclemente, line 1363: I recommend using `std::async` instead of the lower level `std::thread` class

jeffdiclemente (approved):

carneyweb (approved):

PR #1022, opened 2024-07-17
Finish running all queued async jobs

tcormackMW:


compendium/AsyncWorkService/include/cppmicroservices/asyncworkservice/AsyncWorkService.hpp
```diff
@@ -67,6 +67,18 @@ namespace cppmicroservices
...
              */
             virtual void post(std::packaged_task<void()>&& task) = 0;
+
+            /**
+             * Optional method to wait for all currently executing tasks to finish
+             *
+             * @note If the AsyncWorkService does not implement this function and override,
+             * the call will default to noop.
+             */
+            virtual void
```
jeffdiclemente, line 77: should this be optional? I would argue that for correct behavior of any implementation that wants to be an `AsyncWorkService`, a "wait" function that waits on all pending and executing tasks is required. what do you think?

compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
```diff
@@ -367,4 +366,189 @@ namespace test
         ASSERT_TRUE(service) << "GetService failed for CAInterface.";
     }
 
+    class Barrier
+    {
+      public:
+        Barrier(std::size_t count) : threshold(count), count(count), generation(0) {}
```
jeffdiclemente, line 372: this is duplicate code that already exists in compendium\DeclarativeServices\test\gtest\TestComponentConfigurationImpl.cpp. Please place this class in a general location to be used by all tests.

jeffdiclemente (changes requested):

compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
```diff
@@ -367,4 +366,189 @@ namespace test
...
+
+      private:
+        std::mutex mutex;
+        std::condition_variable cond;
+        std::size_t threshold;
+        std::size_t count;
+        std::size_t generation;
+    };
+
+    class sync : public cppmicroservices::async::AsyncWorkService
```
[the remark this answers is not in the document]
tcormackMW: `postWithSynchronization`?

compendium/AsyncWorkService/include/cppmicroservices/asyncworkservice/AsyncWorkService.hpp
```diff
@@ -67,6 +67,12 @@ namespace cppmicroservices
              * with the std::packaged_task<void()> in order to wait on the async task.
              */
             virtual void post(std::packaged_task<void()>&& task) = 0;
+
+            /**
+             * Method to wait for all tasks posted to the service to finish
```
Burgch, line 72: Is it worth adding a note here that implementing this is recommended to avoid odd behaviour if an implementation of this interface is unregistered whilst it has unfinished work posted to it?

Burgch (approved):
Thanks Toby, this looks good to me!

compendium/AsyncWorkService/include/cppmicroservices/asyncworkservice/AsyncWorkService.hpp
Burgch: Perhaps optional (but strongly recommended) until CppMicroServices 4.0, and then it becomes required, as it would be a breaking change for downstream code to react to, in order to build successfully?

tcormackMW: We have transitioned to required for the time being. We'll have to decide if this is the correct way forward

compendium/AsyncWorkService/include/cppmicroservices/asyncworkservice/AsyncWorkService.hpp
tcormackMW: right now this is a required method -- I think maybe this is worth discussing in CppMs Working group tomorrow

jeffdiclemente (approved):

Issue #1025, opened 2024-07-21
Bundle.FindResources is never able to find out the manifest.json inside of compiled shared library. 


al-sabr:
![image](https://github.com/user-attachments/assets/792296df-408c-4f3a-a2cf-ef07a1def227)

this file never gets found although it is inside of the libresources-plugind.so

![image](https://github.com/user-attachments/assets/c43c5ffc-d879-4829-9fee-77a5e1af41d2)

This line is not correct because IsValid is just checking the bundleId which is not a good indicator of it there is the real manifest.json file.

It should be replaced by

![image](https://github.com/user-attachments/assets/644e2d18-e50e-4ead-aa8d-dbcf4db9c112)

The **names** & **indices** are never satistied and filled.

![image](https://github.com/user-attachments/assets/e12e5bdf-f4ec-4d2e-8d01-a8485dc0753b)

jeffdiclemente:
The CMake commands look ok to me. Can you share your code that is using the cppmicroservices APIs to find the bundle resource files?

jeffdiclemente:
> Did you tried the project I underlined?

unfortunately no, I've haven't had a chance to look at it yet

PR #1026, opened 2024-07-22
Bundle resource bug fixes

al-sabr:
Due to this thread I corrected the code so that it works.

https://github.com/CppMicroServices/CppMicroServices/issues/1025


jeffdiclemente:
These changes appear to be trying to work around a problem with a malformed or incorrectly built bundle.

PR #1028, opened 2024-07-24
Draft:  Refactor Framework Tests to Use Mocks

andrew-mathworks:
Work-in-progress PR, opened to allow Github Actions runs.  Draft status will be removed once tests are passing.


cmake/usMacroCreateBundle.cmake
```diff
@@ -5,7 +5,7 @@ macro(usMacroCreateBundle _project_name)
 cmake_parse_arguments(${_project_name}
   "SKIP_EXAMPLES;SKIP_INIT"
   "VERSION;TARGET;SYMBOLIC_NAME;EMBED_RESOURCE_METHOD"
-  "DEPENDS;PRIVATE_INCLUDE_DIRS;LINK_LIBRARIES;SOURCES;PRIVATE_HEADERS;PUBLIC_HEADERS;RESOURCES;BINARY_RESOURCES"
+  "DEPENDS;PRIVATE_INCLUDE_DIRS;LINK_LIBRARIES;SOURCES;PRIVATE_HEADERS;PUBLIC_HEADERS;RESOURCES;BINARY_RESOURCES;BUILD_OBJ"
```
jeffdiclemente, line 8: Making symbols visible to the test code this way is not what we want to do. I believe this will expose any class/function/symbol tagged with `US_ABI_TEST` macro as a public symbol when we ship this library and that is not what we want. I think the best way is to split the cppms core library CMake target into two targets; one that builds all the object files and one that links all the object files into the final shipping shared library. The test bed will link against the object files and not the final shipping shared library.

framework/include/cppmicroservices/detail/BundleResourceBuffer.h
```diff
@@ -66,6 +66,15 @@ namespace cppmicroservices
...
             std::unique_ptr<BundleResourceBufferPrivate> d;
+
+            friend class BundleResourceBufferTest;
+            friend class BundleResourceBufferTest_ConstructionWithEmptyData_Test;
+            friend class BundleResourceBufferTest_UnderflowAndUflow_Test;
+            friend class BundleResourceBufferTest_Pbackfail_Test;
+            friend class BundleResourceBufferTest_Showmanyc_Test;
+            friend class BundleResourceBufferTest_SeekoffAndSeekpos_Test;
+            friend class BundleResourceBufferTest_NewlineConversion_Test;
+            friend class BundleResourceBufferTest_RemoveLastNewline_Test;
```
jeffdiclemente, line 77: I would like to avoid lots of `friend class` declarations when possible. From a design perspective, the shipping code shouldn't be modified for the sole purpose of accommodating tests. In this case the friend classes could be made unnecessary if the tests used the public APIs of the `std::streambuf` class to test the private methods.

jeffdiclemente (changes requested):

framework/src/bundle/BundleRegistry.cpp
```diff
@@ -140,6 +140,12 @@ namespace cppmicroservices
 
     std::vector<Bundle>
     BundleRegistry::Install(std::string const& location, BundlePrivate*, cppmicroservices::AnyMap const& bundleManifest)
+    {
+        return Install1(location, bundleManifest, nullptr);
```
tcormackMW, line 144: I'm not sure why there is this new sub call. 

Purely speaking as someone who frequents the install and start and stop calls across this code base, the presence of start, start0, start1... causes confusion when trying to understand code flow. So if we need this call, I would prefer it called something else

framework/test/gtest/FrameworkEventTest.cpp
```diff
@@ -145,3 +151,55 @@ TEST(FrameworkEventTest, testFrameworkEvents)
...
+
+    std::ostringstream buf;
+    std::string message = "hello world";
+    std::exception_ptr e = std::make_exception_ptr(std::runtime_error("placeholder"));
+    for (auto type : frameworkTypes)
+    {
+        FrameworkEvent evt(type, bundle, message, e);
+        buf << evt << "\n";
+    }
+    std::string goal = "STARTED\n hello world\n Bundle[id=1, loc=MockBundle, name=MockBundle, state=INSTALLED]\n Exception: placeholder\nERROR\n hello world\n Bundle[id=1, loc=MockBundle, name=MockBundle, state=INSTALLED]\n Exception: placeholder\nWARNING\n hello world\n Bundle[id=1, loc=MockBundle, name=MockBundle, state=INSTALLED]\n Exception: placeholder\nINFO\n hello world\n Bundle[id=1, loc=MockBundle, name=MockBundle, state=INSTALLED]\n Exception: placeholder\nSTOPPED\n hello world\n Bundle[id=1, loc=MockBundle, name=MockBundle, state=INSTALLED]\n Exception: placeholder\nSTOPPED_UPDATE\n hello world\n Bundle[id=1, loc=MockBundle, name=MockBundle, state=INSTALLED]\n Exception: placeholder\nWAIT_TIMEDOUT\n hello world\n Bundle[id=1, loc=MockBundle, name=MockBundle, state=INSTALLED]\n Exception: placeholder\n";
```
tcormackMW, line 203: maybe replace with a string literal using the 'R(' syntax for clarity's sake

https://en.cppreference.com/w/cpp/language/string_literal#:~:text=//%20OK%3A%20equivalent%20to%20%22%5Cn)%5C%5C%5Cna%5C%22%5C%22%5Cn%22%0AR%22a(%0A)%5C%0Aa%22%22%0A)a%22%3B

Cubified: Fixed, thanks!  That's a cool feature

framework/src/bundle/BundleRegistry.cpp
Cubified: Ended up renaming it to `BundleRegistry::InstallWithContainer`, is that ok?

The basic issue was that I need to pass in a mocked `BundleResourceContainer`, but the existing `Install` constructed it within the function.  I tried to use default arguments so as to avoid adding a whole new function, but something about setting the container to `nullptr` by default caused a segfault in the existing tests.  Not ideal I realize, but in order to mock bundles it was absolutely necessary to mock `BundleResourceContainer` somehow.

framework/include/cppmicroservices/detail/BundleResourceBuffer.h
Cubified: Removed and replaced with a single `friend class BundleResourceBufferTestSuite`.  Would it be better if I removed that as well?

cmake/usMacroCreateBundle.cmake
Cubified: You're right sorry, I think a commit got lost on my way out since I got lazy with naming my stashes...  Just recreated it, let me know if it's satisfactory.

The way I ended up approaching it was to add a second framework target that defines `US_ABI_TESTING`, then linking only the tests against it.  That way the final `libCppMicroServicesd.so` isn't affected at all, but none of the linking configuration has to change.  I tried the object file approach for a long time, but ultimately it broke too many things in weird ways due to the global bundle context for `system_bundle` (especially on Windows, which allows symbols with duplicate names in different places to refer to different things).  It was much simpler to just redefine the macros responsible for exporting symbols and produce two separate targets.

PR #1029, opened 2024-08-08
Allow for r-value and pure std::shared_ptr reference declaration for DS services

tcormackMW:
Constructor injection of DS service references requires those references to be passed to the constructor by `const std::shared_ptr&` 

This adds the ability to pass by `std::shared_ptr` (by value) and by `std::shared_ptr const&&` (rvalue reference).


carneyweb (approved):
LGTM.

PR #1032, opened 2024-08-14
Performance improvement

BrianWeed:
By-value sink params can be moved into their final destination


tcormackMW (approved):
Looks good to me. Question is whether we should be calling move at the call site of all these functions. That likely requires more code analysis as we would need to make sure those objects are not used again, but potential for more performance boost

PR #1036, opened 2024-08-26
The loop parse the first folder of the zip file and ignores the other folders.

al-sabr:
The identified problem arises at line **(352 in original file)** which triggers a catch at line **(393 in original file)**.

The loop parse the first folder of the zip file and directly goes into catching the exception at line **(393 in original file)** by ignoring the other folders in the zip file which is not correct. By adding the additional catch statements the other folders are parsed and an error is shown. If during the parsing a folder in the zip file contains the manifest file the processing continues.

Example of the error messages:

```
Error occurred: bundle.symbolic_name is not defined in the bundle manifest for bundle assets (location=/mnt/dev/c++/bismika/build/plugins/libresources_plugin.so).
Error occurred: bundle.symbolic_name is not defined in the bundle manifest for bundle src (location=/mnt/dev/c++/bismika/build/plugins/libresources_plugin.so).
Error occurred: bundle.symbolic_name is not defined in the bundle manifest for bundle style (location=/mnt/dev/c++/bismika/build/plugins/libresources_plugin.so).
```

in this case the embedded zip file has a **resources_plugin** folder containing the **manifest.json**.

To see how the bundle with the embedded files is handled check the **bundles/resources_plugin** sub-project.

### DIFF
--- a/framework/src/bundle/BundleRegistry.cpp
+++ b/framework/src/bundle/BundleRegistry.cpp
@@ -387,7 +387,7 @@ namespace cppmicroservices
         catch (...)
         {
             // catch any other errors (that we have no information about)
-            std::cerr << "Unknown failure occurred. Possible memory corruption" << std::endl;
+            std::cerr << "Failed to install bundle library at " + location + ": " + util::GetLastExceptionStr() << std::endl;
 
             for (auto& ba : barchives)
             {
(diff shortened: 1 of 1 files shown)


jeffdiclemente:
> Any feedback ?

The issue described in this PR appears to stem from the same root cause as described in PR #1037 , namely, the bundle resource ZIP file format is not being followed. If bundle resources need to be accessed using the CppMicroServices bundle resource APIs, those resources need to be structured in the ZIP file appropriately (put into a "bundle symbolic name" folder). If the resources in the ZIP file are not supposed to be associated with a bundle and it doesn't make sense for the resources to be part of any bundle, then using the CppMicroServices bundle resource API is not the correct API to use and another mechanism other than CppMicroServices should be used to access resources.

PR #1037, opened 2024-08-26
Merged zip files structure not taken into consideration when fetching a resource from a bundle.

al-sabr:
Added a parameter to bypass the prefix which is causing problem for merged zip files with the main res-0.zip resources file. The usResourceCompiler3 creates a default folder named after the symbolic.name property and adds the manifest in it. Independent zip files can be merged with the res-0.zip file but those are not aware of symbolic.name property causing the call to Bundle.GetResource function to never find the merged files with res-0.zip file.

### DIFF
--- a/framework/include/cppmicroservices/Bundle.h
+++ b/framework/include/cppmicroservices/Bundle.h
@@ -569,7 +569,7 @@ namespace cppmicroservices
          *
          * @throws std::invalid_argument if this bundle is not initialized.
          */
-        BundleResource GetResource(std::string const& path) const;
+        BundleResource GetResource(std::string const& path, bool bypassPrefix = false) const;
 
         /**
          * Returns resources in this bundle.
--- a/framework/include/cppmicroservices/BundleResource.h
+++ b/framework/include/cppmicroservices/BundleResource.h
@@ -299,7 +299,7 @@ namespace cppmicroservices
         uint32_t GetCrc32() const;
 
       private:
-        BundleResource(std::string const& file, std::shared_ptr<BundleArchive const> const& archive);
+        BundleResource(std::string const& file, std::shared_ptr<BundleArchive const> const& archive, bool bypassPrefix = false);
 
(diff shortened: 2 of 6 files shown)


jeffdiclemente:
> > all resources contained in a bundle's zip file need to be in a <symbolic.name> subfolder. The bundle resource system is not designed to look for resources that live outside of the <symblic.name> zip folder. Resources are always associated with a bundle, so they must be within a <symbolic.name> subfolder.
> > Unfortunately, the cppmicroservices documentation isn't explicit about the design of [the bundle resource system](https://docs.cppmicroservices.org/en/stable/framework/doc/resources.html), however this (i.e. all resources need to be in a <symbolic.name> subfolder) is a design choice of the bundle resource system.
> > Can you construct the Independent zip files by running the usResourceCompiler specifying the bundle name (i.e. --bundle-name)? That should solve this specific problem without the need for code changes.
> 
> That design choice is broken because you cannot know what other merged zip files contains. The code has to be adapted so that other cases are not disabled.
> 
> In this case I already in the **bismika** project calling for the bundle's context url with **dynamic://resources_plugin/index.html** ... it doesn't make sense to make it **dynamic://resources_plugin/resources_plugin/index.html**
> 
> See : [#1036 (comment)](https://github.com/CppMicroServices/CppMicroServices/pull/1036#issuecomment-2309275683)
> 
> The project is an interesting case which a lot of things are broken in CppMicroServices.

I respectfully disagree with the design choice being "broken". "Broken" implies it doesn't work as designed, it does. We follow as best we can the OSGi spec and while the spec doesn't explicitly say that you have to place all bundle meta-data in a <bundle symbolic name> sub-directory in a zip file, the spec does imply that all bundle meta-data is associated with a bundle. This is how we've chosen to conform to this requirement. From looking at your various pull requests it looks like all the issues you have encountered is because the resources in the zip are not properly put into a <bundle symbolic name> subfolder in the ZIP file. I have looked at the bismika project and I don't see a technical reason why the zip file cannot be put in a <bundle symbolic name> subfolder.

PR #1040, opened 2024-09-12
Remove deadlock in config notification

tcormackMW:
Removes the 'ordering lock' that was doing little other than causing Deadlocks

Before change, the added test case causes deadlock, after change it does not


compendium/DeclarativeServices/test/TestUtils.cpp
```diff
@@ -275,4 +275,43 @@ namespace test
...
+    }
+
+    void
+    AsyncWorkServiceThreadPool::post(std::packaged_task<void()>&& task)
+    {
+        using Sig = void();
+        using Result = boost::asio::async_result<decltype(task), Sig>;
+        using Handler = typename Result::completion_handler_type;
+
+        Handler handler{std::forward<decltype(task)>(task)};
```
Burgch, line 311: `std::forward` makes sense only for forwarding references (i.e. r-value reference to a templated type). In this case, the type isn't a template and you have a regular r-value, so you can just use `std::move` ?

compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
```diff
@@ -471,4 +421,50 @@ namespace test
...
+        ::test::InstallAndStartConfigAdmin(context);
+
+        // Get a service reference to ConfigAdmin.
+        auto sr1 = context.GetServiceReference<cppmicroservices::service::cm::ConfigurationAdmin>();
+        auto configAdminService = context.GetService<cppmicroservices::service::cm::ConfigurationAdmin>(sr1);
+        ASSERT_TRUE(configAdminService) << "GetService failed for ConfigurationAdmin";
+
+        auto configObject = configAdminService->GetConfiguration(componentName);
+
+        cppmicroservices::AnyMap props(cppmicroservices::AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS);
```
Burgch, line 451: Could use `configObject->GetProperties()` to simplify this?

compendium/test_bundles/TestBundleDSCA10/src/ServiceImpl.cpp
```diff
@@ -0,0 +1,37 @@
...
+    void
+    ServiceComponentCA10::Modified(std::shared_ptr<ComponentContext> const& /*context*/,
+                                   std::shared_ptr<cppmicroservices::AnyMap> const& configuration)
+    {
+        std::lock_guard<std::mutex> lock(propertiesLock);
+        properties = configuration;
+        auto CA = cppmicroservices::any_cast<std::shared_ptr<cppmicroservices::service::cm::ConfigurationAdmin>>(
+            properties->at("CA"));
+
+        auto configObject = CA->GetConfiguration(cppmicroservices::any_cast<std::string>(properties->at("configID")));
```
Burgch, line 22: Seems a funny way to get hold of `ConfigurationAdmin`, why not use a reference & constructor injection?

Burgch (changes requested):
Looks great, thanks Toby!

compendium/test_bundles/TestBundleDSCA10/src/ServiceImpl.cpp
tcormackMW: Indeed.

compendium/DeclarativeServices/test/TestUtils.cpp
tcormackMW: I can fix this for sure. This was just moved from the testAsyncWorkService file so I hadn't thought too much about it.

compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
tcormackMW: Probably :)

compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
```diff
@@ -471,4 +421,54 @@ namespace test
...
+    TEST_F(TestAsyncWorkServiceEndToEnd, testUpdateConfigFromWithinModifiedCallback)
+    {
+        auto const& param = std::make_shared<AsyncWorkServiceThreadPool>(10);
+        auto context = framework.GetBundleContext();
+
+        // ASYNCWORKSERVICE
+        auto reg = context.RegisterService<cppmicroservices::async::AsyncWorkService>(param);
+        // auto sr = context.GetServiceReference<cppmicroservices::async::AsyncWorkService>();
+        // auto asyncWorkService = context.GetService<cppmicroservices::async::AsyncWorkService>(sr);
+        // US_UNUSED(asyncWorkService);
```
jeffdiclemente, line 437: delete this if its not needed.

jeffdiclemente (approved):
LGTM, once the other comments are addressed.

PR #1044, opened 2024-10-04
Functionality to retrieve Service Reference from Service

tcormackMW:


framework/src/service/ServiceReferenceBasePrivate.h
```diff
@@ -124,6 +130,153 @@ namespace cppmicroservices
...
+     *        in BundleContext::GetService method. The aliasing constructor helps automate
+     *        the call to UngetService method.
+     *
+     *        Service consumers can simply call GetService to obtain a shared_ptr to the
+     *        service object and not worry about calling UngetService when they are done.
+     *        The UngetService is called when all instances of the returned shared_ptr object
+     *        go out of scope.
+     */
+    template <class S>
+    struct ServiceHolder
```
jeffdiclemente, line 144: `ServiceHolder` and `UngetHelper` enable the same behavior for two different use cases (singleton service vs. bundle/prototype service). Now that these classes are in the same file, does it make sense to combine these two classes into one class that handles both use cases?

Combining these classes isn't required and I am fine with doing this at a later time if its possible and desirable.

jeffdiclemente (changes requested):
please add a RFC for this to the rfcs repo. Thanks!

framework/src/service/ServiceReferenceBasePrivate.h
```diff
@@ -124,6 +130,153 @@ namespace cppmicroservices
...
+    class CustomServiceDeleter
+    {
+      public:
+        CustomServiceDeleter(ServiceHolder<void>* sh) : sh_(sh), uh_(nullptr) {}
+        CustomServiceDeleter(UngetHelper* uh) : sh_(nullptr), uh_(uh) {}
+
+        void
+        operator()(ServiceHolder<void>* sh)
+        {
+            delete sh;
```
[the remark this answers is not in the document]
carneyweb: However, it's ALWAYS safe to delete a nullptr in c++: https://stackoverflow.com/questions/4190703/is-it-safe-to-delete-a-null-pointer

framework/src/service/ServiceReferenceBasePrivate.h
```diff
@@ -124,6 +130,153 @@ namespace cppmicroservices
...
+        }
+
+        void
+        operator()(UngetHelper* uh)
+        {
+            delete uh;
```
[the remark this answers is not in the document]
carneyweb: it's ALWAYS safe to delete a nullptr in c++: https://stackoverflow.com/questions/4190703/is-it-safe-to-delete-a-null-pointer

The delete operator does the nullptr check itself. Adding a check to the code just clutters things needlessly.

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -479,4 +479,15 @@ namespace cppmicroservices
...
         return false;
     }
+
+    ServiceReferenceU
+    ServiceReferenceFromService(std::shared_ptr<void> const& s)
+    {
+        auto deleter = std::get_deleter<CustomServiceDeleter>(s);
+        if (!deleter)
+        {
+            throw std::runtime_error("Not a managed shared_ptr");
```
Burgch, line 489: Could add test coverage of this by confirming that this throws if you pass in a `shared_ptr` you've created yourself in the test?

Burgch (commented):
Looking good, thanks Toby! +1 to Jeff's comments too :)

framework/src/service/ServiceReferenceBasePrivate.h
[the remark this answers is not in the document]
tcormackMW: Also, given that we are the one creating the ServiceHolder, and only the deleter for the actual object is invoked (ie ServiceHolder when it is a shared_ptr<ServiceHolder> and UngetHelper when it is a shared_ptr<UngetHelper), that pointer will never be null... right?

framework/src/service/ServiceReferenceBasePrivate.h
```diff
@@ -124,6 +130,153 @@ namespace cppmicroservices
...
+        {
+            if (sh_)
+            {
+                return sh_->sref;
+            }
+            return uh_->sref;
+        }
+
+      private:
+        ServiceHolder<void> const* const sh_;
```
carneyweb, line 277: This is just a matter of coding style and personal preference... but I personally really dislike naming member variables with the trailing underscore. I find that it makes reading code like "ah_->do_something()" with the "_->" awkward to read. If we really need to name a member variable in a special way to mark it as such, I prefer using the "m_ah" form. The code doesn't look strange and the "m_" is very visible... more so than a trailing "_".

For what it's worth :-)

carneyweb (approved):

Burgch (approved):
LGTM, thanks Toby!

Burgch (approved):
LGTM 😃

compendium/DeclarativeServices/test/gtest/TestServiceReferenceRetrieval.cpp
```diff
@@ -0,0 +1,91 @@
...
+            return id;
+        }
+
+      private:
+        int id;
+    };
+    /*
+     * Tests that a user is able to retrieve a serviceReference from a serviceObject from the coreFramework or DS
+     */
+    TEST_F(tServiceComponent, testServiceReferenceRetrievalForDSAndManReg)
```
jeffdiclemente, line 61: this is a good set of testpoints. I think there is one case missing; what happens if you:
1. register a service, say `foo`
2. get a service ref to `foo`
3. get an instance of the service object `foo`
4. call `cppmicroservices::ServiceReferenceFromService` passing in the `foo` from step 3
5. unregister `foo` with the cppms framework

What is the expected state of the service ref returned by `cppmicroservices::ServiceReferenceFromService`? I would expect it to act the same as any other service ref that had its service unregistered with the cppms framework.

jeffdiclemente (approved):
I think there is one testpoint missing that would be useful to lockdown the behavior of a `ServiceReference`. Other than that, LGTM and if you agree with me, I don't need to re-review after adding another testpoint.

PR #1046, opened 2024-10-28
Allow users to force checking of `Activate`, `Deactivate`, and `Modified` methods at compile time

tcormackMW:


compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
```diff
@@ -448,7 +448,10 @@ namespace cppmicroservices
                                       "and/or the service implementation does not implement all of the "
                                       "service interface's methods. A default constructor or a constructor "
                                       "with an AnyMap input parameter for the configuration properties is "
-                                      "required when inject-references is false. ");
+                                      "required when inject-references is false. This error can also be "
+                                      "triggered when inheriting from "
+                                      "cppmicroservices::service::component::usrDefinedMethodAssertion "
+                                      "without properly defining the Modified, Activate, and Deactivate methods");
```
bwestcot, line 454: I would personally work on this error message; I think having very clear errors is important to usability.

The first part of this message is written in terms of a solution ("a default constructor with ... is required"), but the new part is written in terms of the problem, and is self-referential ("This error can also be triggered when"). Changing perspective partway through the message is a bit confusing. The phrasing also suggests that the second part describes a cause that has not been mentioned before, even though the first sentence, as I understand it, describes the general case for this situation. I had to read it a couple times to make this connection.

(As someone who has not encountered the conditions that would cause this error) I also wonder if there is something special about usrDefinedMethodAssertion, or if this could be generalized to remind the user that they may need to implement additional methods if they are inheriting from a class with virtual members.

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/usrDefinedMethodAssertion.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+        virtual void Modified(std::shared_ptr<ComponentContext> const& context,
+                              std::shared_ptr<cppmicroservices::AnyMap> const& configuration)
+            = 0;
+        virtual void Activate(std::shared_ptr<ComponentContext> const& context) = 0;
+        virtual void Deactivate(std::shared_ptr<ComponentContext> const& context) = 0;
+    };
+} // namespace cppmicroservices::service::component
+
+#endif
```
bwestcot, line 43: This file is missing a newline at the end.

compendium/ServiceComponent/test/TestCompInst_IncorrectModifiedMethod.cpp
```diff
@@ -0,0 +1,73 @@
...
+        ComponentInstanceImpl<ServiceComponentWrongModifiedSig> compInstance; // compile error
+    }
+} // namespace
+
+int
+main(int argc, char** argv)
+{
+    ::testing::InitGoogleTest(&argc, argv);
+    return RUN_ALL_TESTS();
+}
```
bwestcot, line 73: This file is missing a newline at the end.

tcormackMW:
> I see that you added some tests, but based on the comments, they all seem to be negative tests. Do you have any test cases for when these checks are expected to pass?

@bwestcot The changes to TestBindingPolicy have non-negative tests

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp
tcormackMW: I agree. I updated this

bwestcot (approved):
Looks good to me.

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/usrDefinedMethodAssertion.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#ifndef usrDefinedMethodAssertion_hpp
+#define usrDefinedMethodAssertion_hpp
+
+#include "cppmicroservices/servicecomponent/ComponentContext.hpp"
+
+namespace cppmicroservices::service::component
+{
+    template <typename T>
+    struct usrDefinedMethodAssertion
```
carneyweb, line 31: I hate to be pedantic, but I believe that there's a "coding standard" at MathWorks that says that class names should begin with an upper case letter. Also, the "abbreviation" of 1 letter, usr instead of User... is it worth it to be possibly less clear to save one character?

compendium/ServiceComponent/test/TestCompInst_IncorrectActivateMethod.cpp
```diff
@@ -0,0 +1,74 @@
...
+using cppmicroservices::service::component::detail::ComponentInstanceImpl;
+using ComponentContext = cppmicroservices::service::component::ComponentContext;
+
+namespace
+{
+    struct TestServiceInterface1
+    {
+    };
+
+    class ServiceComponentWrongActivateSig final
```
carneyweb, line 39: Would names for these base classes be more readable and intuitive if they were something like: ServiceComponentCheckActivateSignature??

carneyweb (approved):
Looks good to me.

compendium/DeclarativeServices/test/gtest/TestBindingPolicies.cpp
```diff
@@ -265,6 +265,14 @@ namespace cppmicroservices
                                                false,
                                                MakeInterfaceMap<test::Interface1>(std::make_shared<test::InterfaceImpl>(
                                                    "ServiceComponentDynamicReluctantMandatoryUnary Interface1")) },
+                            DynamicRefPolicy { "TestBundleDSDRMU_wrongSig",
```
jeffdiclemente, line 268: Is this test component testing a failure mode? I would think it would from its name (i.e. "wrongSig"). Looking at this test component, I don't see what is "wrong" with it.

compendium/test_bundles/TestBundleDSDRMU_wrongSig/src/ServiceComponentWrongBindSig.cpp
jeffdiclemente, line 1: What is "wrong" about this test component?

compendium/ServiceComponent/test/CMakeLists.txt
```diff
@@ -150,3 +153,12 @@ set_tests_properties(usServiceComponentInvalidInheritanceTest PROPERTIES PASS_RE
...
 set_tests_properties(usServiceComponentNoDefaultCtorTest PROPERTIES PASS_REGULAR_EXPRESSION ".*An appropriate constructor was not found and/or the service implementation does not implement all of the service interface's methods. A constructor with service reference input parameters or a constructor with an AnyMap input parameter for configuration properties and service reference input parameters is required when inject-references is true.*")
 set_tests_properties(usServiceComponentRefCountTest PROPERTIES PASS_REGULAR_EXPRESSION ".*An appropriate constructor was not found and/or the service implementation does not implement all of the service interface's methods. A constructor with service reference input parameters or a constructor with an AnyMap input parameter for configuration properties and service reference input parameters is required when inject-references is true.*")
+if (US_COMPILER_MSVC)
```
jeffdiclemente, line 156: instead of tightly coupling the test to a compiler version/vendor, can custom strings be added to the `static_asserts` in [UserDefinedMethodAssertion.hpp](https://github.com/CppMicroServices/CppMicroServices/pull/1046/files#diff-801b07fb2f80837cf5fbea83fb8771f044d793cbbc14a3df09493592c1b8bb0e) and those checked instead of the compiler's? This would make the test not brittle to compiler upgrades/changes.

jeffdiclemente (changes requested):

compendium/ServiceComponent/test/CMakeLists.txt
tcormackMW: Struggling to make this work as i don't believe the static assert is actually what ends up throwing. What ends up throwing is that attempt to call a modified method which does not exist and therefore is beholden to the compiler warning for that:

```19: Z:\compendium\ServiceComponent\include\cppmicroservices/servicecomponent/UserDefinedMethodAssertion.hpp(46,55): error C2039: 'Modified': is not a member of '`anonymous-namespace'::ServiceComponentWrongModifiedSig' [Z:\build\config_0\compendium\ServiceComponent\test\usServiceComponentModifiedMethod.vcxproj]
19: Z:\compendium\ServiceComponent\test\TestCompInst_IncorrectModifiedMethod.cpp(43): message : see declaration of '`anonymous-namespace'::ServiceComponentWrongModifiedSig' [Z:\build\config_0\compendium\ServiceComponent\test\usServiceComponentModifiedMethod.vcxproj]
19: Z:\compendium\ServiceComponent\include\cppmicroservices/servicecomponent/UserDefinedMethodAssertion.hpp(38): message : while compiling class template member function 'cppmicroservices::service::component::UserDefinedMethodAssertion<`anonymous-namespace'::ServiceComponentWrongModifiedSig>::~UserDefinedMethodAssertion(void)' [Z:\build\config_0\compendium\ServiceComponent\test\usServiceComponentModifiedMethod.vcxproj]
19: Z:\compendium\ServiceComponent\test\TestCompInst_IncorrectModifiedMethod.cpp(46): message : see reference to class template instantiation 'cppmicroservices::service::component::UserDefinedMethodAssertion<`anonymous-namespace'::ServiceComponentWrongModifiedSig>' being compiled [Z:\build\config_0\compendium\ServiceComponent\test\usServiceComponentModifiedMethod.vcxproj]
19: Z:\compendium\ServiceComponent\include\cppmicroservices/servicecomponent/UserDefinedMethodAssertion.hpp(47): error C2062: type 'unknown-type' unexpected [Z:\build\config_0\compendium\ServiceComponent\test\usServiceComponentModifiedMethod.vcxproj]```

Looking into how we can make this work...

Comment (jeffdiclemente) [reply compendium/ServiceComponent/test/CMakeLists.txt:156]: Using https://cppinsights.io/ may help  you get some insight into what the instantiated template code is that the compiler actually sees and tries to compile.

File: compendium/ServiceComponent/include/cppmicroservices/servicecomponent/UserDefinedMethodAssertion.hpp [hunk 8]
```diff
@@ -0,0 +1,86 @@
...
+
+    template <typename T>
+    UserDefinedMethodAssertion<T>::~UserDefinedMethodAssertion()
+    {
+        static_assert(std::is_base_of_v<UserDefinedMethodAssertion<T>, T>,
+                      "T must derive from UserDefinedMethodAssertion<T>");
+
+        static_assert(has_activate_method<T>::value, "Error: An appropriate Activate method was not found.");
+        static_assert(has_deactivate_method<T>::value, "Error: An appropriate Deactivate method was not found.");
+        static_assert(has_modified_method<T>::value, "Error: An appropriate Modified method was not found.");
```
jeffdiclemente, line 83: It would be very helpful to say what the appropriate method signature is in the error message. That way, users don't have to go looking through documentation for it.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #1048, opened 2024-11-13
Allow for global namespacing in interface definitions and manifests

tcormackMW:


compendium/DeclarativeServices/src/metadata/MetadataParserImpl.cpp
```diff
@@ -38,6 +38,12 @@ namespace cppmicroservices
         namespace metadata
         {
 
+            std::string
+            removeLeadingNamespacing(std::string const& className)
+            {
+                return className.substr(className.find_first_not_of(':'));
```
Burgch, line 44: I don't think calling `std::string::substr` with `std::string::npos` (if there aren't any leading colons) is defined behaviour? It could potentially throw?

framework/src/service/ServiceRegistry.cpp
```diff
@@ -36,6 +36,12 @@
 namespace cppmicroservices
 {
 
+    std::string
+    removeLeadingNamespacing(std::string const& className)
+    {
+        return className.substr(className.find_first_not_of(':'));
```
Burgch, line 42: Ditto?

Burgch (approved):
Looks good, thanks Toby!

compendium/DeclarativeServices/src/metadata/MetadataParserImpl.cpp
tcormackMW: I am not sure I understand what situation you think is dangerous, but it is `find_first_not_of` so it isn't an issue if there are no leading `::` as it will just resolve to the 0th letter. 

What is dangerous, however, is if the classname is `""` or `":::::"`, in which case this will throw. So I need to fix those edge cases. 

Can you clarify if there is another situation that you see as potentially problematic?

framework/src/service/ServiceRegistry.cpp
tcormackMW: see above

compendium/DeclarativeServices/src/metadata/MetadataParserImpl.cpp
Burgch: Sorry, yep, right you are! So handle the npos before calling substr and return the string as it was passed in?

tcormackMW: Indeed!

jeffdiclemente (approved):

compendium/DeclarativeServices/test/gtest/GlobalServiceTest.cpp
```diff
@@ -0,0 +1,61 @@
...
+    };
+
+    struct implTestService : IntTestService
+    {
+    };
+
+    /**
+     * Verify a component with an interface with :: leading can be registered and found
+     */
+    TEST(TestGlobalNamespacing, testServiceWithGlobalNamespaceNoDS)
```
jeffdiclemente, line 21: is this testpoint needed if `TEST_F(ServiceRegistryTest, TestGlobalNamespaceOnServiceRegistration)` exists? If yes, can this testpoint be put in the core framework test suite instead? I don't see the necessity to have a test that doesn't have to do with DS in the DS test suite.

jeffdiclemente (approved):

PR #1049, opened 2024-11-27
moved google benchmark to v1.9.0

jeffdiclemente:
qualification: locally built on Windows 11 w/ VS2022


tcormackMW (changes requested):
Looks fine. But we need to make sure to update the third_party/README for the benchmark version and commit level right?

jeffdiclemente:
> Looks fine. But we need to make sure to update the third_party/README for the benchmark version and commit level right?

yes, you are absolutely right. I forgot to do that. Will do.

tcormackMW (approved):

PR #1064, opened 2025-03-04
New Dynamic Targeting Mechanism

tcormackMW:
This change allows users to, for Factory Components, create 'injection tags' in their reference targets. These tags, enclosed in `{{someKey}}`, correspond to the Configuration that is used to create a specific factory instance.

- If the key does not exist in the configuration that causes the factory instance to be created, an error is thrown.
- If the resulting LDAP filter is invalid, an error is thrown. 

This allows dynamic targeting based on the value of the configuration without forcing a user to create an intermediary to listen for specific configs and then manually inject a `refName.target=...` value.

### DIFF
--- a/compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
+++ b/compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
@@ -31,80 +31,43 @@
 #include "cppmicroservices/SharedLibraryException.h"
 #include "cppmicroservices/asyncworkservice/AsyncWorkService.hpp"
 #include "cppmicroservices/cm/ConfigurationAdmin.hpp"
+#include <regex>
 
 namespace cppmicroservices::scrimpl
 {
-    std::string const targetOpenDelimiter = "${";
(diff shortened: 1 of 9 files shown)


jeffdiclemente (commented):
In general, I think this change warrants a small, standalone RFC and a changes to the existing DS RFC to explain the rules for using dynamic targeting. These RFCs should not hold up this PR, but they are necessary in the near future.

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
```diff
@@ -1,140 +1,185 @@
...
+            std::string key = match[1].str();
+            if (auto iter = properties.find(key); iter != properties.end())
             {
-                throw std::invalid_argument("ComponentFactoryImpl Constructor "
-                                            "provided with invalid arguments");
+                result.append(cppmicroservices::ref_any_cast<std::string>(iter->second));
             }
+            else
+            {
+                throw std::invalid_argument("reference target not in configuration " + key);
```
jeffdiclemente, line 64: it looks like there is no test for this code path. please add one.

PR #1067, opened 2025-03-11
Resolve valgrind 'possibly lost' leaks

jeffdiclemente:
Add a general suppression to ignore pthread tls allocation. Through investigation this memory flagged as possibly lost by valgrind appears to be a false positive; a result of how pthread manages allocating TLS memory and repeatedly creating/joining and destroying std::thread objects in the test process.


framework/src/util/FrameworkPrivate.cpp
```diff
@@ -51,6 +51,14 @@ namespace cppmicroservices
         stopEvent = FrameworkEventInternal { true, FrameworkEvent::Type::FRAMEWORK_ERROR, std::string(), nullptr };
     }
 
+    FrameworkPrivate::~FrameworkPrivate() noexcept
+    {
+        if (shutdownThread.joinable())
```
tcormackMW, line 56: I agree this should be happening, but I'm curious where this was happening before -- or was it not?

tcormackMW (approved):

framework/src/util/FrameworkPrivate.cpp
jeffdiclemente: prior to this change, `shutdownThread` was being joined on line 149 of `FrameworkPrivate::WaitForStop()`.

PR #1069, opened 2025-03-27
Update minimum compilers for Github Workflows

Monika032:
Updated the minimum compilers used on Linux to GCC 11 and Windows mingw configurations to use Windows 2022. This enables full support for std::filesystem APIs. 

Signed-off-by: The MathWorks, Inc. [mphadnis@mathworks.com](mailto:mphadnis@mathworks.com)


jeffdiclemente (commented):
@Monika032, is this PR still needed? can it be deleted?

PR #1070, opened 2025-04-01
`getServiceObjects` fix to use `customDeleter`

tcormackMW:
Not sure how I missed this in #1044, but GetServiceObjects doesn't use the custom deleter. That is a problem.


jeffdiclemente (approved):

Burgch:
Nice, thanks Toby!

PR #1072, opened 2025-04-15
latent bug -- concurrent configuration creations for factories race

tcormackMW:
Similar to #834

Lock down the `managers` object within `SCRBundleExtension` to make sure concurrent updates don't happen

before fix, the new test cases failed 50-80% of the time either at:

`            managers.push_back(std::move(compManager));` 

or

`                auto fut = compManager->Disable(singleInvoke);`

After fix, ran 1000 times with no failure

### DIFF
--- a/compendium/DeclarativeServices/test/gtest/TestFactoryPid.cpp
+++ b/compendium/DeclarativeServices/test/gtest/TestFactoryPid.cpp
@@ -20,6 +20,7 @@
 
 =============================================================================*/
 
+#include "ConcurrencyTestUtil.hpp"
 #include "TestFixture.hpp"
 #include "gtest/gtest.h"
 
@@ -72,7 +73,8 @@ namespace test
         auto fut = factoryConfig->Update(props);
         fut.get();
         // Confirm the properties have been updated in DS.
-        compDescDTO = dsRuntimeService->GetComponentDescriptionDTO(testBundle, factoryComponentName + "_" + factoryInstance);
+        compDescDTO
+            = dsRuntimeService->GetComponentDescriptionDTO(testBundle, factoryComponentName + "_" + factoryInstance);
         EXPECT_EQ(compDescDTO.implementationClass, factoryComponentName)
             << "Implementation class in the returned component description must be " << factoryComponentName;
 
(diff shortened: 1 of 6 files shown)


compendium/DeclarativeServices/test/gtest/TestFactoryPid.cpp
```diff
@@ -209,4 +211,59 @@ namespace test
         auto instances = GetInstances<test::CAInterface>();
         EXPECT_EQ(instances.size(), count);
     }
+
+    /* test concurrentFactoryCreation.
+     * This test creates 1000 factory objects without waiting for the previous one to complete.
```
Burgch, line 216: Does each config create 10 objects? Or should this be 100?

compendium/DeclarativeServices/test/gtest/TestFactoryPid.cpp
```diff
@@ -209,4 +211,59 @@ namespace test
...
+                fut.wait();
+            };
+
+            // Create the factory configuration object
+            auto factoryConfig = configAdminService->CreateFactoryConfiguration(configurationPid);
+
+            // Update the properties for the factory configuration object
+            cppmicroservices::AnyMap props(cppmicroservices::AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS);
+            std::string const instanceId { "instance" + std::to_string(i) };
+            props["uniqueProp"] = instanceId;
```
Burgch, line 254: I'm guessing this was intended to be passed to `processConfiguration` and then be passed in to `Update` ?

compendium/DeclarativeServices/test/gtest/TestFactoryPid.cpp
```diff
@@ -209,4 +211,59 @@ namespace test
...
+
+        // Get a service reference to ConfigAdmin to create the factory component instances.
+        auto configAdminService = GetInstance<cppmicroservices::service::cm::ConfigurationAdmin>();
+        ASSERT_TRUE(configAdminService) << "GetService failed for ConfigurationAdmin";
+
+        // Create some factory configuration objects. Don't wait for one to complete before
+        // creating the next one.
+        constexpr auto count = 100;
+        std::vector<std::shared_future<void>> futures;
+        Barrier sync_point(count); // 60 threads to synchronize
```
Burgch, line 236: 100?

Burgch (approved):
Nice catch, thanks Toby!

compendium/DeclarativeServices/test/gtest/TestFactoryPid.cpp
tcormackMW: I suppose yes. That being said, I don't think we need it. There's no reason, in the test, for us to have that unique prop. This was just a copy and paste from a different test that had this sort of functionality, but I don't think it is needed at all

compendium/DeclarativeServices/test/gtest/TestFactoryPid.cpp
tcormackMW: it should be 100... leftover from when I was a bit overzealous in how much concurrency this test needed :)

jeffdiclemente (approved):

PR #1079, opened 2025-05-07
Fix binding for prototype services

tcormackMW:
Binding for prototype services used to call `getService` twice for every bind operation:

1. once to put into the `compContext` `boundservicesCache`
2. anther time to invoke `bind` on the referencING service (i.e. to actually give the service the pointer to the bound service)

Because, for prototype services, `getService` ALWAYS returns a new instance of the service, these pointers were not the same. 

In removal from the `boundServicesCache`, we were checking the value of this service to remove it... this doesn't work if the pointers aren't the same. 

This commit seeks to fix that and add some testing.


compendium/DeclarativeServices/src/ComponentContextImpl.hpp
```diff
@@ -222,7 +247,11 @@ namespace cppmicroservices
 
             std::weak_ptr<ComponentConfiguration> configManager;
             cppmicroservices::Bundle usingBundle;
-            mutable Guarded<std::unordered_map<std::string, std::vector<cppmicroservices::InterfaceMapConstPtr>>>
+            // map of reference name to std::vector of pairs of ServiceReferenceBase to std::shared_ptr<void> of bound
+            // service in order that they were added to the boundServiceCache
+            mutable Guarded<std::unordered_map<
+                std::string,
+                std::vector<std::pair<cppmicroservices::ServiceReferenceBase, InterfaceMapConstPtr>>>>
```
Burgch, line 254: Any reason not to make this an `unordered_map<ServiceReferenceBase, InterfaceMapConstPtr>` ?
```suggestion
                std::unordered_map<cppmicroservices::ServiceReferenceBase, InterfaceMapConstPtr>>>
```

compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.cpp
```diff
@@ -21,12 +21,12 @@
   =============================================================================*/
 
 #include "SingletonComponentConfiguration.hpp"
-#include "SingleInvokeTask.hpp"
 #include "../ComponentRegistry.hpp"
 #include "ComponentManager.hpp"
 #include "ReferenceManager.hpp"
 #include "ReferenceManagerImpl.hpp"
 #include "RegistrationManager.hpp"
+#include "SingleInvokeTask.hpp"
```
Burgch, line 29: Not used?
```suggestion
```

Burgch (approved):
Thanks Toby, this looks great!

compendium/DeclarativeServices/src/ComponentContextImpl.hpp
tcormackMW: yep. the ordering is used to determine which has the lower service id... locateService relies on that ordering... 

but that is a second conversation to have later on. In theory `LocateService` (the old one) returns the highest ranked service. This is not true at all... I think this used to be the case when dynamic binding didn't exist and the cache was the cache and never changed and the initial bound services which are retrieved from teh context are returned in order so by virtue of that, the first one is the highest ranked, but with dynamic rebinding and changes to the cache this is just not true anymore. However, that is probably separate from this pr and can be addressed separately.

compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.cpp
tcormackMW: it is on line 184

in getService

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -281,14 +319,20 @@ namespace cppmicroservices
...
+                    serviceInfos.end(),
+                    [&services, &configManagerPtr, &name, &type](
+                        std::pair<cppmicroservices::ServiceReferenceBase, InterfaceMapConstPtr> const& serviceInfo)
+                    {
+                        auto const& intMapPtr = serviceInfo.second;
+                        auto const servicePtr = GetServicePointer(configManagerPtr, intMapPtr, name, type);
+                        if (servicePtr)
+                        {
+                            services.push_back(servicePtr);
+                        }
```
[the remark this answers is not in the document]
tcormackMW: Yeah probably. I'm not sure if that is relied on -- golly i hope not. but definitely worth noting

compendium/DeclarativeServices/src/ComponentContextImpl.hpp
Burgch: Ah yes, I see - that definitely sounds separate PR-worthy! Thanks :)

compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.cpp
Burgch: Ah, apologies!

tcormackMW: no worries :)

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -80,12 +80,13 @@ namespace cppmicroservices
                                       auto bc = GetBundleContext();
                                       auto boundServicesCacheHandle = boundServicesCache.lock();
                                       auto& serviceMap = (*boundServicesCacheHandle)[refName];
+                                      // get us the service instance
                                       cppmicroservices::ServiceObjects<void> sObjs = bc.GetServiceObjects(sRefU);
                                       auto interfaceMap = sObjs.GetService();
-                                      if (interfaceMap)
+                                      if (interfaceMap && interfaceMap->size() != 0)
```
carneyweb, line 86: Consider "!interfaceMap->empty()" as more descriptive that comparing size to zero.

carneyweb (approved):
LGTM.

Fix, then ship.

compendium/DeclarativeServices/test/gtest/TestBundleValidation.cpp
```diff
@@ -166,6 +166,8 @@ TEST(TestBundleValidation, BundleValidationFailure)
     compDesc = dsRuntimeService->GetComponentDescriptionDTO(*bundleIter, "sample::ServiceComponentBV1_3");
     ASSERT_TRUE(dsRuntimeService->IsComponentEnabled(compDesc));
     svcRef = f.GetBundleContext().GetServiceReference<test::Interface1>();
+    auto allRefs = f.GetBundleContext().GetServiceReferences<test::Interface1>();
```
jeffdiclemente, line 169: `allRefs` is never used. What is the significance of getting all the service references here?

jeffdiclemente (approved):

compendium/DeclarativeServices/test/gtest/TestBundleValidation.cpp
tcormackMW: was added for debugging when we were failing this test due to ordering. removed it.

compendium/DeclarativeServices/src/manager/BundleLoader.cpp
```diff
@@ -163,9 +164,9 @@ namespace cppmicroservices
                 bundleBinaries.lock()->emplace(bundleLoc, handle);
             }
 
-            const std::string symbolName = std::regex_replace(compName, std::regex("::"), "_");
-            const std::string newInstanceFuncName("NewInstance_" + symbolName);
-            const std::string deleteInstanceFuncName("DeleteInstance_" + symbolName);
+            std::string const symbolName = std::regex_replace(compName, std::regex("::"), "_");
```
ShivamNegi, line 167: Regex replace has the reputation of being very slow. Especially for simple replacements.
As a suggestion a simple replace lambda loop could be added here. For estimated 10x improvement.

PR #1085, opened 2025-05-20
replace use of std::regex_replace

tcormackMW:


Burgch (approved):
Looks good, thanks Toby!

ShivamNegi:
Passing by reference for function.

tcormackMW:
> Passing by reference for function.

@ShivamNegi  not sure I agree here. I don't know that we want to edit the original variable. The function is written to return the edited variable. I think that's how I prefer it. The performance gains from a string pass by reference instead of an inline function is virtually non-existent imo

jeffdiclemente (changes requested):
putting this in the public API doesn't seem right. There's no reason for users of cppms to call this. Its adding a public API surface that we don't want to expose. Is it possible to put this in the `util` static library and then link that into both libraries?

PR #1086, opened 2025-05-21
fixes #1083: export locateService

### LINKED ISSUE #1083: `LocateService(refName, sRef)` is a public api within OSGI
`LocateService(std::string refName, ServiceReferenceBase sRef)` is [a supported DS public API](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.component.html#org.osgi.service.component.ComponentContext.locateService-String-ServiceReference-) for downstream clients to use.

This should be a public API for users through `include "cppmicroservices\servicecomponent\ComponentContext.hpp"`

tcormackMW:


compendium/ServiceComponent/include/cppmicroservices/servicecomponent/ComponentContext.hpp
```diff
@@ -172,6 +172,26 @@ namespace cppmicroservices
                     return std::static_pointer_cast<T>(sObj);
                 }
 
+                /**
+                 * Returns the service object from the cache matching the passed in
```
jeffdiclemente, line 176: the cache is an implementation detail that I don't think should be mentioned in the public facing documentation. The user shouldn't care from where the service object is retrieved.

compendium/ServiceComponent/include/cppmicroservices/servicecomponent/ComponentContext.hpp
```diff
@@ -221,6 +241,23 @@ namespace cppmicroservices
                  */
                 virtual std::shared_ptr<void> LocateService(std::string const& name, std::string const& type) const = 0;
 
+                /**
+                 * Returns the service object from the cache matching the passed in
```
jeffdiclemente, line 245: the cache is an implementation detail that I don't think should be mentioned in the public facing documentation. The user shouldn't care from where the service object is retrieved.

compendium/DeclarativeServices/src/ComponentContextImpl.cpp
```diff
@@ -250,30 +250,30 @@ namespace cppmicroservices
...
+                    return GetServicePointer(configManagerPtr, interfaceMapPtr, refName, type);
                 }
             }
 
             return nullptr;
         }
 
         std::shared_ptr<void>
-        ComponentContextImpl::LocateService(std::string const& name, cppmicroservices::ServiceReferenceBase const& sRef)
+        ComponentContextImpl::LocateService(std::string const& name, cppmicroservices::ServiceReferenceBase const& sRef) const
```
jeffdiclemente, line 276: is this `name` parameter a reference name? If so, I recommend changing the name of parameter to `refName`

jeffdiclemente (changes requested):

PR #1089, opened 2025-05-22
enable Windows and Linux ARM builds

jeffdiclemente:
experimenting with Windows and Linux ARM github hosted runners.

### DIFF
--- a/.github/workflows/build_and_test_arm.yml
+++ b/.github/workflows/build_and_test_arm.yml
@@ -1,0 +1,51 @@
+name: BuildAndTestARM
+
+on:
+  push:
+    branches: [ development ]
+  pull_request:
+    branches: [ development ]
+  workflow_dispatch:
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Release
+  BUILD_DIR: ${{github.workspace}}/build
+  BUILD_CONFIGURATION: 0
+  WITH_COVERAGE: 1
+  MY_COVERAGE: gcov
+
+jobs:
+  build_arm:
+    name: Build and Test on ARM [${{matrix.os}},Configuration=${{matrix.buildconfiguration}}]
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-22.04-arm, windows-11-arm]
+        # build configurations:
+        # 0 = threading ON / shared lib ON
+        # 1 = threading ON / shared lib OFF
+        # 2 = threading OFF / shared lib ON
+        # 3 = threading OFF/ shared lib OFF
+        buildconfiguration: [0,1]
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
YarHokie007, line 1: Is there an issue open to track enabling the services defined within US_NEVER_GOING_TO_BE_DEFINED

compendium/ConfigurationAdmin/src/CMakeLists.txt
YarHokie007, line 1: Why was this change originally just for x86_64, is this a Linux only code path or a no-op for non Linux platforms and that allowed you to generalize it?

CMakeLists.txt
jeffdiclemente: no, this was a workaround to build the stable, core libraries of cppms.

The service disabled in this PR are "experimental" and not meant for production use.  this change will be resolved as part of this PR; either the experimental services will be made to work for ARM or we will not build them for ARM.

compendium/ConfigurationAdmin/src/CMakeLists.txt
jeffdiclemente: Its not a linux only code path. The change originated in [this PR](https://github.com/CppMicroServices/CppMicroServices/pull/487/files#diff-8b8183a19e60dc18bb8006f816783dd1da2c47f730f43e853442db720ab7f4ef). I don't know why it was made, my guess is we didn't think we would be building on ARM. I can't determine anything else about the reason to restrict to x86_64 from the PR itself.

compendium/DeclarativeServices/src/CMakeLists.txt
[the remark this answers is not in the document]
jeffdiclemente: Its not a linux only code path. The change originated in [this PR](https://github.com/CppMicroServices/CppMicroServices/commit/35e20cc08ae1905c43431440cce8b8cb223dde8c#diff-11674beef75ce6f2cba8b43171e0ba00360c728287c02ce15e5c46f0350a3b19). I don't know why it was made, my guess is we didn't think we would be building on ARM. I can't determine anything else about the reason to restrict to x86_64 from the PR itself.

cmake/usFunctionCreateTestBundle.cmake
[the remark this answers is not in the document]
jeffdiclemente: Its not a linux only code path. The change originated in [this PR](https://github.com/CppMicroServices/CppMicroServices/commit/93f751ccfbec6f3c67218c8dc70ea4b932a4f8b0#diff-ccf50ac46c9c2a07093fbad6f4e6613fdd6b1fad943c62bc57d577e6e5613769). I don't know why it was made, my guess is we didn't think we would be building on ARM. I can't determine anything else about the reason to restrict to x86_64 from the PR itself.

tcormackMW (approved):

carneyweb (approved):

Issue #1005, opened 2024-04-18
Add support for JSON comments similar to those supported by VSCode


BuzzwordSea:
The lack of official comment support in manifest.json files is a significant pain, particularly for products with many manifest.json files.  Please add support for comments.  As an example VSCode configuration supports // and /* */ style comments matching JavaScript comments.

PR #1094, opened 2025-05-22
Add support for C++ Style Comments

ShivamNegi:
#1005

The jsoncpp and rapidjson libraries both support C++ style comments now. Enabling them in the code.

Updated a testbundle to validate if parsing is working as expected.

### DIFF
--- a/framework/test/bundles/libM/resources/manifest.json
+++ b/framework/test/bundles/libM/resources/manifest.json
@@ -1,18 +1,19 @@
 {
+  // Adding comments to see if parsing test still work
   "bundle.symbolic_name": "TestBundleM",
   "bundle.description": "My Bundle description",
   "bundle.version": "1.0.0",
-  "bundle.activator" : true,
+  "bundle.activator": true, /* C style comments are enabled now. */
   "number": 5,
   "double": 1.1,
   "vector": [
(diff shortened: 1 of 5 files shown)


framework/test/bundles/libM/resources/manifest.json
```diff
@@ -1,18 +1,19 @@
 {
+  // Adding comments to see if parsing test still work
```
tcormackMW, line 2: This isn't a c-style comment right? If not, how is it allowed to be in the manifest?

ShivamNegi: Oops, I meant C/ C++ style comments. Allowing both // and /* type comments in the JSON files.

tcormackMW: Ah so c++ style works too. cool cool

tcormackMW (approved):

jeffdiclemente (commented):
is it possible to access the comments programmatically using RapidJSON and/or jsoncpp after the JSON is parsed?

ShivamNegi:
It may be possible to access the comments, as these libraries support saving them. The key question is whether there is a real need to save the comments?

Issue #268, opened 2018-04-24
Bundle symbolic name requirements


Innokentiy-Alaytsev:
Hello!

I have searched through documentation and did not find any requirements on bundle symbolic name. In fact, symbolic name is required to be a correct C identifier.

In case someone decides to use a naming scheme like in Java OSGi implementations (`org.some.domain.bundle.component.name`) in `manifest.json` alongside with a proper `US_BUNDLE_NAME` macro definition (1org_some_domain_bundle_component_name1), everything will compile but nothing will actually work.

Best regards,
Innokentiy

karthikreddy09:
Good catch! The documentation needs to be updated.

Also, The resource compiler could be modified to detect illegal symbolic names and emit an error at build time instead of silent failure at runtime.

@Innokentiy-Alaytsev 
I am assuming it works fine if the symbolic name in the manifest.json and US_BUNDLE_NAME are both set to `org_some_domain_bundle_component_name`. Let us know otherwise.

Innokentiy-Alaytsev:
@karthikreddy09, everything works fine if the symbolic name in the manifest.json and US_BUNDLE_NAME are both set to `org_some_domain_bundle_component_name`.

About resource compiler modification to catch this kind of errors: it should be possible to check correctness with [JSON Schema](http://json-schema.org/). I have used [RapidJSON](https://github.com/Tencent/rapidjson) for this purpose. The library is header only so it should be possible to provide it alongside with the C++ Micro Services sources without introducing an external dependency.

karthikreddy09:
@Innokentiy-Alaytsev 
Thanks for the pointers. Schema validation has been on our wishlist for other purposes.
Currently, CppMicroServices ships with [jsoncpp](https://github.com/open-source-parsers/jsoncpp) which does not support schema validation. We may have to re-evaluate the JSON dependency.

PR #1100, opened 2025-05-28
I0268

carneyweb:
Issue #268 

- Updated schema to correctly check that the symbolic_name follows c identifier rules
- Updated bundle documentation with those requirements.

### DIFF
--- a/framework/doc/bundle_properties.rst
+++ b/framework/doc/bundle_properties.rst
@@ -17,7 +17,8 @@ Bundle authors must always add the following property to their bundle's
 ``manifest.json`` file:
 
 -  ``bundle.symbolic_name`` - The human readable name of the bundle (type
-   ``std::string``)
+   ``std::string``). The name must start with either an underscore or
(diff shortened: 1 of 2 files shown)


framework/doc/bundle_properties.rst
```diff
@@ -17,7 +17,8 @@ Bundle authors must always add the following property to their bundle's
 ``manifest.json`` file:
 
 -  ``bundle.symbolic_name`` - The human readable name of the bundle (type
-   ``std::string``)
+   ``std::string``). The name must start with either an underscore or
+   letter and contain only underscores, letters, or numbers.
```
tcormackMW, line 21: I would add "(i.e. must be a correct c-identifier)"

This just gives some context to an otherwise arbitrary rule

tcormackMW (approved):

framework/doc/bundle_properties.rst
carneyweb: Done.

PR #1107, opened 2025-06-11
Fix Bundle Find Hooks to take into consideration systemBundle

tcormackMW:


framework/test/gtest/BundleHooksTest.cpp
```diff
@@ -161,24 +161,85 @@ TEST_F(BundleHooksTest, TestFindHook)
...
+
+    // now that it exists, if installed with non-system bundle it will fail
+    bool caught = false;
+    try {
+        bundleA = cppmicroservices::testing::InstallLib(bundleB.GetBundleContext(), "TestBundleA");
     }
+    catch (...) {
+        caught = true;
+    }
+    ASSERT_TRUE(caught);
```
jeffdiclemente: Its much easier to put all of this code into an `ASSERT_THROWS(...)` macro instead of having the `try/catch` block

framework/test/gtest/BundleHooksTest.cpp
```diff
@@ -161,24 +161,78 @@ TEST_F(BundleHooksTest, TestFindHook)
...
+        // because it invoked getBundles, nothing is thrown... that only happens on an actual instal
+#if defined(US_BUILD_SHARED_LIBS)
+    // on first installation, the bundle will not exist and therefore will be installed regardless of hooks
+    bundleA = cppmicroservices::testing::InstallLib(bundleB.GetBundleContext(), "TestBundleA");
+    ASSERT_TRUE(bundleA);
+    ASSERT_EQ(bundleA.GetSymbolicName(), "TestBundleA");
+
+
+    // now that it exists, if installed with non-system bundle it will fail
+    ASSERT_THROW(cppmicroservices::testing::InstallLib(bundleB.GetBundleContext(), "TestBundleA"), std::runtime_error);
```
jeffdiclemente, line 220: Given that the `TestBundleFindHook` is being used and it doesn't throw, I'm not sure I see how this line throws an exception. Shouldn't this line not throw an exception?

tcormackMW: It throws because all bundles were rejected by the filter hook on a bundle that is already installed. This is a peculiarity of the choices made in the filtering logic. I am guessing it is some logic copied from knoplerfish or felix, but if an install on a bundle that already exists fails for all possible bundles, it throws:

https://github.com/CppMicroServices/CppMicroServices/blob/257679ed22b9842f96d1d3859b3488619e413354/framework/src/bundle/BundleRegistry.cpp#L203

jeffdiclemente: I see now, this behavior looks like its an implementation of the [bundle find hooks spec](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.bundlehooks.html#org.osgi.framework.hooks.bundle.FindHook.find-BundleContext-Collection-):

> Bundle install operations when an existing bundle is already installed at a given location. In this case, the find method is called to determine if the context performing the install operation is able to find the bundle. If the context cannot find the existing bundle then the install operation must fail with a [BundleException.REJECTED_BY_HOOK](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.api.html#org.osgi.framework.BundleException.REJECTED_BY_HOOK) exception.

PR #1112, opened 2025-06-27
New mechanism for `inject-references`

tcormackMW:
This value defaults to true already, but the user can supply a different bool to not inject any references.

This proposes an extension where the user can supply a bool or an array of strings. These strings would list the names of the references the service wishes to inject.


jeffdiclemente:
I'm struggling to understand how this would be used by service authors. Can you elaborate on the use case for `require-bind`?

compendium/tools/SCRCodeGen/ManifestParserImpl.cpp
```diff
@@ -169,6 +169,12 @@ ManifestParserImplV1::ParseAndGetComponentInfos(Json::Value const& scr) const
                     refInfo.target
                         = JsonValueValidator(jsonRefInfo, "target", Json::ValueType::stringValue).GetString();
                 }
+
+                if (jsonRefInfo.isMember("require-bind")){
```
Burgch: Could we use `inject-reference` ? Then it could apply to both static and dynamic references? For static, it would mean they are skipped out of the references injected into the constructor. For dynamic, we wouldn't require Bind/Unbind methods for that reference?

Burgch (commented):
I think a per-reference override makes a lot of sense - we have references now that we use purely as "enabling" dependencies, and it would be nice not to need to include them in the shared_ptr arguments in the constructor!

compendium/tools/SCRCodeGen/ManifestParserImpl.cpp
tcormackMW: Yeah I was thinking about this, right now the fix only applies to dynamic so I will look into making it work for static injections as well. Then I'll rename it

tcormackMW:
...
> 
> I feel like there is a simpler way than introducing `require-bind` to accomplish the goal of defining a service dependency that doesn't participate in any of the dependency injection methods (i.e. constructor and bind/unbind). I would like us to explore the existence of other options.
> 
> As far as I can tell, there is nothing in the OSGi spec that says that a dynamic service reference requires a bind/unbind method to exist. Research has not uncovered a definite requirement out in the wild either. Nor do I see a similar requirement for constructor injection (i.e. you can write a constructor that doesn't take a statically bound service reference as a parameter). So if we change the DS design to not require a bind/unbind nor require every service ref to be a constructor parameter, does that solve the same problem as `require-bind`? What other problems would it create (i.e. what are the trade-offs)?
> 
> Re-reading the [DS RFC](https://github.com/CppMicroServices/rfcs/blob/master/text/0003-declarative-services.md#dependency-injection), I don't see a reason why we originally decided to make bind/unbind and all static service refs in a ctor param list mandatory (nor do I remember why we decided to do it that way). I suspect it was with a mind towards not making user errors "hidden" or hard to detect. For example, if a user expects to be notified of a dynamic bind but forgot to write a "bind" method, how would they know? when would they know? Enforcing that bind/unbind must exist is a way to solve this problem. For constructor injection, I believe it was due to the limitations of the C++ language at the time DS was implemented. Requiring all static refs to be in the ctor param list was the easiest way to implement OSGi ctor injection. C++26 Reflection _**could**_ make it such that we don't need `inject-references` anymore.

@jeffdiclemente @Burgch 
Mmm interesting. 

I read through the OSGI spec and I think I have some idea why we wrote it this way... As far as I can tell, there is a somewhat analogous field in osgi, that being the _init_ field. For constructor injection at least, "If the constructor has a parameter count that does not match the value of the init attribute in the component element, then the constructor must not be considered. If the value of the init attribute is 0, the default value, then the public no-parameter constructor must be used."

This _init_ field essentially indicates how many of the references should be injected... in our case we just said that it should be an all or nothing type situation...

Additionally, I don't see anywhere in the osgi spec where, if the _init_ value < total number of static references, we specify that this one should be one of the injected ones rather than this one. For references where each implements a different interface, this is obvious based on the constructors provided by the user, but for cases where a service references multiple of the same interface differentiated only by the _target_, for example, this is not an obvious case. 

I think we likely said "this is vague and unclear and probably a little bit of a hand wavy part of the spec, lets just do the all or nothing which we will indicate with **inject-references**..."

What do you think?

https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#d0e29786:~:text=If%20the%20constructor%20has%20a%20parameter%20count%20that%20does%20not%20match%20the%20value%20of%20the%20init%20attribute%20in%20the%20component%20element%2C%20then%20the%20constructor%20must%20not%20be%20considered.%20If%20the%20value%20of%20the%20init%20attribute%20is%200%2C%20the%20default%20value%2C%20then%20the%20public%20no%2Dparameter%20constructor%20must%20be%20used.

Burgch:
I think there's a lot of value in avoiding ambiguity, and in helping developers to catch coding errors. Neither is possible by trying to infer what the user intended. Various use cases:

1. As Toby mentions, two references of the same interface, but only one argument in the constructor arguments with that type. Which reference is injected to that argument?
2. Multiple constructors which support different combinations of the references. Which constructor should be called?
3. BindX method with a typo in the method name. There's no way to tell whether this is an intentional method which is unrelated to CppMicroServices dynamic reference injection, and the reference wasn't meant to be injected, OR it is a typo and the user expected this method to be called but it never is.

These cases are not resolved by C++26 reflection, because they are always going to be ambigious. All of these would be addressed with an explicit, per-reference override of the `inject-references` value, so I support the overall direction that Toby is pursuing here, as long as it applies to both dynamic and static references.

jeffdiclemente:
> I think there's a lot of value in avoiding ambiguity, and in helping developers to catch coding errors. Neither is possible by trying to infer what the user intended. Various use cases:
> 
> 1. As Toby mentions, two references of the same interface, but only one argument in the constructor arguments with that type. Which reference is injected to that argument?
> 2. Multiple constructors which support different combinations of the references. Which constructor should be called?
> 3. BindX method with a typo in the method name. There's no way to tell whether this is an intentional method which is unrelated to CppMicroServices dynamic reference injection, and the reference wasn't meant to be injected, OR it is a typo and the user expected this method to be called but it never is.
> 
> These cases are not resolved by C++26 reflection, because they are always going to be ambigious. All of these would be addressed with an explicit, per-reference override of the `inject-references` value, so I support the overall direction that Toby is pursuing here, as long as it applies to both dynamic and static references.

That all makes sense and I agree that there we should favor eliminating ambiguity. The approach in this PR looks good to me.

compendium/tools/SCRCodeGen/ComponentInfo.hpp
```diff
@@ -52,12 +54,52 @@ namespace codegen
...
+            // instantiate the set of injected refs or the 'all' bool
+            void setInjectReferences(bool allRefs, std::unordered_set<std::string> refNames = {}){
+                // if the set of refNames is empty, set the injectAll param
+                if (refNames.empty()){
+                    injectAllRefs = allRefs;
+                    return;
+                }
+
+                // set injectedRefs
+                injectedRefs = refNames;
```
vivekkushwaha39, line 81: Since these two variables (injectedRefs, injectAllRefs) are dependent, we can store them as pair or variant. This will make code more expressive.

tcormackMW: Hmm interesting. I'm not sure I am 100% on board here. if we were to combine this into a pair or variant, there would either be work to decide of which type the variant is including try/catch blocks, or referencing the first and second part of a pair, ie `refs.first` or `refs.second` which seems less expressive in my opinion. 

What do you think @vivekkushwaha39 ?

jeffdiclemente:
@tcormackMW , Thread sanitizer found an issue when running `TestServicePropsChangeForBind`. It would be good to understand the nature of the issue.

tcormackMW:
> @tcormackMW , Thread sanitizer found an issue when running `TestServicePropsChangeForBind`. It would be good to understand the nature of the issue.

Interesting, I didn't see that... 

It looks like M1 in this case is the `oneAtATimeMutex` in `CCActiveState` and M0 is the mutex in `ComponentContextImpl` that is used in the `Guarded` member variable `boundServiceCache`. 

The two stacks look like this:

```
#0 pthread_mutex_lock
#1 cppmicroservices::scrimpl::CCActiveState::Activate FINDME<LOCK M1>
#2 cppmicroservices::scrimpl::CCRegisteredState::Activate
#3 cppmicroservices::scrimpl::ComponentConfigurationImpl::Activate
#4 cppmicroservices::scrimpl::SingletonComponentConfigurationImpl::GetService
#5 non-virtual thunk to cppmicroservices::scrimpl::SingletonComponentConfigurationImpl::GetService
#6 cppmicroservices::ServiceReferenceBasePrivate::GetServiceFromFactory
#7 cppmicroservices::ServiceReferenceBasePrivate::GetServiceInterfaceMap
#8 cppmicroservices::ServiceObjectsBasePrivate::GetServiceInterfaceMap
#9 cppmicroservices::ServiceObjectsBase::GetServiceInterfaceMap
#10 cppmicroservices::ServiceObjects<void>::GetService
#11 cppmicroservices::scrimpl::ComponentContextImpl::InitializeServicesCache() FINDME <LOCK M0 WHILE GETTING SERVICE>
#12 cppmicroservices::scrimpl::ComponentContextImpl::ComponentContextImpl
#14 cppmicroservices::scrimpl::ComponentConfigurationImpl::CreateAndActivateComponentInstanceHelper
#15 cppmicroservices::scrimpl::SingletonComponentConfigurationImpl::CreateAndActivateComponentInstance
#16 cppmicroservices::scrimpl::CCActiveState::Activate
#17 cppmicroservices::scrimpl::CCRegisteredState::Activate
#18 cppmicroservices::scrimpl::ComponentConfigurationImpl::Activate
#19 cppmicroservices::scrimpl::CCUnsatisfiedReferenceState::Register
#20 cppmicroservices::scrimpl::ComponentConfigurationImpl::RefSatisfied
#21 cppmicroservices::scrimpl::ComponentConfigurationImpl::RefChangedState
#23 cppmicroservices::scrimpl::ReferenceManagerBaseImpl::BatchNotifyAllListeners
#24 cppmicroservices::scrimpl::ReferenceManagerBaseImpl::BindingPolicyDynamicReluctant::ServiceAdded
#25 cppmicroservices::scrimpl::ReferenceManagerBaseImpl::AddingService
#27 non-virtual thunk to cppmicroservices::detail::TrackedService::CustomizerAdding
#28 cppmicroservices::detail::BundleAbstractTracked<cppmicroservices::ServiceReference::TrackAdding
#29 cppmicroservices::detail::BundleAbstractTracked::Track
#30 cppmicroservices::detail::TrackedService:ServiceChanged
#32 cppmicroservices::ServiceListenerEntry::CallDelegate(cppmicroservices::ServiceEvent const&)
#33 cppmicroservices::ServiceListeners::ServiceChanged
#34 cppmicroservices::ServiceListeners::ServiceChanged
#35 cppmicroservices::ServiceRegistry::RegisterService
#36 cppmicroservices::BundleContext::RegisterService
#38 cppmicroservices::scrimpl::ComponentConfigurationImpl::RegisterService()
#39 cppmicroservices::scrimpl::CCUnsatisfiedReferenceState::Register(cppmicroservices::scrimpl::ComponentConfigurationImpl&)
#40 cppmicroservices::scrimpl::ComponentConfigurationImpl::Initialize()
#41 cppmicroservices::scrimpl::ComponentConfigurationFactory::CreateConfigurationManager
#42 cppmicroservices::scrimpl::CMEnabledState::CreateConfigurations
#43 std::enable_if<is_invocable_r_v<void, cppmicroservices::scrimpl::ComponentManagerImpl::PostAsyncDisabledToEnabled
```

```
#0 pthread_mutex_lock <null> 
#1 cppmicroservices::scrimpl::SingletonComponentConfigurationImpl::DestroyComponentInstances() FINDME <LOCK M0 IN DESTRUCTION OF COMPONENTCONTEXTIMPL>
#2 cppmicroservices::scrimpl::CCActiveState::Deactivate
#3 cppmicroservices::scrimpl::ComponentConfigurationImpl::RefUnsatisfied
#4 cppmicroservices::scrimpl::ComponentConfigurationImpl::RefChangedState
#6 cppmicroservices::scrimpl::ReferenceManagerBaseImpl::BatchNotifyAllListeners
#7 cppmicroservices::scrimpl::ReferenceManagerBaseImpl::BindingPolicy::DynamicRemoveService
#8 cppmicroservices::scrimpl::ReferenceManagerBaseImpl::BindingPolicyDynamicReluctant::ServiceRemoved
#9 non-virtual thunk to cppmicroservices::scrimpl::ReferenceManagerBaseImpl::RemovedService
#10 non-virtual thunk to cppmicroservices::detail::TrackedService::CustomizerRemoved
#11 cppmicroservices::detail::BundleAbstractTracked::Untrack
#12 cppmicroservices::detail::TrackedService::ServiceChanged
#14 cppmicroservices::ServiceListenerEntry::CallDelegate
#15 cppmicroservices::ServiceListeners::ServiceChanged
#17 cppmicroservices::ServiceRegistrationBase::SetProperties
#18 cppmicroservices::scrimpl::RegistrationManager::SetProperties
#19 cppmicroservices::scrimpl::ComponentConfigurationImpl::SetRegistrationProperties()
#20 cppmicroservices::scrimpl::CCActiveState::Modified     FINDME<LOCK M1>
#21 cppmicroservices::scrimpl::ComponentConfigurationImpl::Modified() 
#22 cppmicroservices::scrimpl::ComponentConfigurationImpl::ConfigChangedState
#24 cppmicroservices::scrimpl::ConfigurationNotifier::NotifyAllListeners
#26 std::enable_if<is_invocable_r_v<cppmicroservices::cmimpl::ConfigurationAdminImpl::NotifyConfigurationUpdated
```

This change has NOT created a new deadlock potential, rather exposed one

tcormackMW:
The bug here is described below:

There are two services, **serviceA** and **serviceB**. **serviceA** depends on **serviceB**. The two threads in the above comment:

**THREAD 1**
**serviceA** is is being `Activate`d asynchronously because its reference to `serviceB` has become satisfied. In doing its `Activate`ion, it calls `InitializeServicesCache` which initializes the member variable `boundServicesCache` which stores the _referenced_ services that are bound to this service. For each referenced interface, `boundServicesCache`, which is a `Guarded` variable, is locked (**Lock Mutex 2**) and `GetService` is called on the _referenced_ `ServiceObject`. 

In the call to `GetService`, eventually `Activate` is invoked on the `CCActiveState` of **serviceB**, the _referenced_ service. Within the call to `serviceB::CCActiveState::Activate`, we lock the `oneAtAtime` mutex  (**Lock Mutex 1**).

**THREAD 2**
**serviceB** is modified. The `CCActiveState::Modified` method is invoked (**Lock Mutex 1**). As a result, all `ServiceListeners` are alerted to the changed service. **serviceA** is one of those listeners. As a result of the changed property of **serviceB**, **serviceA** becomes unsatisfied and through `CCActiveState::Deactivate` the `SingletonComponentConfiguration::DestroyComponentInstances` is invoked. This method clears the `ComponentContextImpl` associated with the _referencing_ singleton (**serviceA**), invoking the destructor for that object which invokes the destructor of the `Guarded` `boundServicesCache` object of `serviceA`  (**Lock Mutex 2**).

I have created a new issue for this #1192

PR #1119, opened 2025-07-04
clang-tidy lint fixes

vivekkushwaha39:
## Warnings Addressed:

📄ComponentCallbackGenerator.cpp
Line 44: Constructor does not initialize fields: mHeaderIncludes, mComponentInfos
Line 154: Member mHeaderIncludes of type const int is const qualified
Line 155: Member mComponentInfos of type const int is const qualified

### 📄 ComponentInfo.cpp
Line 29: Nested namespaces can be concatenated

### 📄 Main.cpp
Line 51: An exception may be thrown in function main which should not throw exceptions

### 📄 ManifestParser.hpp
Line 47: Function ParseAndGetComponentInfos should be marked [[nodiscard]]

### 📄 TestCodeGenerator.cpp
Line 671: Two adjacent parameters of CodegenInvalidManifestState of similar type (std::string) are easily swapped by mistake

### 📄 Util.hpp
Line 35: Nested namespaces can be concatenated
Line 50: Constructor does not initialize field: jsonVal
Line 71: Constructor does not initialize field: jsonVal
Line 134: Function GetString should be marked [[nodiscard]]

### DIFF
--- a/compendium/tools/SCRCodeGen/ComponentCallbackGenerator.hpp
+++ b/compendium/tools/SCRCodeGen/ComponentCallbackGenerator.hpp
@@ -151,8 +151,8 @@ namespace codegen
             }
         }
 
-        std::vector<std::string> const mHeaderIncludes;
-        std::vector<ComponentInfo> const mComponentInfos;
+        std::vector<std::string> mHeaderIncludes;
+        std::vector<ComponentInfo> mComponentInfos;
         std::stringstream mStrStream;
     };
 
--- a/compendium/tools/SCRCodeGen/Main.cpp
+++ b/compendium/tools/SCRCodeGen/Main.cpp
@@ -53,44 +53,44 @@ main(int argc, char const** argv, char**)
     int returnCode = 0;
     int const FailureReturnCode = -1;
 
-    std::vector<std::string> args(argv, argv + argc);
-
-    // Validate program options ordering.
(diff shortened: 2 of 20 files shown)


compendium/tools/SCRCodeGen/ComponentCallbackGenerator.hpp
```diff
@@ -151,8 +151,8 @@ namespace codegen
             }
         }
 
-        std::vector<std::string> const mHeaderIncludes;
```
tcormackMW, line 154: Why are we removing these const variables?

compendium/tools/SCRCodeGen/Main.cpp
```diff
@@ -53,44 +53,44 @@ main(int argc, char const** argv, char**)
...
-    // Validate if file stream is open. Otherwise, throw
-    auto checkFileOpenOrThrow = [](auto const& fstream)
-    {
-        if (!fstream.is_open())
-        {
-            throw std::runtime_error("Failed to open manifest file");
-        }
-    };
+    	// Validate program options ordering.
+	// Return the iterator corresponding to the option argument.
```
tcormackMW, line 61: some of these are tabs, some are spaces

while this doesn't matter for compilation, depending on the way a given ide is setup, it will make it look super weird and confusing for the developer.

These should all be spaces so that the view is consistent across any ide

PR #1120, opened 2025-07-07
Clang-tidy warnings fix for framework/src/Bundle

ghost:
🛠️ **PR Description**: Fix clang-tidy warnings in framework/src/bundle`.

---

#### 🔁 **Use of range-based for-loops**

* **Replaced** manual iterator loops with range-based loops for better readability and type safety.
* **Files affected**:

  * `Bundle.cpp`
  * `BundlePrivate.cpp`

#### 🧹 **Implicit return simplification**

* **Replaced** explicit return of default-constructed objects with braced-init `{}` for brevity.
* **Examples**:

  * `return std::string();` → `return {};`
  * `return BundleResource();` → `return {};`
* **Files affected**:

  * `BundleArchive.cpp`
  * `BundleContext.cpp`
  * `BundleManifest.cpp`
  * `Constants.cpp`
  * `CoreBundleContext.cpp`

#### 🧱 **Flatten nested namespaces**

* **Converted** nested `namespace` blocks into C++17-style `namespace a::b`.
* **Files affected**:

  * `BundleUtils.cpp`
  * `Constants.cpp`

#### 🛡️ **Constructor modernizations & copy simplifications**

* Used `= default` where copy constructors do not need to be explicitly defined.
* Used `std::move` in constructors to avoid unnecessary copies.
* **Files affected**:

  * `BundleManifest.cpp`
  * `BundleResource.cpp`

#### 🎯 **Consistent use of brace-initialization for return values**

* Unified object return syntax for `return T(args)` to `return {args}`.
* **Example**: `return BundleVersion(version2);` → `return {version2};`
* **Files affected**:

  * `BundleVersion.cpp`
  * `BundleContextPrivate.cpp`
  * `BundleEventInternal.h`

#### 🔄 **Type-correct arithmetic and iterator handling**

* Avoided unsafe pointer arithmetic by explicitly casting with `static_cast<std::ptrdiff_t>`.
* Replaced raw pointer increment with `std::next()` for clarity and portability.
* **Files affected**:

  * `BundleResourceBuffer.cpp`

#### 🧯 **Better exception handling**

* **Handled** unused exceptions explicitly with `(void)ex;` to silence compiler warnings.
* **File affected**:

  * `BundlePrivate.cpp`

---

### 🔍 Rationale:

* 🧵 Reduces verbosity with modern initialization and loop constructs.
* 🛡️ Enhances safety and correctness by removing raw pointer arithmetic.
* ✅ Aligns with C++17 and above.
* 🔧 Lowers friction for future static analysis or clang-tidy passes.

---

### 📦 Scope:

* Changes are **non-functional**, purely **code style, cleanup, and modernization**.
* All modifications are limited to the `framework/src/bundle` directory.
* **No logic or behavior** has been altered. These are safe refactors.

---


framework/src/bundle/BundlePrivate.cpp
```diff
@@ -826,23 +826,24 @@ namespace cppmicroservices
             {
                 sr.Unregister();
             }
-            catch (std::logic_error const& /*ignore*/)
+            catch (std::logic_error const& ex/*ignore*/)
```
tcormackMW, line 829: remove "/*ignore*/" if we are no longer ignoring the value

framework/src/bundle/BundleManifest.cpp
```diff
@@ -165,7 +165,7 @@ namespace cppmicroservices
 
     BundleManifest::BundleManifest() : m_Headers(AnyMap::UNORDERED_MAP_CASEINSENSITIVE_KEYS) {}
 
-    BundleManifest::BundleManifest(AnyMap const& m) : m_Headers(m) {}
+    BundleManifest::BundleManifest(AnyMap m) : m_Headers(std::move(m)) {}
```
jeffdiclemente, line 168: I don't think this should be moved from based on the usage in BundlePrivate.cpp. This `BundleManifest` constructor is currently used in the `BundlePrivate::BundlePrivate(CoreBundleContext* coreCtx, std::shared_ptr<BundleArchive> const& ba)` constructor.

framework/src/bundle/BundlePrivate.cpp
jeffdiclemente, line 829: its odd to me that the linter flagged `catch (std::logic_error const& /*ignore*/)`. Isn't this a best practice when not using the exception object in the catch clause?

PR #1137, opened 2025-07-21
701  Ensure that bundle.Start() after the framework has stopped throws (#979)

RobertLauferElektrobit:
Ensure that bundle.Start() after the framework has stopped throws #979

According to the OSGi [spec](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.api.html#org.osgi.framework.Bundle.start-int-), we must throw if the framework has been stopped when bundle.Start() is called.

cherry-pick of [d4f63c3](https://github.com/CppMicroServices/CppMicroServices/commit/d4f63c3119d0a95e564e487c9792ba45fa49f849)

see discussion #701

### DIFF
--- a/framework/src/bundle/BundlePrivate.cpp
+++ b/framework/src/bundle/BundlePrivate.cpp
@@ -76,7 +76,7 @@ namespace cppmicroservices
 
             if (state == Bundle::STATE_UNINSTALLED)
             {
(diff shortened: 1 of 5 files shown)


RobertLauferElektrobit:
not building - thus closing

PR #1193, opened 2025-08-07
Ensure `serviceReference` object properly manages interface on assignment

tcormackMW:
When assigning a `ServiceReference<T>` to a `ServiceReferenceBase`, rather than getting into the copy constructor in `ServieReference` which properly managed the `interfaceID` object, we were getting into the `ServiceReferenceBase::operator=` which ignored the `interfaceID` object. This led to situations where a `ServiceReference<T>` didn't know that it was of type `T` and when retrieving the service in the `getService` call, it would default to the first service in its interfaceMap rather than the proper one

### DIFF
--- a/framework/test/gtest/ServiceReferenceTest.cpp
+++ b/framework/test/gtest/ServiceReferenceTest.cpp
@@ -79,6 +79,7 @@ class ServiceReferenceTest : public ::testing::Test
     SetUp() override
     {
         framework.Start();
+        context = framework.GetBundleContext();
     }
 
     void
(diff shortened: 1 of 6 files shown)


framework/test/gtest/ServiceReferenceTest.cpp
```diff
@@ -88,8 +89,27 @@ class ServiceReferenceTest : public ::testing::Test
...
+TEST_F(ServiceReferenceTest, TestServiceReferenceAssignmentAndCopy)
+{
+    auto s1 = std::make_shared<TestServiceA>();
+
+    ServiceRegistrationBase reg1 = context.RegisterService<ServiceNS::ITestServiceA>(s1);
+    ServiceReferenceBase ref1 = reg1.GetReference();
+
+    ASSERT_EQ(ref1.GetInterfaceId(), "");
+    ServiceReference<ServiceNS::ITestServiceA> typedRefAssign;
+    ServiceReference<ServiceNS::ITestServiceA> typedRefCopy = ref1;
```
Burgch, line 105: Perhaps:
```suggestion
    ServiceReference<ServiceNS::ITestServiceA> typedRefCopy{ref1};
```
to guarantee the copy constructor is always called, instead of copy assignment to a default constructed object?

Burgch (approved):
Great catch Toby, nice work!

jeffdiclemente (approved):

framework/test/gtest/ServiceReferenceTest.cpp
tcormackMW: we created changes here to make sure the copy-assignment was doing the correct thing, which is what this test is trying to verify. 

If we did the constructor then we are just testing whether the already functional behavior worked right?

@Burgch

PR #1195, opened 2025-08-12
Fix ambiguous overload for C++20 build

ShivamNegi:


tcormackMW (approved):
lgtm

framework/include/cppmicroservices/ServiceReference.h
```diff
@@ -102,7 +102,13 @@ namespace cppmicroservices
...
+#if __cplusplus >= 202002L
+	bool operator==(ServiceReference const& reference) const {
+		return ServiceReferenceBase::operator==(reference);
+	}
+#else
         using ServiceReferenceBase::operator==;
+#endif
 
         using ServiceReferenceBase::Hash;
     };
```
tcormackMW, line 114: @ShivamNegi do we need to do this as well in the ServiceReference<void> 30 lines below this change? see line 142 of this file

ShivamNegi: Added the second change too.

PR #1197, opened 2025-08-21
Enhance bind logs

carneyweb:
Updated log messages with information about the service getting the binding added

### DIFF
--- a/compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
+++ b/compendium/DeclarativeServices/src/manager/BindingPolicy.cpp
@@ -109,17 +109,17 @@ namespace cppmicroservices
                         if (!boundRefsHandle->empty() && mgr.IsUnary())
                         {
                             svcRefToBind = *(boundRefsHandle->begin());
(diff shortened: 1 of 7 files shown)


jeffdiclemente (approved):

tcormackMW:
> > Looks good. I think it is possible we may want to clarify what the configName is... given it is an impl class, maybe we want to say "Service with impl class "xxxx" has been ..."
> > I'm not sure what is right. i think the extra words i suggested will make it more verbose, perhaps to a fault, but I could imagine someone being confused by service identification using impl class
> 
> I like the idea of giving more info, but I'm not sure it's right in this case. The value of configName is not always the implementation class, but can also be the name of the component if specified in the manifest, and I don't know of a way to differentiate the two cases in that section of code.

fair enough. This seems fine to me. People are going to find these logs using the identifier they know about. If it is impl class that they know, that's what they will search, but if they actively supplied a name to the service (which is rare in my experience), that's what they will search.... In either case they'll find the logs.

tcormackMW (approved):
lgtm

PR #1198, opened 2025-08-28
Fix sporadic test failure with bundle trackers

tcormackMW:


framework/include/cppmicroservices/detail/BundleAbstractTracked.hpp
```diff
@@ -129,6 +130,7 @@ namespace cppmicroservices
                         return;
                     }
                     adding.push_back(item); /* mark this item is being added */
+                    previouslyAdded.insert(item);
```
jeffdiclemente, line 133: doesn't the item also need to be removed from `previouslyAdded` when someone wants to stop tracking the item?

tcormackMW: No I don't think so. This list is about whether the trackInitial method has been called on a given bundle. Once it has, that bundle should never be trackInitial'd again

tcormackMW: More accurately, this list is for bundles that have had a callback invoked on them so trackInitial does not need to be invoked on it again ever

Burgch, line 133: Do we need to keep doing this forever? Perhaps `TrackInitial` could set a boolean and clear `previouslyAdded`, and then here, if that boolean is set, we don't need to do the insert (and thus can prevent this set from growing indefinitely)?

tcormackMW: ah this was already merged. I'll make a new issue to fix this.

PR #1199, opened 2025-08-29
silence warning for use of std::atomic_load for use with std::shared_ptr.

tcormackMW:
This _should_ be supported in c++20 but as we still support 17, we are going to silence it for now.

### DIFF
--- a/framework/include/cppmicroservices/detail/Threads.h
+++ b/framework/include/cppmicroservices/detail/Threads.h
@@ -23,10 +23,6 @@
 #ifndef CPPMICROSERVICES_THREADS_H
 #define CPPMICROSERVICES_THREADS_H
 
-#if defined(__cplusplus) && __cplusplus >= 202002L
-#define _SILENCE_CXX20_OLD_SHARED_PTR_ATOMIC_SUPPORT_DEPRECATION_WARNING
-#endif
-
 #include "cppmicroservices/FrameworkConfig.h"
 
 #include <atomic>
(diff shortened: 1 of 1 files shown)


Burgch (approved):

framework/include/cppmicroservices/detail/Threads.h
```diff
@@ -232,13 +232,16 @@ namespace cppmicroservices
...
 
-#if !defined(__GNUC__) || __GNUC__ > 4
-        // The std::atomic_load() et.al. overloads for std::shared_ptr are only available
-        // in libstdc++ since GCC 5.0. Visual Studio 2013 has it, but the Clang version
-        // is unknown so far.
 
-        // Specialize cppmicroservices::Atomic for std::shared_ptr to use the standard library atomic
-        // functions:
+// Check for C++17 atomic shared_ptr support
+#if defined(__cpp_lib_atomic_shared_ptr) && __cpp_lib_atomic_shared_ptr >= 201711L
```
jeffdiclemente: `__cpp_lib_atomic_shared_ptr` [is supported in C++20](https://en.cppreference.com/w/cpp/utility/feature_test.html) and this repo doesn't support C++20 officially.
I am ok with supporting C++20 in the code base and if we decide to, we need to make it official within the build infrastructure and docs.

framework/include/cppmicroservices/detail/Threads.h
```diff
@@ -232,13 +232,16 @@ namespace cppmicroservices
...
+        template <typename T>
+        class Atomic<std::shared_ptr<T>> : public std::atomic<std::shared_ptr<T>>
```
jeffdiclemente: I don't think this will work when compiled with support for `std::atomic<std::shared_ptr<T>>` since all usage in this code base is using method names that are not the same as those of `std::atomic<std::shared_ptr<T>>`

Burgch: Right - instead of inheriting from `std::atomic<std::shared_ptr<T>>`, this could use composition and act as a simple wrapper around `std::atomic<std::shared_ptr<T>>` instead?

tcormackMW: Updated it, I think I captured what you were asking for.

PR #1202, opened 2025-09-09
Ensure synchronous config updates per-config

tcormackMW:


compendium/ConfigurationAdmin/src/ConfigurationImpl.hpp
```diff
@@ -190,6 +201,7 @@ namespace cppmicroservices
             }
 
           private:
+            std::shared_ptr<AsyncWorkService> strand { nullptr };
```
Burgch: This will be initialised to `nullptr` by default :)

Burgch (approved):
Oh wow, I had no idea you were even working on this - it looks great, thank you very much Toby!

compendium/ConfigurationAdmin/src/CMAsyncWorkService.cpp
```diff
@@ -83,20 +86,51 @@ namespace cppmicroservices
...
             std::shared_ptr<cppmicroservices::logservice::LogService> logger;
+
+            // Inner class for strand-based async work service
+            class StrandAsyncWorkService : public AsyncWorkService
+            {
+              public:
+                StrandAsyncWorkService(std::shared_ptr<Strand> strand_,
+                                       std::shared_ptr<cppmicroservices::logservice::LogService> logger_)
+                    : strand(std::move(strand_))
+                    , logger(std::move(logger_))
```
Burgch: This doesn't seem to be used - is it needed?

Burgch (approved):
Very nice - the test looks great!

compendium/ConfigurationAdmin/src/CMAsyncWorkService.cpp
```diff
@@ -83,20 +86,51 @@ namespace cppmicroservices
...
-                    using Result = boost::asio::async_result<decltype(task), Sig>;
-                    using Handler = typename Result::completion_handler_type;
-
-                    Handler handler(std::forward<decltype(task)>(task));
+                    boost::asio::post(threadpool->get_executor(), std::move(task));
+                }
+            }
 
-                    boost::asio::post(threadpool->get_executor(),
-                                      [handler = std::move(handler)]() mutable { handler(); });
```
Burgch, line 93: Ah - these lines are key - instead of posting the `task` directly, you need to wrap it up into an object which doesn't have `get_future()` or the `boost::asio` internals will call that `get_future` to try and return the future from the call to `post`. I think you can still do away with the `using` declarations and use something like:
```
boost::asio::post(threadpool->get_executor(),
                                      [task = std::move(task)]() mutable { task(); });
```

compendium/ConfigurationAdmin/src/CMAsyncWorkService.cpp
```diff
@@ -83,20 +86,51 @@ namespace cppmicroservices
...
+                    , logger(std::move(logger_))
+                {
+                }
+
+                void
+                post(std::packaged_task<void()>&& task) override
+                {
+                    if (strand)
+                    {
+                        boost::asio::post(*strand, std::move(task));
```
Burgch: Ditto here - wrap the `task` in a lambda to stop `asio` from trying to obtain the future?

Burgch (approved):
The test failures made me curious - should be a simple fix I think :)

compendium/ConfigurationAdmin/src/CMAsyncWorkService.cpp
tcormackMW: rats. i was trying to simplify the post because it seemed verbose compared to what I found online, but clearly I over-simplified. Thanks for the clarification

compendium/ConfigurationAdmin/src/CMAsyncWorkService.cpp
tcormackMW: Seemingly not. Oops

PR #1206, opened 2025-09-22
Create custom error for invalid bundleContext

tcormackMW:
Create custom error for easy catch logic for invalid bundle Contexts. As this is a common occurrence, adding this functionality helps to clarify client code.

### DIFF
--- a/framework/include/cppmicroservices/BundleContextException.h
+++ b/framework/include/cppmicroservices/BundleContextException.h
@@ -1,0 +1,52 @@
+/*=============================================================================
+
+  Library: CppMicroServices
+
+  Copyright (c) The CppMicroServices developers. See the COPYRIGHT
+  file at the top-level directory of this distribution and at
+  https://github.com/CppMicroServices/CppMicroServices/COPYRIGHT .
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


framework/include/cppmicroservices/BundleContextException.h
```diff
@@ -0,0 +1,52 @@
...
+// https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4275
+#ifdef _MSC_VER
+#    pragma warning(push)
+#    pragma warning(disable : 4275)
+#endif
+
+namespace cppmicroservices
+{
+
+    class US_Framework_EXPORT BundleContextException final : public std::runtime_error
```
jeffdiclemente: I recommend following the OSGi spec for this. [It specs out](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.api.html#org.osgi.framework.BundleContext) that an `IllegalStateException` (which is part of the Java language) should be thrown when the `BundleContext` is invalid.  I would just change this class' name to `IllegalStateException`.

jeffdiclemente (changes requested):

framework/include/cppmicroservices/IllegalStateException.h
```diff
@@ -0,0 +1,52 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+=============================================================================*/
+
+#ifndef CPPMICROSERVICES_ILLEGALSTATEEXCEPTION_H
+#define CPPMICROSERVICES_ILLEGALSTATEEXCEPTION_H
+
+#include "cppmicroservices/Bundle.h"
```
jeffdiclemente: I don't think this include is required. can it be removed?

mwmcarney (commented):
LGTM!

PR #1208, opened 2025-09-26
Verify that a singleton's instance is valid before invoking unbind

tcormackMW:
This is to ensure that we don't invoke a method on a nullptr if the `componentInstance` has already been reset.

### DIFF
--- a/framework/src/service/ServiceReferenceBase.cpp
+++ b/framework/src/service/ServiceReferenceBase.cpp
@@ -143,6 +143,10 @@ namespace cppmicroservices
             // if they are valid, we are less
             return ref->coreInfo != nullptr;
         }
+
+        if (!ref->coreInfo){
+            return false;
+        }
         /// A deadlock caused by mutex order locking will happen if these two scoped blocks
         /// are combined into one. Multiple threads can enter this function as a result of
         /// adding/removing ServiceReferenceBase objects from STL containers. If that occurs
--- a/framework/test/gtest/ServiceReferenceTest.cpp
+++ b/framework/test/gtest/ServiceReferenceTest.cpp
@@ -290,3 +290,115 @@ TEST_F(ServiceReferenceTest, TestRegisterUnregisterCompareReference)
 
     ASSERT_NE(set.find(sr2), set.end());
 }
+
+TEST_F(ServiceReferenceTest, TestEqualityOperator)
+{
+    auto context = framework.GetBundleContext();
+
+    auto impl = std::make_shared<TestServiceA>();
+    auto reg = context.RegisterService<ServiceNS::ITestServiceA>(impl);
+
+    auto sr1 = reg.GetReference();
+    auto sr2 = reg.GetReference();
+
+    // Should be equal: both refer to the same service
+    ASSERT_TRUE(sr1 == sr2);
+
+    // Different service
+    auto implB = std::make_shared<TestServiceB>();
+    auto regB = context.RegisterService<ServiceNS::ITestServiceB>(implB);
+    auto srB = regB.GetReference();
+
+    ASSERT_FALSE(sr1 == srB);
+
+    // After unregistration, references to different services should still not be equal
+    reg.Unregister();
+    regB.Unregister();
+    ASSERT_FALSE(sr1 == srB);
+
+    // References to the same (now unregistered) service should still compare equal
+    ASSERT_TRUE(sr1 == sr2);
+}
+
+TEST_F(ServiceReferenceTest, TestLessThanOperator_Ranking)
+{
+    auto context = framework.GetBundleContext();
+
+    // Register first service with ranking 10
+    ServiceProperties propsA;
+    propsA[Constants::SERVICE_RANKING] = 10;
+    auto implA = std::make_shared<TestServiceA>();
+    auto regA = context.RegisterService<ServiceNS::ITestServiceA>(implA, propsA);
+    auto srA = regA.GetReference();
+
+    // Register second service with ranking 20
+    ServiceProperties propsB;
+    propsB[Constants::SERVICE_RANKING] = 20;
+    auto implB = std::make_shared<TestServiceA>();
+    auto regB = context.RegisterService<ServiceNS::ITestServiceA>(implB, propsB);
+    auto srB = regB.GetReference();
+
+    // Lower ranking is less
+    ASSERT_TRUE(srA < srB);
+    ASSERT_FALSE(srB < srA);
+
+    // Clean up
+    regA.Unregister();
+    regB.Unregister();
+
+    ASSERT_TRUE(srA < srB);
+    ASSERT_FALSE(srB < srA);
+}
+
+TEST_F(ServiceReferenceTest, TestLessThanOperator_SameRanking_DifferentIDs)
+{
+    auto context = framework.GetBundleContext();
+
+    // Register first service (will have lower ID)
+    auto implA = std::make_shared<TestServiceA>();
+    auto regA = context.RegisterService<ServiceNS::ITestServiceA>(implA);
+    auto srA = regA.GetReference();
+
+    // Register second service (will have higher ID)
+    auto implB = std::make_shared<TestServiceA>();
+    auto regB = context.RegisterService<ServiceNS::ITestServiceA>(implB);
+    auto srB = regB.GetReference();
+    // Same ranking (default), so compare by ID
+    // Higher ID is less than lower ID
+    ASSERT_TRUE(srB < srA);
+    ASSERT_FALSE(srA < srB);
+
+    // Clean up
+    regA.Unregister();
+    regB.Unregister();
+}
+
+TEST_F(ServiceReferenceTest, TestLessThanOperator_InvalidReferences)
+{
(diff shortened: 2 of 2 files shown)


compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.cpp
```diff
@@ -224,7 +224,14 @@ namespace cppmicroservices
             }
             try
             {
-                GetComponentInstance()->InvokeBindMethod(refName, svcToBind);
+                if (auto inst = GetComponentInstance(); inst)
```
Burgch, line 233: I believe `GetComponentContext()` can also be impacted by the same race condition - ought we protect the two usages of that too?

tcormackMW:
Wrote a test case to try and get `rebind` and `bundle.stop` to happen at the same time  and maybe reset the `instance` pointer... After 40k runs, I didn't see it happen. 

I am not sure if this test case is bad, or if I have misidentified the problem context

@Burgch does this test conform with your understanding of the problematic situation or have I missed something?

compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.cpp
```diff
@@ -224,15 +228,23 @@ namespace cppmicroservices
...
             }
-            catch (std::exception const&)
+            catch (std::exception const& ex)
             {
                 GetLogger()->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
                                  "Exception received from user code while binding "
                                  "service reference"
                                      + refName + ".",
                                  std::current_exception());
+                std::cout << ex.what() << std::endl;
```
Burgch: Remove?

compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.cpp
```diff
@@ -241,18 +253,30 @@ namespace cppmicroservices
...
             }
-            catch (std::exception const&)
+            catch (std::exception const& ex)
             {
                 GetLogger()->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
                                  "Exception received from user code while unbinding "
                                  "service reference"
                                      + refName + ".",
                                  std::current_exception());
+                std::cout << ex.what() << std::endl;
```
Burgch: Remove?

compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.cpp
```diff
@@ -241,18 +253,30 @@ namespace cppmicroservices
...
+                {
+                    inst->InvokeUnbindMethod(refName, svcToUnbind);
+                }
+                else
+                {
+                    throw std::runtime_error("Invalid Singleton Component Instance, likely destroyed");
+                }
             }
-            catch (std::exception const&)
+            catch (std::exception const& ex)
```
Burgch: Can go back to unnamed?

compendium/DeclarativeServices/src/manager/SingletonComponentConfiguration.cpp
```diff
@@ -224,15 +228,23 @@ namespace cppmicroservices
...
+                {
+                    inst->InvokeBindMethod(refName, svcToBind);
+                }
+                else
+                {
+                    throw std::runtime_error("Invalid Singleton Component Instance, likely destroyed");
+                }
             }
-            catch (std::exception const&)
+            catch (std::exception const& ex)
```
Burgch: Can go back to unnamed?

Burgch (changes requested):
Thanks Toby, looks good!

compendium/DeclarativeServices/test/gtest/TestBindingPolicies.cpp
```diff
@@ -1329,5 +1330,68 @@ namespace cppmicroservices
...
+        TEST_F(BindingPolicyTest, TestRebindConcurrentWithBundleStop)
+        {
+            auto bc = GetFramework().GetBundleContext();
+
+            // make sure there is concurrency
+            auto aws = std::make_shared<test::AsyncWorkServiceThreadPool>(20);
+            auto reg = bc.RegisterService<cppmicroservices::async::AsyncWorkService>(aws);
+
+            test::InstallAndStartDS(bc);
+            auto testBundle = test::InstallAndStartBundle(bc, "TestBundleDSTOI5");
```
Burgch: I think this needs to be a bundle with a `greedy` reference?

Burgch:
> Wrote a test case to try and get `rebind` and `bundle.stop` to happen at the same time and maybe reset the `instance` pointer... After 40k runs, I didn't see it happen.
> 
> I am not sure if this test case is bad, or if I have misidentified the problem context
> 
> @Burgch does this test conform with your understanding of the problematic situation or have I missed something?

Sorry - missed this! I'm not sure `Rebind` is being called because the reference isn't `greedy` ?

compendium/DeclarativeServices/test/gtest/TestBindingPolicies.cpp
```diff
@@ -1329,5 +1330,65 @@ namespace cppmicroservices
...
+                                            reg.Unregister();
+                                        }
+                                    });
+
+            auto bun = std::async(std::launch::async,
+                                  [&testBundle, &sync_point]()
+                                  {
+                                      sync_point.Wait(); // Wait for all threads to reach this point
+                                      testBundle.Stop();
+                                  });
```
Burgch, line 1394: Yep, this was exactly what I had in mind. I guess to increase the chances of a collision even more, you could use a second barrier which the stop thread waits on after the `sync_point.Wait()`, and release that barrier after you trigger the first `Unregister`, and then we know we're in the process of unregistering whilst the bundle is stopping, vs. the bundle stop getting in first and completing before the first Unregister even gets to trigger anything?

framework/src/service/ServiceReferenceBase.cpp
```diff
@@ -54,7 +54,11 @@ namespace cppmicroservices
         d.Load()->interfaceId = interfaceId;
     }
 
-    ServiceReferenceBase::operator bool() const { return static_cast<bool>(GetBundle()); }
+    ServiceReferenceBase::
+    operator bool() const
```
Burgch: Funky formating?

framework/src/service/ServiceReferenceBase.cpp
```diff
@@ -132,7 +141,14 @@ namespace cppmicroservices
...
         {
-            return true;
+            if (reference)
+            {
+                return true;
+            }
+            else
+            {
+                return false;
+            }
```
Burgch: `return reference`? Or `return static_cast<bool>(reference)` if you want to be explicit about it using the bool coversion?

compendium/DeclarativeServices/test/gtest/TestBindingPolicies.cpp
tcormackMW: yeah i can do that, been running that for a while and no repeat of the failure

compendium/DeclarativeServices/test/gtest/TestBindingPolicies.cpp
```diff
@@ -1329,5 +1330,70 @@ namespace cppmicroservices
...
+                    bc.RegisterService<test::Interface1>(std::make_shared<MockImpl>(),
+                                                         ServiceProperties({
+                                                             { Constants::SERVICE_RANKING, Any(i) }
+                })));
+            }
+
+            auto testBundle = test::InstallAndStartBundle(bc, "TestBundleDSTOI5");
+
+            Barrier sync_point(2); // 2 threads to synchronize
+            Barrier sync_point1(2); // 2 threads to synchronize
```
jeffdiclemente: nit-pick: more descriptive variable names would be helpful

framework/src/service/ServiceReferenceBase.cpp
```diff
@@ -130,22 +133,16 @@ namespace cppmicroservices
...
-            return false;
-        }
 
+        // if we don't have core info, this is an invalid serviceReference which was never valid
+        // should be safe and consistent across the process to use this as a mechanism of comparison
+        if (!self->coreInfo){
+            // if we are invalid, return whether they are valid.
+            // if they are valid, we are less
+            return ref->coreInfo != nullptr;
+        }
```
Burgch, line 145: 👍 Is it possible to lock this down with a test?

framework/src/service/ServiceReferenceBase.cpp
```diff
@@ -195,7 +192,9 @@ namespace cppmicroservices
     bool
     ServiceReferenceBase::operator==(ServiceReferenceBase const& reference) const
     {
-        return d.Load()->registration.lock() == reference.d.Load()->registration.lock();
+        // because core info stays valid even after unregistration, this guarantees that two references to two
+        // different, but both unregistered, services do NOT evaluate to be the same
+        return d.Load()->coreInfo == reference.d.Load()->coreInfo;
```
Burgch, line 201: Likewise, can this be locked down with a test?

PR #1209, opened 2025-10-02
Solve issue where many threads install the same bundles concurrently

tcormackMW:
We were not locking properly and were getting off by one errors for the file path and the actual installed bundle (alphabetically because it's using a multimap in BundleRegistry).

This fixes that critical section. 

The new test fails ~80-90% of the time without this fix, but has not yet failed with the fix.

### DIFF
--- a/compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
+++ b/compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
@@ -487,7 +487,7 @@ namespace test
         {
             futures.emplace_back(
                 std::async(std::launch::async,
-                           [&, i]()
+                           [&]()
                            {
                                b.Wait();
                                auto threadCtx = framework.GetBundleContext();


jeffdiclemente (approved):
excellent work. can a test be written to catch this in the future? I don't think you need to physically create thousands of bundles; I'm hoping you can dynamically generate thousands of bundle manifests in memory and then call

`std::vector<Bundle>
    BundleRegistry::Install(std::string const& location,
                            BundlePrivate* installingBundle,
                            cppmicroservices::AnyMap const& bundleManifest)`
passing in the generated bundle manifest as `bundleManifest`

mwmcarney (approved):
LGTM. Well done!

compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
```diff
@@ -469,4 +471,69 @@ namespace test
...
+                                       if (it != installedBundles.end())
+                                       {
+                                           std::cout << "Bundle '" << name << "' already installed from '" << it->second
+                                                     << "'. New install attempt from '" << path << "'." << std::endl;
+                                           failed.store(true);
+                                       }
+                                       else
+                                       {
+                                           installedBundles[name] = path;
+                                       }
```
Burgch: `auto const [preexistingBundle, newBundleWasInserted] = installedBundles.try_emplace(name, path)` perhaps?

Burgch (approved):
Very nice, this is so much cleaner and easier to follow as well!

compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
tcormackMW: who knew. I have updated the code. I like that `try_emplace` method

Burgch (approved):
Thanks!

Issue #1218, opened 2025-10-30
`SCRExtensionRegistry` and `ComponentRegistry` index stored objects using `bundleIds` which can disappear as the bundle dies. Causes orphaned data


tcormackMW:

tcormackMW:
For example:

```
libCppMicroServicesd.so.3.8.8!cppmicroservices::Bundle::GetBundleId framework/src/bundle/Bundle.cpp:174
libDeclarativeServicesd.so!cppmicroservices::scrimpl::ComponentManagerImpl::GetBundleId compendium/DeclarativeServices/src/manager/ComponentManagerImpl.hpp:122
libDeclarativeServicesd.so!cppmicroservices::scrimpl::ComponentRegistry::RemoveComponentManager compendium/DeclarativeServices/src/ComponentRegistry.cpp:81
libDeclarativeServicesd.so!cppmicroservices::scrimpl::SCRBundleExtension::DisableAndRemoveAllComponentManagers compendium/DeclarativeServices/src/SCRBundleExtension.cpp:142
libDeclarativeServicesd.so!cppmicroservices::scrimpl::SCRBundleExtension::~SCRBundleExtension compendium/DeclarativeServices/src/SCRBundleExtension.cpp:119
libDeclarativeServicesd.so!std::shared_ptr<cppmicroservices::scrimpl::SCRBundleExtension>::~shared_ptr
libDeclarativeServicesd.so!cppmicroservices::scrimpl::ComponentFactoryImpl::CreateFactoryComponent compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp:171
libDeclarativeServicesd.so!cppmicroservices::scrimpl::ConfigurationNotifier::AnyListenersForPid compendium/DeclarativeServices/src/manager/ConfigurationNotifier.cpp:178
libDeclarativeServicesd.so!cppmicroservices::service::cm::ConfigurationListenerImpl::configurationEvent
cppmicroservices::cmimpl::ConfigurationAdminImpl::NotifyConfigurationUpdated
```

In this instance, nothing really bad happens because it just throws nad the `scrBundleExtension` destructor (for better or worse) ignores this throw. 

But in other cases, like the `ComponentRegistry`:

```
ComponentRegistry::AddComponentManager(std::shared_ptr<ComponentManager> const& cm)
    {
        std::lock_guard<std::mutex> lock(mMapsMutex);
        auto result = mComponentsByName.insert(
[...]

PR #1219, opened 2025-10-30
Fix issue with concurrent creation of `configuration`s to trigger `factoryInstance` and `bundle.stop`

tcormackMW:
This test failed ~100% of the time before the fix, with the ComponentManager throwing an uncaught exception because the `bundleContext` was bad in its constructor

I ran it 5000 times after and it didn't fail. That being said, this feels a bit like a bandaid. I made this issue: https://github.com/CppMicroServices/CppMicroServices/issues/1218 because I feel like there are a lot of places where we rely on `bundleId` to index data. And as soon as the bundle is no longer good, that indexing is no longer accessible. I feel like maybe `bundle` objects and `bundlePrivates` should maybe both own the same `std::shared_ptr<const long> id` and then the id is alive and is persistent for a given `bundle` object

Not that the above issue means this isn't a good patch, but I do think we maybe need to delve into the lifetime of the objects we use as keys in key-value structures

### DIFF
--- a/compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
+++ b/compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
@@ -179,10 +179,10 @@ namespace cppmicroservices::scrimpl
         {
             throw;
         }
-        catch (...)
+        catch (std::exception const& e)
         {
             logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
-                        "Failed to create ComponentManager with name " + newMetadata->name);
+                        "Failed to create ComponentManager with name " + newMetadata->name + ": " + e.what());
         }
     }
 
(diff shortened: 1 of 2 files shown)


jeffdiclemente (approved):

compendium/DeclarativeServices/src/manager/ComponentFactoryImpl.cpp
```diff
@@ -179,6 +179,11 @@ namespace cppmicroservices::scrimpl
         {
             throw;
         }
+        catch (...)
+        {
+            logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_ERROR,
+                        "Failed to create ComponentManager with name " + newMetadata->name);
```
Burgch: Might we be able to print the exception to the log too?

Burgch (approved):
Thanks Toby!

PR #1221, opened 2025-11-21
Increase floating point precision for anys

tcormackMW:
Increase floating precision from default (6) to 12.

### DIFF
--- a/framework/include/cppmicroservices/Any.h
+++ b/framework/include/cppmicroservices/Any.h
@@ -191,7 +191,7 @@ namespace cppmicroservices
     any_value_to_json(std::ostream& os, T const& val, uint8_t const, int32_t const)
     {
         detail::ScopeGuard sg([&os, oldPrec = os.precision()]() { os.precision(oldPrec); });
-        os << std::setprecision(12) << val; // up to 12 significant digits, no trailing zeros
+        os << std::setprecision(DOUBLE_PRECISION) << val;
         return os;
     }
 
@@ -200,7 +200,7 @@ namespace cppmicroservices
     any_value_to_string(std::ostream& os, T const& val)
     {
         detail::ScopeGuard sg([&os, oldPrec = os.precision()]() { os.precision(oldPrec); });
-        os << std::setprecision(12) << val;
+        os << std::setprecision(DOUBLE_PRECISION) << val;
         return os;
     }
 
@@ -209,7 +209,7 @@ namespace cppmicroservices
     any_value_to_cpp(std::ostream& os, T const& val, uint8_t const, int32_t const)
     {
         detail::ScopeGuard sg([&os, oldPrec = os.precision()]() { os.precision(oldPrec); });
-        os << std::setprecision(12) << val;
+        os << std::setprecision(DOUBLE_PRECISION) << val;
         return os;
     }
     /**
--- a/framework/include/cppmicroservices/Any.h
+++ b/framework/include/cppmicroservices/Any.h
@@ -34,6 +34,7 @@ DEALINGS IN THE SOFTWARE.
 #define CPPMICROSERVICES_ANY_H
 
 #include "cppmicroservices/FrameworkConfig.h"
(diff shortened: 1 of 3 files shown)


framework/include/cppmicroservices/Any.h
```diff
@@ -156,28 +158,76 @@ namespace cppmicroservices
     template <typename ValueType>
     ValueType const& ref_any_cast(Any const& operand);
 
+    // Guard to restore stream state
+    struct OStreamFormatGuard
```
mwmcarney: Couldn't this just be a standard scope guard?

framework/include/cppmicroservices/Any.h
```diff
@@ -156,28 +158,76 @@ namespace cppmicroservices
...
+        return os;
+    }
+
+    template <class T>
+    std::enable_if_t<std::is_floating_point_v<T>, std::ostream&>
+    any_value_to_string(std::ostream& os, T const& val)
+    {
+        OStreamFormatGuard g(os);
+        os.unsetf(std::ios::floatfield);
+        os << std::setprecision(12) << val;
```
jeffdiclemente: did you mean to use `DOUBLE_PRECISION` here instead of `12`?

framework/include/cppmicroservices/Any.h
```diff
@@ -156,28 +158,76 @@ namespace cppmicroservices
...
         return os << val;
     }
 
+    template <class T>
+    std::enable_if_t<std::is_floating_point_v<T>, std::ostream&>
+    any_value_to_json(std::ostream& os, T const& val, uint8_t const, int32_t const)
+    {
+        OStreamFormatGuard g(os);
+        os.unsetf(std::ios::floatfield);    // defaultfloat semantics
+        os << std::setprecision(12) << val; // up to 12 significant digits, no trailing zeros
```
jeffdiclemente: did you mean to use `DOUBLE_PRECISION` here instead of `12`?

framework/include/cppmicroservices/Any.h
```diff
@@ -156,28 +158,76 @@ namespace cppmicroservices
...
+        return os;
+    }
+
+    template <class T>
+    std::enable_if_t<std::is_floating_point_v<T>, std::ostream&>
+    any_value_to_cpp(std::ostream& os, T const& val, uint8_t const, int32_t const)
+    {
+        OStreamFormatGuard g(os);
+        os.unsetf(std::ios::floatfield);
+        os << std::setprecision(12) << val;
```
jeffdiclemente: did you mean to use `DOUBLE_PRECISION` here instead of `12`?

framework/include/cppmicroservices/Any.h
tcormackMW: yup

jeffdiclemente (commented):
what's the driving use case for changing from 6 to 12?

tcormackMW:
> what's the driving use case for changing from 6 to 12?

This is done as a result of requests that floating point values have increased precision in our any implementation.

mwmcarney (approved):
LGTM!

framework/include/cppmicroservices/Any.h
```diff
@@ -156,28 +159,59 @@ namespace cppmicroservices
...
     any_value_to_cpp(std::ostream& os, T const& val, uint8_t const = 0, int32_t const = 0)
     {
         return os << val;
     }
 
+    template <class T>
+    std::enable_if_t<std::is_floating_point_v<T>, std::ostream&>
+    any_value_to_json(std::ostream& os, T const& val, uint8_t const, int32_t const)
+    {
+        detail::ScopeGuard sg([&os, oldPrec = os.precision()]() { os.precision(oldPrec); });
```
dleffingwell, line 194: Since this implementation is more complex than the default implementation and it is duplicated three times, can we have a helper function for the body of these three.

dleffingwell (approved):

PR #1223, opened 2025-12-09
update serviceTracker code flow to eliminate extraneous throws

tcormackMW:
This causes hotspots in profiling code. Using an invalid `ServiceReference` instead is faster and cleaner.

### DIFF
--- a/framework/include/cppmicroservices/ServiceTracker.h
+++ b/framework/include/cppmicroservices/ServiceTracker.h
@@ -242,7 +242,7 @@ namespace cppmicroservices
          * <p>
          * This implementation calls GetService() to determine if a service
          * is being tracked.
-         * 
+         *
          * @throws std::logic_error If the <code>BundleContext</code>
          *         with which this <code>ServiceTracker</code> was created is no
          *         longer valid or became invalid while waiting on Service.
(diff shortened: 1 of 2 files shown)


framework/include/cppmicroservices/detail/ServiceTracker.hpp
```diff
@@ -331,7 +331,7 @@ namespace cppmicroservices
         std::size_t length = references.size();
         if (length == 0)
         { /* if no service is being tracked */
-            throw ServiceException("No service is being tracked");
+            return ServiceReference<S>();
```
jeffdiclemente, line 345: this changes the behavior of [the public API](https://docs.cppmicroservices.org/en/stable/framework/doc/api/main/ServiceTracker.html#_CPPv2NK16cppmicroservices14ServiceTracker19GetServiceReferenceEv). I'm not opposed to this change in behavior, but there needs to be an update the docs along with this change.

tcormackMW: Ohhhh interesting.

I was thinking this was only an internal method. Hmm. Ok, we'll have to think about whether this should change or we should just have a helper method internall, one that throws and the other that doesn't... I'm thinking this may be better in case folks rely on the throw.

framework/include/cppmicroservices/ServiceTracker.h
```diff
@@ -478,6 +478,15 @@ namespace cppmicroservices
...
 
       private:
+        /**
+         * Internal method to retrieve a reference following same rules for retrieval
+         * as exported <code>GetServiceReference</code> method
+         *
+         * @return A <code>ServiceReference</code> for a tracked service.
+         * @note If no services are being tracked, a default <code>ServiceReference</code> is returned
+         */
+        virtual ServiceReference<S> GetServiceReference_internal() const;
```
jeffdiclemente: can this be marked `noexcept`?

framework/include/cppmicroservices/ServiceTracker.h
```diff
@@ -478,6 +478,15 @@ namespace cppmicroservices
                             std::shared_ptr<TrackedParamType> const& service) override;
 
       private:
+        /**
+         * Internal method to retrieve a reference following same rules for retrieval
+         * as exported <code>GetServiceReference</code> method
```
jeffdiclemente: it would be good to note briefly why this function exists (i.e. why do we need a "no throw" version).

jeffdiclemente (approved):

PR #1225, opened 2026-01-09
Migrate away from option parser to use the CLI11

atul20031:
This change migrates CppMicroServices CLI tools from optionparser.h to CLI11 library.

Files Changed

Modified (4 files):
- tools/change_namespace/main.cpp
- tools/jsonschemavalidator/main.cpp
- tools/rc/ResourceCompiler.cpp
- framework/test/gtest/ResourceCompilerTest.cpp
- Added: third_party/CLI/CLI11.hpp (CLI11 header)
- Removed: third_party/optionparser.h

What to Review

Option definitions - Verify all options match original (names, flags, required/optional status)
Custom validators - Check non_empty_non_flag validator in change_namespace/main.cpp
Built-in validators - Verify CLI::ExistingFile and CLI::Range(0,9) usage
Count-based checks - In ResourceCompiler.cpp, review count() > 0 logic for detecting provided options
Error handling - Verify try-catch blocks properly handle CLI::ParseError
Test comments - Review inline comments explaining error code changes (105, 114)
Multi-value options - Check ->take_all() for -r, -z, -m flags in ResourceCompiler


framework/test/gtest/ResourceCompilerTest.cpp
```diff
@@ -750,15 +755,18 @@ TEST_F(ResourceCompilerTest, testFailureModes)
...
 
     cmd.str(std::string());
     cmd << rcbinpath;
     cmd << " --manifest-add manifest.json";
     cmd << " --bundle-name foo ";
     cmd << " --compression-level -1";
     // Test Failure mode: invalid --compression-level argument (-1)
-    ASSERT_EQ(EXIT_FAILURE, runExecutable(cmd.str()));
+    // CLI11 migration note: Range validator rejects invalid values during parsing (error 105).
+    ASSERT_NE(0, runExecutable(cmd.str()));
```
jeffdiclemente: I would like to see the same return code contract maintained; if there is any error, `EXIT_FAILURE` is always returned. I would prefer we don't change that behavior for users and maintain backward compatibility for error codes returned from `usResourceCompiler`

tools/change_namespace/main.cpp
```diff
@@ -1,141 +1,71 @@
...
+    
+    try {
+        app.parse(argc, argv);
+    } catch(const CLI::CallForHelp&) {
+        std::cout << app.help() << std::endl;
+        return 0;
+    } catch(const CLI::Error &e) {
+        std::cerr << e.what() << std::endl;
+        std::cerr << app.help() << std::endl;
+        throw std::runtime_error(e.what());
```
jeffdiclemente: This doesn't seem correct to throw an exception. It doesn't look like this implementation threw before, why now? 
This is most likely the cause of clang-tidy's flag on line 5 (https://clang.llvm.org/extra/clang-tidy/checks/bugprone/exception-escape.html)

jeffdiclemente (changes requested):

tools/jsonschemavalidator/main.cpp
```diff
@@ -99,92 +99,58 @@ validate(std::string const& jsonfile, std::string const& schemafile) noexcept
...
+
+    // Custom usage message
+    app.footer("\nExample:\n  jsonschemavalidator -s myschema.json -j mydata.json\n");
+
+    int retVal = EXIT_SUCCESS;
+    
+    // If no arguments (argc==1) or only --help/-h, print usage and exit 0
+    bool only_help = false;
+    for(int i=1; i<argc; ++i) {
+        std::string arg(argv[i]);
```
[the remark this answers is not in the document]
atul20031: Fixed this issue

tools/rc/ResourceCompiler.cpp
```diff
@@ -614,306 +614,124 @@ ZipArchive::AddResourcesFromArchive(std::string const& archiveFileName)
...
-    VERBOSE,
-    BUNDLENAME,
-    COMPRESSIONLEVEL,
-    OUTFILE,
-    RESADD,
-    ZIPADD,
-    MANIFESTADD,
-    BUNDLEFILE
-};
+int main(int argc, char** argv) {
```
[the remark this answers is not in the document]
atul20031: Fixed this.

framework/test/gtest/ResourceCompilerTest.cpp
atul20031: Updated the ResourceCompiler.cpp to return EXIT_FAILURE for all cases where parsing is unsuccessful.

tools/change_namespace/main.cpp
atul20031: Updated the code to return EXIT_FAILURE as before.

jeffdiclemente (changes requested):
CLI11 needs to be added to the thirdparty [README file](https://github.com/CppMicroServices/CppMicroServices/blob/development/third_party/README), following the same format as the other entries in this file

atul20031:
> CLI11 needs to be added to the thirdparty [README file](https://github.com/CppMicroServices/CppMicroServices/blob/development/third_party/README), following the same format as the other entries in this file

Added the entry in README file

jeffdiclemente (approved):

framework/test/gtest/ResourceCompilerTest.cpp
```diff
@@ -1393,4 +1393,4 @@ TEST_F(ResourceCompilerTest, testManifestWithNullTerminator)
 
     // Test that the JSON content matches the expected JSON content.
     ASSERT_EQ(0, nullTerminatorJSON.compare(expectedJSON));
-}
+}
```
vivekkushwaha39: Can we remove this change. I don't see any related code changes?

atul20031: Removed the file from change.

PR #1226, opened 2026-01-14
New lambda for shutdown work

tcormackMW:


framework/include/cppmicroservices/Constants.h
```diff
@@ -317,6 +317,13 @@ namespace cppmicroservices
         US_Framework_EXPORT extern const std::string
             FRAMEWORK_BUNDLE_VALIDATION_FUNC; // = "org.cppmicroservices.framework.bundle.validation.function"
 
+        /**
+         * Framework extra shutdown functionality property specifying a function that is invoked by the framework after
+         * waitForStop has finished
```
jeffdiclemente: are there any restrictions on what the shutdown function can do? e.g. can it spawn threads? can it do I/O access? etc... We should be as clear as we can about the expectations for an implementation of a shutdown function.

framework/src/util/FrameworkPrivate.cpp
```diff
@@ -109,6 +109,24 @@ namespace cppmicroservices
...
+                try
+                {
+                    if (!func.Empty())
+                    {
+                        any_cast<std::function<void(void)>>(func)();
+                    }
+                }
+                catch (...)
+                {
+                    throw std::runtime_error { "The framework shutdown callback threw an exception" };
```
jeffdiclemente: can we append the original exception message, if it exists, to our re-thrown exception?

framework/test/gtest/FrameworkTest.cpp
```diff
@@ -849,4 +848,21 @@ TEST(FrameworkTest, LoadLibraryLogsMessagesTest)
 }
 #endif
 
+TEST(FrameworkTest, ConfigurationWithExtraShutdownWork)
```
jeffdiclemente, line 834: Its worthwhile to have another test case that tests the error case when the lambda throws an exception.

jeffdiclemente (approved):

tcormackMW:
changed to invoke once and only once, and to not allow throw. shutdown lambda shouldn't throw. it is best effort work and leads to issues if it throws

framework/src/util/FrameworkPrivate.cpp
```diff
@@ -109,6 +109,17 @@ namespace cppmicroservices
     FrameworkEvent
     FrameworkPrivate::WaitForStop(std::chrono::milliseconds const& timeout)
     {
+        auto shutdownFuncIter
+            = coreCtx->frameworkProperties.find(cppmicroservices::Constants::FRAMEWORK_EXTRA_SHUTDOWN_FUNC);
+        detail::ScopeGuard extraFunc(
```
Burgch, line 115: Does it make sense to invoke this if the wait times out?

framework/test/gtest/FrameworkTest.cpp
```diff
@@ -849,4 +848,25 @@ TEST(FrameworkTest, LoadLibraryLogsMessagesTest)
...
+{
+    int capt { 0 };
```
Burgch: As this is being modified and read from different threads, `atomic<int>`?

framework/include/cppmicroservices/Constants.h
```diff
@@ -317,6 +317,20 @@ namespace cppmicroservices
...
+         *
+         * @return void
+         * @note The value must be a callable convertible to: std::function<void()>
+         * @note MUST NOT throw exceptions.
+         * @note MUST be idempotent -- callable repeatedly.
+         * @note The callback may create threads or async work, but **must wait for all such
+         * work to finish before returning**.
+         * 
+         */
+        US_Framework_EXPORT extern std::string const FRAMEWORK_EXTRA_SHUTDOWN_FUNC;; // = "org.cppmicroservices.framework.shutdown.function"
```
carneyweb: extra semicolon.

framework/include/cppmicroservices/Any.h
```diff
@@ -132,23 +132,23 @@ namespace cppmicroservices
                                                        int32_t const);
     US_Framework_EXPORT std::ostream& any_value_to_cpp(std::ostream& os, bool val, uint8_t const, int32_t const);
 
-    template <typename T>
+    template <typename R, typename... Args>
```
carneyweb: Perhaps it'd be worth it to give a more descriptive name to the R template argument... something like:

template <typename ReturnT, typename... Args>

YMMV.

framework/test/gtest/FrameworkTest.cpp
tcormackMW: technically yes haha. I was lazy because I know that this thread won't touch it till AFTER the other thread has but yes, it should be atomic.

carneyweb (approved):
LGTM.

mwmcarney (approved):
LGTM.

framework/src/util/FrameworkPrivate.cpp
tcormackMW: Nope. Good call

framework/include/cppmicroservices/Constants.h
tcormackMW: Ugh. Thanks!

PR #1227, opened 2026-01-15
Fix issue of concurrent ref removal and addition in activate/deactivate of component

tcormackMW:
This fixes half of the problem, which is that repetative deactivates were being invoked because between when the `set_value` invocation was made and returning to the top of the while, the service became active again and was then deactivated again. Instead, we just break. That reduces failure in the test from roughly 10% to closer to 1%.

the second failure mode has to do with the async delivery of notifications. It's progress can be tracked in this issue: #1228

### DIFF
--- a/compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
+++ b/compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
@@ -534,4 +534,91 @@ namespace test
         framework.Stop();
         framework.WaitForStop(std::chrono::milliseconds::zero());
     }
+
+    class basicDS5 : public test::DSGraph05
+    {
+      public:
+        basicDS5() = default;
+        ~basicDS5() override = default;
+        std::string
+        Description() override
+        {
+            return "basicDS5";
+        };
+    };
+
+    TEST_F(TestAsyncWorkServiceEndToEnd, VerifyNoDuplicateDeactivate)
+    {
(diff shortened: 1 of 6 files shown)


Burgch (approved):
Looks good, thanks Toby!

compendium/DeclarativeServices/test/gtest/TestAsyncWorkService.cpp
```diff
@@ -534,4 +534,73 @@ namespace test
...
+        basicDS5() = default;
+        ~basicDS5() override = default;
+        std::string
+        Description() override
+        {
+            return "basicDS5";
+        };
+    };
+
+    TEST_F(TestAsyncWorkServiceEndToEnd, VerifyNoDuplicateDeactivate)
```
jeffdiclemente, line 550: I don't understand how this verifies that `Deactivate` is not called more times than expected.

Would it be better to mock the Component Configuration like is done in [TestCCActiveState.cpp](https://github.com/CppMicroServices/CppMicroServices/blob/development/compendium/DeclarativeServices/test/gtest/TestCCActiveState.cpp) and verify through mock expectations that a Deactivate call is not duplicated?

tcormackMW: It doesn't directly, but if the bug that exists were to occur, the assertion on state of service as satisfied would fail

jeffdiclemente (approved):

PR #1229, opened 2026-01-27
Add mutual exclusion between addition/removal of referenced services

tcormackMW:
Reduced the failure of the test `TestAsyncWorkServiceEndToEnd.VerifyNoDuplicateDeactivate` from .33% to no failures after 10k runs. This merely enforces that the removal of a reference happens independently of the addition of a reference to make sure the notifications arrive at the binding policy in the same order they were found in the reference manager. Otherwise you get satisfied services that are marked as `satisfied->satisfied->unsatisfied` instead of `satisfied->unsatisfied->satisfied` with rapid removal and addition of references

### DIFF
--- a/compendium/DeclarativeServices/src/manager/ReferenceManagerImpl.cpp
+++ b/compendium/DeclarativeServices/src/manager/ReferenceManagerImpl.cpp
@@ -183,6 +183,7 @@ namespace cppmicroservices
         cppmicroservices::InterfaceMapConstPtr
         ReferenceManagerBaseImpl::AddingService(cppmicroservices::ServiceReference<void> const& reference)
         {
+            std::unique_lock<std::mutex> l(stateChangeMutex);
             // Each service registered by DS contains a service property representing the component configuration name
             // to which it belongs. By checking the component configuration name of a service it can be determined
             // whether this service will satisfy its own reference. If it would, it is not a matched reference as
@@ -226,6 +227,7 @@ namespace cppmicroservices
         ReferenceManagerBaseImpl::RemovedService(cppmicroservices::ServiceReference<void> const& reference,
                                                  cppmicroservices::InterfaceMapConstPtr const& /*service*/)
         {
+            std::unique_lock<std::mutex> l(stateChangeMutex);
             { // acquire lock on matchedRefs
                 auto matchedRefsHandle = matchedRefs.lock();
                 matchedRefsHandle->erase(reference);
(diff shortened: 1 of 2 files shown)


compendium/DeclarativeServices/src/manager/ReferenceManagerImpl.cpp
Burgch, line 207: My only worry is that this will call into code outside of CppMicroServices whilst holding the lock. I don't know if there's anything we can do about that, but I guess it makes this non-reentrant, or else we'll deadlock?

Burgch (commented):
I agree these should be mutually exclusive - just wonder if we can somehow release the lock before we trigger external code?

compendium/DeclarativeServices/src/manager/ReferenceManagerImpl.cpp
tcormackMW: mmmmmm im sure you're right. we had one instance of reentrancy (see failing check in previous commit) which I have resolved by using a `std::recursive_mutex` which I think... is safe.

But I hadn't thought about reentrancy caused by user code. If a user, when their bind/unbind method is invoked, causes state changes on a **new thread** for the **same component** (which would be totally bizarre... the work of a component ends up causing state changes in that same component??) then that could lead to a deadlock if the component waited on that work to finish.

Burgch: Right - the only way I can think of to make it both robust to rapid events AND potential re-entrancy is to make all of the internal state changes under the lock, but then trigger the invocation of user-code after the lock is released. I think that might be possible, but I think it might require passing something with an e.g. `unlock()` method to `RegisterService` so it could be released before [this line](https://github.com/CppMicroServices/CppMicroServices/blob/development/framework/src/service/ServiceRegistry.cpp#L143) ?

compendium/DeclarativeServices/src/manager/ReferenceManagerImpl.cpp
```diff
@@ -183,6 +183,7 @@ namespace cppmicroservices
         cppmicroservices::InterfaceMapConstPtr
         ReferenceManagerBaseImpl::AddingService(cppmicroservices::ServiceReference<void> const& reference)
         {
+            std::lock_guard<std::recursive_mutex> l(stateChangeMutex);
```
Burgch: Does CppMicroServices require C++17 now? If so, you can use `std::lock_guard l{stateChangeMutex};` and rely on type-deduction guidelines to determine the class template type automatically.

Burgch (approved):
I think this is a good change to make, so it should go in. As to how we can be robust to user-code deadlocks, I think that's a larger challenge, which will require more thinking!

mwmcarney (approved):
LGTM.

compendium/DeclarativeServices/src/manager/ReferenceManagerImpl.cpp
tcormackMW: hm. i hadn't thought of that.

PR #1230, opened 2026-02-09
Fix race for bundle starts and framework stop

tcormackMW:
Latent race identified by Conor. 

New test failed ~50% before fix, 0 times in 1k after. 

I think this new test exposes the issue we have been seeing with bundle state changes concurrently with framework shutting down is causing problems. On this new branch or the current tip, the test fails primarily in serviceRegistry with segfaults... I think we are missing a race here external to the one this geck solves.

### DIFF
--- a/framework/src/bundle/CoreBundleContext.cpp
+++ b/framework/src/bundle/CoreBundleContext.cpp
@@ -252,12 +252,11 @@ namespace cppmicroservices
         return std::string();
     }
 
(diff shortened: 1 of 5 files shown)


Burgch (approved):
This looks great, thank you very much Toby!

Burgch (approved):
Thanks!

jeffdiclemente (approved):

PR #1233, opened 2026-03-02
Remove unnecessary 3rdparty libs, removed unnecessary patches

achristoforides:
This change does some prework required for our move to use Conan for dependencies.

`tinyscheme` is removed as we no longer use it and the patches for `dirent_win32` and `jsoncpp` were removed as they are not necessary and are an impedance to switching to using Conan.


third_party/README
```diff
@@ -12,28 +12,10 @@ CppMicroServices library, it is marked as such.
...
-   longer than MAX_PATH.
-
-jsoncpp
--------
-
-MIT License or Public Domain
-
-https://github.com/open-source-parsers/jsoncpp
-(formerly http://jsoncpp.sourceforge.net)
-1.9.5
```
jeffdiclemente, line 32: shouldn't the section about jsoncpp remain?

third_party/README
jeffdiclemente: I know this has nothing to do with your changes, but can you also remove this section. it should have been removed with PR #1225, but wasn't. Thanks.

jeffdiclemente (changes requested):

third_party/README
achristoforides: Ah, yes. Thanks for catching. Thought I was deleting the "patches" section but apparently we didn't have one for that.

third_party/README
achristoforides: Yep, removed.

jeffdiclemente (approved):

tcormackMW (approved):

PR #1234, opened 2026-03-03
`getcomponentDescriptionDTOs` safety

tcormackMW:
Make sure we handle the case where a bundle has stopped so that exceptions don't fall out of SCR service.


compendium/DeclarativeServices/test/gtest/TestInteg.cpp
```diff
@@ -0,0 +1,369 @@
...
+        // Wait for all threads to complete
+        for (auto& fut : futures)
+        {
+            fut.get();
+        }
+
+        ASSERT_FALSE(failed.load());
+
+        framework.Stop();
+        framework.WaitForStop(std::chrono::milliseconds::zero());
```
jeffdiclemente: isn't this unnecessary redundant with the `IntegrationTestFixture::Teardown()` ?

compendium/DeclarativeServices/test/gtest/TestInteg.cpp
```diff
@@ -0,0 +1,369 @@
...
+            fut.get();
+        }
+
+        sRef = ctx.GetServiceReference<test::DSGraph04>();
+        ASSERT_TRUE(sRef);
+        svc = ctx.GetService<test::DSGraph04>(sRef);
+        ASSERT_EQ(svc->Description(), "DSGraph04");
+
+        framework.Stop();
+        framework.WaitForStop(std::chrono::milliseconds::zero());
```
jeffdiclemente: isn't this unnecessary with the IntegrationTestFixture::Teardown() ?

compendium/DeclarativeServices/src/ServiceComponentRuntimeImpl.cpp
```diff
@@ -133,7 +133,20 @@ namespace cppmicroservices::scrimpl
...
+            }
+            catch (std::exception const& e)
+            {
+                logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_DEBUG,
+                           "Failed to create ComponentDescriptionDTO: " + std::string(e.what()));
+            }
+            catch (...)
+            {
+                logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_DEBUG,
+                           "Failed to create ComponentDescriptionDTO: Unknown exception");
```
jeffdiclemente: given the catch of `std::exception` above, I don't think this catches anything we should catch. If we are concerned about a bundle already being stopped by the time that `CreateDTO` is called, the only realistic thing we'll catch is a `std::exception` of some type.

compendium/DeclarativeServices/src/ServiceComponentRuntimeImpl.cpp
```diff
@@ -133,7 +133,20 @@ namespace cppmicroservices::scrimpl
...
         for (auto holder : compMgrs)
         {
-            componentDTOs.push_back(CreateDTO(holder));
+            try
+            {
+                componentDTOs.push_back(CreateDTO(holder));
+            }
+            catch (std::exception const& e)
+            {
+                logger->Log(cppmicroservices::logservice::SeverityLevel::LOG_DEBUG,
```
jeffdiclemente: I'm inclined to log this as a warning instead. If we can't create a DTO when the caller expects one to, then it may be a warning that something is wrong with a bundle (e.g., stopped when it shouldn't have been).
This is minor and I'm fine with the level remaining what it is.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #1238, opened 2026-03-24
Concurrent access to coreInfo bundle

tcormackMW:
the running theory here is as follows:

`coreInfo->bundle_` is read via `.lock()` in many places under only the ServiceRegistry lock, while `Unregister()` calls `bundle_.reset()` under `LockServiceRegistration()`. These are different locks, so concurrent `weak_ptr::lock()` and `weak_ptr::reset()` on the same object could result in something like a segfault like this:

cppmicroservices::ServiceRegistry::GetRegisteredByBundle at [ServiceRegistry.cpp](https://github.com/CppMicroServices/CppMicroServices/blob/f71ad06d42ef5d44f7656d702bbf5c5a6332e274/framework/src/service/ServiceRegistry.cpp#L342-L343)
cppmicroservices::Bundle::GetRegisteredServices at [Bundle.cpp](https://github.com/CppMicroServices/CppMicroServices/blob/f71ad06d42ef5d44f7656d702bbf5c5a6332e274/framework/src/bundle/Bundle.cpp#L284-L285)

### DIFF
--- a/framework/src/service/ServiceReferenceBasePrivate.cpp
+++ b/framework/src/service/ServiceReferenceBasePrivate.cpp
@@ -93,8 +93,8 @@ namespace cppmicroservices
                         MakeBundle(bundle->shared_from_this()),
                         message,
                         std::make_exception_ptr(ServiceException(message, ServiceException::Type::FACTORY_ERROR))));
-                    return smap;
                 }
+                return smap;
             }
             {
                 std::shared_ptr<BundlePrivate> bundleSnapshot;
--- a/framework/src/service/ServiceReferenceBasePrivate.cpp
+++ b/framework/src/service/ServiceReferenceBasePrivate.cpp
@@ -64,6 +64,7 @@ namespace cppmicroservices
     ServiceReferenceBasePrivate::SafelyGetBundle() const
     {
         auto regLock = LockServiceRegistration();
+        US_UNUSED(regLock);
         auto bundle = coreInfo->bundle_.lock();
         return bundle;
     }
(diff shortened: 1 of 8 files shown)


framework/src/service/ServiceRegistry.cpp
```diff
@@ -339,6 +340,8 @@ namespace cppmicroservices
 
         for (auto& sr : serviceRegistrations)
         {
+            auto regLock = sr.LockServiceRegistration();
+            US_UNUSED(regLock);
```
jeffdiclemente: performance: should this locking be done outside of the loop instead?

framework/src/service/ServiceReferenceBasePrivate.cpp
```diff
@@ -71,29 +71,41 @@ namespace cppmicroservices
                                                             ServiceRegistrationBase(registration.lock()));
             if (!smap || smap->empty())
             {
-                if (auto bundle_ = coreInfo->bundle_.lock())
+                std::shared_ptr<BundlePrivate> bundle_;
+                {
+                    auto regLock = LockServiceRegistration();
+                    US_UNUSED(regLock);
+                    bundle_ = coreInfo->bundle_.lock();
```
jeffdiclemente: This pattern is used many times that I think its worth making a function instead of duplicating the code pattern.

jeffdiclemente (changes requested):

framework/src/service/ServiceRegistry.cpp
tcormackMW: No, this is done per `ServiceRegistration` so it has to happen within the loop... This is a good thing to note though, it is worth looking at performance results for this commit. Honestly, the need for locking here furthers what we spoke about which is that GetRegisteredByBundle is SO inefficient, moreso now. May be worth switching this to also hold a map to make this function call reasonable. It would eliminate `n-1` locks of 2 mutexes where `n` is the number of services in the full system... zoinks

tcormackMW:
local run of the bench tests had some improvements... not huge but potentially 5-20% depending on load. We'll see what github actions says when we merge.

jeffdiclemente (approved):

PR #1239, opened 2026-03-24
Upgrade miniz to 3.1.1 and remove custom patches

achristoforides:
# Description

This PR upgrades miniz to the latest GitHub release (v3.1.1) and also removes the custom patches we had to reduce friction when migrating to use Conan.

## Notes
Interestingly, we had a bug in `BundleElfFile.h`. When using the "link" option for embedding resources, our code wasn't handling the fact that for ELF binaries, when the linker does this, it must be paged align and thus padding is introduced.

The pointer that we get when reading the data sections of the binary thus points to the beginning of the page-aligned section where the linker embedded the zip binary contents.

This pointer is passed to miniz when we call `mz_*_init_mem`. Miniz finds the start of the zip header correctly however the offset field that defines the offset between the zip header and the EOCD header is relative to the pointer base. Since there is some `N` bytes of padding, the offset reported in the header information is actually `N` bytes off from where the EOCD header actually is.

This is handled by adjusting the base pointer to the data returned by `MappedFile::GetData()` to be `N` bytes advanced (by the padding amount) and by adjusting the data size returned by `MappedFile::GetSize()` to be `total section length - padding bytes`.


third_party/README
```diff
@@ -33,10 +33,10 @@ MIT License (GitHub) or Public Domain
 
 https://github.com/richgel999/miniz
 (formerly https://code.google.com/p/miniz/)
-v3.0.2
+v3.1.1
```
jeffdiclemente, line 36: the miniz.h file has 3.1.0, there seems to be a mismatch here. what's the correct version?

util/include/cppmicroservices/util/MappedFile.h
```diff
@@ -71,18 +72,21 @@ namespace cppmicroservices
...
+                return nullptr;
+            }
+            return static_cast<char*>(mappedAddress) + dataOffset;
         }
         std::size_t
         GetSize() const override
-        {
-            return mapSize;
-        }
+        { return mapSize - dataOffset; }
```
jeffdiclemente: minor, formatting: please fix the brace formatting to be consistent with the rest of the file

util/include/cppmicroservices/util/BundleMachOFile.h
```diff
@@ -93,7 +93,7 @@ namespace cppmicroservices
             fs.seekg(fileOffset);
             Mhdr mhdr;
             fs.read(reinterpret_cast<char*>(&mhdr), sizeof mhdr);
-            if (mhdr.filetype != MH_DYLIB && mhdr.filetype != MH_BUNDLE)
+            if (mhdr.filetype != MH_DYLIB && mhdr.filetype != MH_BUNDLE && mhdr.filetype != MH_EXECUTE)
             {
                 throw InvalidMachOException("Not a Mach-O dynamic shared library or bundle file.");
```
jeffdiclemente: message should be updated to also mention executable

util/include/cppmicroservices/util/MappedFile.h
jeffdiclemente: If `mmap` fails, `mapSize` is set to 0 but `dataOffset` is not, causing an underflow. There needs to be a check for `mapSize`.

tools/rc/CMakeLists.txt
```diff
@@ -27,6 +27,13 @@ if(NOT APPLE AND US_COMPILER_GNU_ARM AND CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64
 endif()
 
 add_executable(${US_RCC_EXECUTABLE_TARGET} ${srcs})
+
+# Suppress this compilation warning on unix OSes. Required with removing our patches
+# to miniz.
+if(UNIX)
+  target_compile_options(${US_RCC_EXECUTABLE_TARGET} PRIVATE -Wno-old-style-cast)
```
jeffdiclemente: improve as you go: use idiomatic `target_compile_definitions` instead

framework/test/gtest/CMakeLists.txt
```diff
@@ -134,6 +134,8 @@ if (US_COMPILER_MSVC AND BUILD_SHARED_LIBS)
   target_compile_options(${us_gtest_test_exe_name} PRIVATE -DGTEST_LINKED_AS_SHARED_LIBRARY)
 endif()
 
+target_compile_options(${us_gtest_test_exe_name} PRIVATE -DMINIZ_NO_ARCHIVE_WRITING_API -DMINIZ_NO_ZLIB_COMPATIBLE_NAMES)
```
jeffdiclemente: improve as you go: use idiomatic `target_compile_definitions` instead

CMakeLists.txt
```diff
@@ -584,6 +584,12 @@ else()
     usFunctionCheckCompilerFlags(-Wno-mismatched-tags US_CXX_FLAGS)
   endif()
 
+  # Suppress this compilation warning on unix OSes. Required with removing our patches
+  # to miniz.
+  if(UNIX OR MINGW)
+    usFunctionCheckCompilerFlags(-Wno-old-style-cast US_CXX_FLAGS)
```
jeffdiclemente: we should not be applying `-Wno-old-style-cast` globally, instead, it should be applied in a targeted way to the source files we know we need it for.

jeffdiclemente (changes requested):
good find on the page alignment bug!

third_party/README
achristoforides: The release on the GitHub page (https://github.com/richgel999/miniz/releases/tag/3.1.1) is the correct version.

I manually downloaded the source and looked and it does in-fact say 3.1.0.

This is not a mistake on our part.

util/include/cppmicroservices/util/MappedFile.h
```diff
@@ -52,6 +53,7 @@ namespace cppmicroservices
                 {
                     mappedAddress = nullptr;
                     mapSize = 0;
+                    dataOffset = 0;
```
jeffdiclemente, line 56: if `open` fails, `dataOffset` also needs to be set appropriately as not to cause a bug when calling `GetData()`

jeffdiclemente (changes requested):
looks like the macos builds are running into cast warnings

util/include/cppmicroservices/util/MappedFile.h
jeffdiclemente: I should revise my previous comment because its vague; What I mean to say is, if open() returns -1, `GetData()` should return a consistent answer that indicates that the open couldn't be opened for read, e.g. returning 0.

framework/test/gtest/CMakeLists.txt
```diff
@@ -134,6 +134,8 @@ if (US_COMPILER_MSVC AND BUILD_SHARED_LIBS)
   target_compile_options(${us_gtest_test_exe_name} PRIVATE -DGTEST_LINKED_AS_SHARED_LIBRARY)
 endif()
 
+target_compile_definitions(${us_gtest_test_exe_name} PRIVATE -DMINIZ_NO_ARCHIVE_WRITING_API -DMINIZ_NO_ZLIB_COMPATIBLE_NAMES)
```
jeffdiclemente: `MINIZ_NO_ARCHIVE_WRITING_API` is not a macro used by miniz. I think you meant `MINIZ_NO_ARCHIVE_WRITING_APIS`?

tools/rc/CMakeLists.txt
```diff
@@ -35,8 +36,7 @@ endif()
 target_link_libraries(${US_RCC_EXECUTABLE_TARGET} nowide::nowide)
 target_include_directories(${US_RCC_EXECUTABLE_TARGET} PRIVATE ${CppMicroServices_SOURCE_DIR}/third_party/boost/nowide/include)
 
-set_property(TARGET ${US_RCC_EXECUTABLE_TARGET} APPEND PROPERTY
-             COMPILE_DEFINITIONS "MINIZ_NO_ARCHIVE_READING_API;MINIZ_NO_ZLIB_COMPATIBLE_NAMES")
+target_compile_definitions(${US_RCC_EXECUTABLE_TARGET} PRIVATE -DMINIZ_NO_ARCHIVE_READING_API -DMINIZ_NO_ZLIB_COMPATIBLE_NAMES)
```
jeffdiclemente: It doesn't look like `MINIZ_NO_ARCHIVE_READING_API` is a macro used by miniz, it doesn't exist in the miniz source code.

jeffdiclemente (changes requested):
do we need to add any tests in framework\test\gtest\BundleObjFileTest.cpp for this change?

framework/test/gtest/CMakeLists.txt
achristoforides: We actually do need the APIs. With the define spelled wrong, obviously it wasn't removing them but once I fixed it locally, the files started having compilation errors.

We now don't have the warnings that were coming from having the APIs since we handle them via pragma statements.

tools/rc/CMakeLists.txt
achristoforides: We actually do need the APIs. With the define spelled wrong, obviously it wasn't removing them but once I fixed it locally, the files started having compilation errors.

We now don't have the warnings that were coming from having the APIs since we handle them via pragma statements.

jeffdiclemente:
It looks like the clang tidy check action failed to update this PR with issues it found. Take a look at the logs for this action to see the issues it found. Some of them look like they should be fixed.

Also, CodeQL action found an issue that needs to be triaged: https://github.com/CppMicroServices/CppMicroServices/pull/1239/checks?check_run_id=69229326139.

achristoforides:
> It looks like the clang tidy check action failed to update this PR with issues it found. Take a look at the logs for this action to see the issues it found. Some of them look like they should be fixed.
> 
> Also, CodeQL action found an issue that needs to be triaged: https://github.com/CppMicroServices/CppMicroServices/pull/1239/checks?check_run_id=69229326139.

The clang-tidy warnings are nothing to be concerned with for this change. Three relate to constants used in the new tests I added and suggest converting them to variables, which isn’t necessary. The remaining warning claims it can’t find TestUtilBundleListener.h, which doesn’t make sense given that the code compiles successfully on all platforms.

The miniz CodeQL issue is also not a concern, as we don’t use the function where the warning was reported, neither in the framework code (framework/ds/ca/etc.) nor in our tests. The affected code isn’t referenced by miniz.c or miniz.h either.

jeffdiclemente (approved):

PR #1240, opened 2026-03-24
fix for failed `serviceReferenceFromService`

tcormackMW:
`ServiceReferenceFromService` fails for services provided by a ServiceFactory

`ServiceReferenceFromService` relies on `std::get_deleter<CustomServiceDeleter>` to recover the `ServiceReference` from a service `shared_ptr`. This works when a service is obtained via `undleContext::GetService<S>()`, because that path wraps the service pointer in a `ServiceHolder` with a `CustomServiceDeleter` before handing it to the consumer.

However, when a service comes from a `ServiceFactory`, the path goes through `ServiceObjectsBase::GetServiceInterfaceMap()`, which returns an `InterfaceMapConstPtr`. The `InterfaceMapConstPtr` itself was correctly aliased through a `ServiceHolder` with `CustomServiceDeleter`, but the individual entries inside the map were plain `shared_ptr`s with their own independent control blocks. Any code that extracted a service pointer from the map — `ExtractInterface`, Declarative Services bind methods, `LocateService`, etc. — got a pointer without the `CustomServiceDeleter`, causing `ServiceReferenceFromService` to throw.

In `ServiceObjectsBase::GetServiceInterfaceMap()` in `ServiceObjects.cpp`, instead of copying the factory's `InterfaceMap` entries as-is, the method now builds a consumer-facing map where each entry aliases through the `ServiceHolder` (via the shared_ptr aliasing constructor). This ensures that any pointer extracted from the map carries the `CustomServiceDeleter` and works with `ServiceReferenceFromService`.

The `ServiceHolder` stores the original `InterfaceMap` from the factory for lifecycle management `(UngetService/UngetPrototypeService)`, while consumers see wrapped entries. This is consistent with how `BundleContext::GetService` and `ServiceObjectsBase::GetService` already wrap service pointers through a `ServiceHolder`.


jeffdiclemente (changes requested):
there are [some compiler errors on macOS](https://github.com/CppMicroServices/CppMicroServices/actions/runs/23511603835/job/68433720282?pr=1240) that need to be fixed,

compendium/DeclarativeServices/test/gtest/TestInteg.cpp
```diff
@@ -0,0 +1,95 @@
...
+        auto dependentSvcRef = context.GetServiceReference<test::FactoryServiceDependent>();
+        auto dependentSvc = context.GetService<test::FactoryServiceDependent>(dependentSvcRef);
+        ASSERT_TRUE(dependentSvc && dependentSvc->didBind());
+
+        // verify that two calls to getServiceFromReference will return the same service, with the right interfaceid
+        ASSERT_EQ(cppmicroservices::ServiceReferenceFromService(
+                      context.GetService<test::FactoryCreatedService>(
+                          context.GetServiceReference<test::FactoryCreatedService>()))
+                      .GetInterfaceId(),
+                  "test::FactoryCreatedService");
```
Burgch: Might it make sense to assert no throw around this?

compendium/DeclarativeServices/test/gtest/TestInteg.cpp
```diff
@@ -0,0 +1,95 @@
...
+        ASSERT_EQ(cppmicroservices::ServiceReferenceFromService(context.GetService<test::FactoryCreatedService>(
+                      context.GetServiceReference<test::FactoryCreatedService>())),
+                  cppmicroservices::ServiceReferenceFromService(context.GetService<test::FactoryCreatedService>(
+                      context.GetServiceReference<test::FactoryCreatedService>())));
```
Burgch: I'm not sure I understand the point of this assertion?

Burgch (approved):
Nice find - sorry for beliving this worked correctly! I'm not quite sure why the DS test wasn't failing - DS also uses a `ServiceFactory` so I thought it was already exercising the relevant code!

Thanks Toby!

compendium/DeclarativeServices/test/gtest/TestInteg.cpp
tcormackMW: I think if it threw, gtest would fail the test, no? If not I can update it but that's what I assumed. I'll double check.

compendium/DeclarativeServices/test/gtest/TestInteg.cpp
tcormackMW: Yeah... fair. This one was just that calling serviceReferenceFromService gave us the same underlying object on this factory created service. I was just making sure that all our wrapping and unwrapping got us back to the same serviceReference objects... I can remove it if you think it's not worth doing. I was just writing something to give some peace of mind

compendium/DeclarativeServices/test/gtest/TestInteg.cpp
jeffdiclemente: Yes, if `ServiceReferenceFromService` throws, gtest will report that this testpoint failed. wrapping `ASSERT_EQ` in a `ASSERT_NO_THROW` is useful here.

On a related note, I noticed that `ServiceReferenceFromService` exported API documentation (in framework\include\cppmicroservices\ServiceReference.h) doesn't note that it throws. That should be changed since it can throw a `std::runtime_error` exception

compendium/test_bundles/ServiceFactoryDependentBundle/src/ServiceImpl.hpp
```diff
@@ -0,0 +1,40 @@
...
+  public:
+    dependentImpl() = default;
+    ~dependentImpl() override = default;
+
+    void
+    BindcreatedSvc(std::shared_ptr<test::FactoryCreatedService> factoryCreated)
+    {
+        try
+        {
+            auto srefFromService1 = cppmicroservices::ServiceReferenceFromService(factoryCreated);
```
jeffdiclemente, line 19: minor: unused variable

compendium/test_bundles/ServiceFactoryDependentBundle/src/ServiceImpl.hpp
```diff
@@ -0,0 +1,40 @@
...
+            _bound = true;
+        }
+        catch (std::exception const& e)
+        {
+            std::cout << "Exception getting ref: " << e.what() << std::endl;
```
jeffdiclemente, line 25: would it be more useful to call `FAIL()` here with this string streamed through the `FAIL` macro? 

It looks like `didBind` serves the same goal, but calling using FAIL() seems like it eliminates the need to have `didBind`, if I'm understanding the test correctly.

compendium/DeclarativeServices/test/gtest/TestInteg.cpp
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        void
+        TearDown() override
+        {
+            framework.Stop();
+            framework.WaitForStop(std::chrono::milliseconds::zero());
+        }
+
+        std::shared_ptr<cppmicroservices::service::component::runtime::ServiceComponentRuntime> dsRuntimeService;
```
jeffdiclemente, line 67: `dsRuntimeService` doesn't appear to be referenced anywhere, consider removing it.

jeffdiclemente (changes requested):

compendium/test_bundles/ServiceFactoryDependentBundle/src/ServiceImpl.hpp
tcormackMW: I can fail, but I like having the didbind personally. I like verifying that I can actually invoke something on the returned service... This may be overkill and not actually useful, but I think verifying the object actually works (at least to some extent) rather than it is just an object returned seems useful. What do you think?

tcormackMW:
> Nice find - sorry for beliving this worked correctly! I'm not quite sure why the DS test wasn't failing - DS also uses a `ServiceFactory` so I thought it was already exercising the relevant code!
> 
> Thanks Toby!

No worries, I was the one who implemented this fix... should've caught it way back when. Thanks for the help!

compendium/test_bundles/ServiceFactoryDependentBundle/src/ServiceImpl.hpp
jeffdiclemente: that's fair. there is already a testpoint that tests for `ServiceReferenceFromService` throwing, so that seems reasonable to not `FAIL()` here.

jeffdiclemente (approved):

PR #1241, opened 2026-03-26
Fix race in miniz data access

tcormackMW:
Attempt to fix the miniz race reported by tsan

https://github.com/CppMicroServices/CppMicroServices/actions/runs/23594271478/job/68706877510#:~:text=11%3A%20WARNING%3A,2718

### DIFF
--- a/framework/src/bundle/BundleResourceContainer.cpp
+++ b/framework/src/bundle/BundleResourceContainer.cpp
@@ -100,11 +100,12 @@ namespace cppmicroservices
     BundleResourceContainer::GetStat(BundleResourceContainer::Stat& stat)
     {
         OpenAndInitializeContainer();
+        std::unique_lock<std::mutex> l(m_ZipFileStreamMutex);
         int fileIndex
             = mz_zip_reader_locate_file(const_cast<mz_zip_archive*>(&m_ZipArchive), stat.filePath.c_str(), nullptr, 0);
         if (fileIndex >= 0)
         {
-            return GetStat(fileIndex, stat);
+            return GetStatLocked(fileIndex, stat);
         }
         return false;
     }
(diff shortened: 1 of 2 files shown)


tcormackMW:
This pr should fix the tsan report that has been happening, but there is still one being created which I believe to be a false positive. 

The fix in this PR adds `m_ZipFileStreamMutex` locking to both `GetStat` overloads, matching the protection already present in `GetData`. This fixes the real race where two threads concurrently called `GetStat` (unlocked) and `GetData` on the same `BundleResourceContainer`, both entering `mz_zip_file_stat_internal` on the same `mz_zip_archive`.

A second TSAN report was observed post-fix, also in `mz_zip_file_stat_internal`. However, this one seems to be a false positive.

**First report (real race):** One thread holds a mutex, the other holds none. This means `GetStat` was unprotected while `GetData` was protected while accessing the same `mz_zip_archive`.
**Second report (false positive)**: Both threads hold a mutex, but at different addresses (M0 ≠ M1). This means they are operating on different `BundleResourceContainer` instances, each with its own independent `mz_zip_archive` member and its own `m_ZipFileStreamMutex`. There is no shared mutable state between them.

The test (`BundleStartAfterFrameworkStopKitchenSink`) launches `std::async` tasks that each install a different bundle (`TestBundleRL, TestBundleS, TestBundleSL1,` etc.), each creating its own `BundleResourceContainer`. The TSAN report is maybe triggered by heap address recycling — malloc on one thread reuses an address recently free'd on another thread...

To verify, I added instrumentation logging `this`, `&m_ZipFileStreamMutex`, and `&m_ZipArchive` in `GetStat` and `GetData`. The output confirms that every `GetStat/GetData` pair operates on the same container (matching pointers), and concurrently running threads operate on different containers (distinct pointers):
```
GetStat(path) this=0x7f6a08000df0 mutex=0x7f6a08000f00 archive=0x7f6a08000e20
GetData this=0x7f6a08000df0 mutex=0x7f6a08000f00 archive=0x7f6a08000e20
GetStat(path) this=0x7f6a04000df0 mutex=0x7f6a04000f00 archive=0x7f6a04000e20
GetData this=0x7f6a04000df0 mutex=0x7f6a04000f00 archive=0x7f6a04000e20
GetStat(path) this=0x7f6a0c000df0 mutex=0x7f6a0c000f00 archive=0x7f6a0c000e20
GetData this=0x7f6a0c000df0 mutex=0x7f6a0c000f00 archive=0x7f6a0c000e20
```

No two concurrent threads share a `this/mutex/archive` pointer, confirming independent containers with no shared mutable state.

Locally the tsan does not reproduce, we will see what is printed and the tsan report with my latest push.

PR #1243, opened 2026-03-28
fix Bundle.Uninstall, Did not call FreeLibrary

yansong1221:
fix Bundle.Uninstall, Did not call FreeLibrary

### DIFF
--- a/framework/include/cppmicroservices/Bundle.h
+++ b/framework/include/cppmicroservices/Bundle.h
@@ -831,7 +831,7 @@ namespace cppmicroservices
          *
          * @see #Stop()
          */
-        void Uninstall();
+        void Uninstall(bool freeSharedLibrary = false);
 
       protected:
         Bundle(std::shared_ptr<BundlePrivate> const& d);
--- a/framework/include/cppmicroservices/ServiceReference.h
+++ b/framework/include/cppmicroservices/ServiceReference.h
@@ -26,6 +26,14 @@
 #include "cppmicroservices/ServiceInterface.h"
 #include "cppmicroservices/ServiceReferenceBase.h"
 
+#ifndef __cpp_vsersion
+#    if defined(_MSC_VER)
+#        define __cpp_vsersion _MSVC_LANG
+#    else
+#        define __cpp_vsersion __cplusplus
+#    endif
+#endif // !__cpp_vsersion
+
 namespace cppmicroservices
 {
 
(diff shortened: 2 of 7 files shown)


tcormackMW:
@jeffdiclemente @yansong1221 

Does unloading the dll here create problems for consumers still holding onto objects created by the underlying libraries?

We allow folks to hold on to the `shared_ptr` to the objects created by bundles indefinitely, and right now those services stay alive. If we unload the dll that owns them, do we create a potential segfault? In my reading we do, but maybe I am wrong here?

If I am right, this is not a change that can happen because it would require large scale changes in any codebase using microservices to essentially have guards so that people cannot hold onto their services after the bundle has gone away (which I am not even sure how we would enforce)

Though I don't really see internal calls to `bundle::uninstall`. This is maybe something that customers invoke, which I guess gives us a bit more control -- i.e. regular `bundle.stop` invocations don't cause this fallout I mentioned.

jeffdiclemente:
The absence of calling `FreeLibrary` is deliberate. 

CppMicroServices does not support unloading a C++ shared library that was installed as a cppmicroservbices bundle. The main reason is that on posix systems, it is effectively impossible to actually unload a C++ shared library from a process and while its possible on Windows, cppmicroservices wants to remain consistent on all OS platforms.

tcormackMW:
@yansong1221 

Thank you very much for this request! Always great to get more people looking into the codebase. However, in this case, I think the choice to not invoke `FreeLibrary` was intentional.

The OSGI spec does say that this should be invoked, however in Java the requirements are different. Java has a garbage collector that can track object lifetimes and ensure that no references to classes from an unloaded bundle remain before reclaiming the classloader. In C++, we have no such luxury. Once a `shared_ptr<S>` to a service object has been handed to a consumer, we have no way to guarantee that the consumer has released it before calling `dlclose/FreeLibrary`. If the library is unloaded while a consumer still holds a service object, any call through that object's vtable will jump into unmapped memory and crash the process.

The current design is intentional:

- `~SharedLibrary()` deliberately does not unload the library — the destructor comment explicitly states this. 
- `Stop` unregisters services and invalidates the `BundleContext`, preventing new service retrieval, but leaves the `.so/.dll` mapped so existing objects remain valid.
- `Uninstall` removes the bundle from the registry but still does not unload the binary.

Please let me know if there was anything I missed or if you have thoughts about this. 

We'd also be happy to review other contributions — thanks again for taking the time!

yansong1221:
Can it be made optional? In this way, there is no need to restart the whole software when updating the module. The risk brought by FreeLibrary is borne by the developer himself.

tcormackMW:
Hi,

I see your point here. If you were to replace a binary (bundle) with a new one with the same name and no versioning, the system wouldn't be able to update without fully unloading.

I think optional functionality here that is configurable (should likely be defaulted OFF as that is the current behavior) could be good. If this sounds good, please provide a PR that shows this behavior and we will be happy to review it for you!

Thanks again!

PR #1250, opened 2026-04-23
Support cross compilation -- embed the LD_LIBRARY_PATH into the cmake options

tcormackMW:


cmake/usFunctionEmbedResources.cmake
```diff
@@ -149,8 +159,8 @@ function(usFunctionEmbedResources)
     endif()
     add_custom_command(
       OUTPUT ${_zip_archive}
-      COMMAND ${resource_compiler} -o ${_zip_archive} ${_zip_args}
-      DEPENDS ${_res_zips} ${resource_compiler}
+      COMMAND ${_rc_env_cmd} ${resource_compiler} -o ${_zip_archive} ${_zip_args}
```
jeffdiclemente, line 162: warn - uninitialized variable: `_rc_env_cmd` is only defined if cross compiling. For non cross compiling logic paths, CMake should treat an uninitialized variable as an empty string, but its better to initialize it to empty explicitly.

cmake/usFunctionAddResources.cmake
```diff
@@ -167,9 +177,9 @@ function(usFunctionAddResources)
   add_custom_command(
     OUTPUT ${_res_zip}
     COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_CURRENT_BINARY_DIR}/${US_RESOURCE_TARGET}"
-    COMMAND ${resource_compiler} ${cmd_line_args} -o ${_res_zip} ${_bundle_args} ${_file_args} ${_us_zip_args}
+    COMMAND ${_rc_env_cmd} ${resource_compiler} ${cmd_line_args} -o ${_res_zip} ${_bundle_args} ${_file_args} ${_us_zip_args}
```
jeffdiclemente, line 180: warn - uninitialized variable: `_rc_env_cmd` is only defined if cross compiling. For non cross compiling logic paths, CMake should treat an uninitialized variable as an empty string, but its better to initialize it to empty explicitly.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #1252, opened 2026-04-28
`symbolicName` and `bundleId` should be persistent

### LINKED ISSUE #1218: `SCRExtensionRegistry` and `ComponentRegistry` index stored objects using `bundleIds` which can disappear as the bundle dies. Causes orphaned data

tcormackMW:
These mechanisms are used to index data. They should persist after bundles are stopped.


compendium/DeclarativeServices/test/gtest/TestComponentManagerImpl.cpp
```diff
@@ -111,24 +111,16 @@ namespace cppmicroservices
...
-            {
-                EXPECT_NO_THROW({
-                    US_UNUSED(std::make_shared<ComponentManagerImpl>(mockMetadata,
-                                                                     mockRegistry,
-                                                                     bc,
-                                                                     fakeLogger,
-                                                                     asyncWorkService,
-                                                                     notifier));
-                });
-            }
```
Burgch, line 123: Why remove this?

framework/src/bundle/Bundle.cpp
```diff
@@ -74,23 +91,40 @@ namespace cppmicroservices
...
+            // if only one has sharedState, return true if us, false if them
+            return sharedState ? true : false;
+        }
+        // if neither have sharedState, return false
+        return false;
     }
 
-    Bundle::operator bool() const { return d != nullptr; }
+    Bundle::
+    operator bool() const
```
Burgch: This is some funky formatting?

framework/src/service/ServiceReferenceBasePrivate.h
```diff
@@ -184,9 +184,10 @@ namespace cppmicroservices
             {
                 // Make sure that we don't crash if the shared_ptr service object outlives
                 // the BundlePrivate or CoreBundleContext objects.
-                if (!b.expired())
+                auto bPrivateShared = b.lock();
+                if (bPrivateShared)
                 {
-                    DIAG_LOG(*b.lock()->coreCtx->sink)
```
Burgch, line 189: Nice catch!

.clang-format
```diff
@@ -14,8 +14,8 @@ AlignConsecutiveMacros:
 AlignEscapedNewlines: Left
 AllowAllArgumentsOnNextLine: false
 AllowAllParametersOfDeclarationOnNextLine: false
-AllowShortBlocksOnASingleLine: Always
-AllowShortFunctionsOnASingleLine: All
+AllowShortBlocksOnASingleLine: Never
+AllowShortFunctionsOnASingleLine: None
```
Burgch, line 18: I suspect one of these is the cause of the funky formatting of that `operator bool` ?

Burgch (approved):
Thanks!

framework/src/bundle/Bundle.cpp
tcormackMW: yep.

.clang-format
tcormackMW: was actually because of `AlwaysBreakAfterReturnType: AllDefinitions`

It happened in this push because I actually ran the clang and caused this to fire.

compendium/DeclarativeServices/test/gtest/TestComponentManagerImpl.cpp
tcormackMW: I removed it because it was redundant with the test code I added which instantiates an object like this. That being said, probably a good call to leave it and have it fail before a test that isn't meant to verify the base functionality and is doing more work than this.

Burgch (approved):
Thanks!

PR #1253, opened 2026-04-29
Remove jsoncpp usage

### LINKED ISSUE #1108: Two third party libraries providing same functionality
There are two third-party libraries, jsoncpp and rapidjson, that offer similar functionality. To minimize the attack surface of external dependencies and improve maintainability, only one of them should be used. The other can be removed from the codebase.

aadityap-mathworks:
Summary:

  - Replaced all jsoncpp usage with rapidjson across ResourceCompiler, SCRCodeGen, and their tests
  - Deleted third_party/jsoncpp.cpp and third_party/json/ , consolidating on rapidjson which was already used by the
  framework core (BundleManifest.cpp)
  - Migrated JSON parsing from jsoncpp's CharReaderBuilder/StreamWriterBuilder to rapidjson's Document::ParseStream and PrettyWriter
  - Implemented checkDuplicateKeys to replace jsoncpp's rejectDupKeys setting, since rapidjson silently accepts duplicate keys
  - JSON comments are no longer allowed in manifest files — jsoncpp permitted // and /* */ comments, but rapidjson's strict parser rejects them. Added a test to verify that manifests with comments are rejected.
  - Migrated JsonValueValidator (SCRCodeGen) from jsoncpp Json::Value to rapidjson::Value, including all manifest field validation in ManifestParserImpl.cpp
  - Fixed MinGW build — #undef GetObject after <Shlwapi.h> to prevent Windows macro from mangling rapidjson::Value::GetObject()
  - Fixed dangling references — replaced temporaries bound to const& with named JsonValueValidator variables in Main.cpp, TestCodeGenerator.cpp, and ManifestParserImpl.cpp
  - Updated third_party/README, CMakeLists files, and test code accordingly
  
  Feedback Addressed:
  - checkDuplicateKeys and toStyledString moved to util


compendium/tools/SCRCodeGen/Util.hpp
```diff
@@ -120,63 +124,70 @@ namespace codegen
...
 
-            // Return the Json value data[name]
-            // (data and name are the Json data and name specified in the constructors)
-            Json::Value
+            // Return a const reference to the rapidjson::Value for data[name].
+            rapidjson::Value const&
             operator()() const
             {
-                return jsonVal;
+                return *jsonVal;
```
jeffdiclemente, line 136: if the "choices" constructor was used, i.e. ```template <std::size_t S>
            JsonValueValidator(rapidjson::Value const& data, std::string name, ValidChoices<S> const& choices)``` was called, jsonVal should be `nullptr`. There should be some error handling here for that case.

compendium/tools/SCRCodeGen/ManifestParserImpl.cpp
```diff
@@ -131,45 +136,47 @@ ManifestParserImplV1::ParseAndGetComponentInfos(Json::Value const& scr) const
...
             // interfaces
-            auto const jsonServiceInterfaces
-                = JsonValueValidator(jsonServiceInfo, "interfaces", Json::ValueType::arrayValue)();
-            for (auto const& jsonServiceInterface : jsonServiceInterfaces)
+            auto const interfacesValidator = JsonValueValidator(jsonServiceInfo, "interfaces", rapidjson::kArrayType);
+            auto const& jsonServiceInterfaces = interfacesValidator();
+            for (auto const& jsonServiceInterface : jsonServiceInterfaces.GetArray())
             {
-                if (!jsonServiceInterface.isString() || jsonServiceInterface.asString() == "")
+                if (!jsonServiceInterface.IsString() || std::string(jsonServiceInterface.GetString()).empty())
```
jeffdiclemente: minor: `std::string(jsonServiceInterface.GetString())` is unnecessary. It could cause a heap allocation. Consider checking for an empty string using `jsonServiceInterface.GetStringLength()`

jeffdiclemente:
> Open question: @jeffdiclemente @tcormackMW
> 
> * checkDuplicateKeys and toStyledString is duplicated in tools/rc/ResourceCompiler.cpp and compendium/tools/SCRCodeGen/Util.cpp; toStyledString logic is duplicated inline in ResourceCompilerTest.cpp.
>   Should I extract these into a shared header (e.g. third_party/RapidJsonUtils.hpp)?

Yes, duplicate implementations should be consolidated. I suggest putting the consolidated implementations in the top level 'util' folder. The `third_party` folder is only meant for code that we don't own.

jeffdiclemente (changes requested):

jeffdiclemente (approved):

PR #1256, opened 2026-05-04
Add changes required for building with Conan recipe and using vendored dependencies instead of bundled ones

achristoforides:


tcormackMW:
As far as I can see it looks fine. I tried to understand the CMake changes, I wouldn't say I'm an expert -- but they seem to be making changes that make sense.

CMakeLists.txt
```diff
@@ -385,7 +390,68 @@ endif()
 #-----------------------------------------------------------------------------
 
 if(US_USE_SYSTEM_BOOST)
-    find_package(Boost 1.74.0 REQUIRED)
+    if(WIN32)
+        # On Windows, Boost.Nowide has compiled sources.  Create an INTERFACE
+        # wrapper that links the compiled lib and matches the bundled target name.
+        find_package(Boost 1.74.0 REQUIRED)
```
proceduralnoisy: Several comments here:
- There's a lot of duplication between these two code paths (I think the first path just has one extra line). Perhaps more readable to pull the common code out of the conditional blocks?
- `target_link_libraries` should work fine for header-only targets. In other words, I would expect that you could use the first code path regardless of whether boost::boost is a header-only target or a compiled target (CMake should handle the difference under the hood)
- Could all of this perhaps be replaced with something like:
  ```
  find_package(Boost REQUIRED)
  add_library(nowide::nowide ALIAS boost::boost)?
  
  ```

jeffdiclemente (commented):
...resorting to top level comments since line/file comments are current unavailable... 


- **in cmake/CppMicroServicesHelpers.cmake**
  - avoid hard coding the `3` in `usResourceCompiler3` in this file. When we do a major upgrade, these instances need to be manually changed.
  - how will a developer know to add a new `include("${_us_cmake_dir} ...` line if they add a new `.cmake` file in the `cmake/` folder? Consider making this more maintainable.


- there should be no need for a forwarding CL!11 header. The vendored copy should mirror how Conan provides the header in this case

jeffdiclemente (commented):
should a github action be added in this repo to test this build infra?

achristoforides:
> should a github action be added in this repo to test this build infra?

100%. However since the conan recipe is not anywhere at this time (e.g., CCI), we cannot reference it yet. Once a recipe is published we can create an action for it.

jeffdiclemente:
@achristoforides , please merge from the development branch, jsoncpp was recently removed and that change will require changes to this PR

tools/rc/CMakeLists.txt
```diff
@@ -68,6 +73,10 @@ set(CMAKE_ENABLE_EXPORTS ON)
 # US_USE_DETERMINISTIC_BUNDLE_BUILDS activates the locale-setting and sorting logic in
 # ResourceCompiler.cpp that is also required for deterministic output.
 if (US_USE_DETERMINISTIC_BUNDLE_BUILDS)
+  if(US_USE_SYSTEM_MINIZ)
+    message(WARNING "US_USE_DETERMINISTIC_BUNDLE_BUILDS requires bundled miniz (MINIZ_NO_TIME). "
```
jeffdiclemente: per the RFC, this should be a fatal error.

jeffdiclemente (changes requested):

jeffdiclemente:
> > should a github action be added in this repo to test this build infra?
> 
> 100%. However since the conan recipe is not anywhere at this time (e.g., CCI), we cannot reference it yet. Once a recipe is published we can create an action for it.

that's fair, but is there any value to having other conan related github workflows running as part of every PR? for example, auditing the conan managed dependencies for vulnerabilities?

achristoforides:
@copilot resolve the merge conflicts in this pull request

achristoforides:
> > > should a github action be added in this repo to test this build infra?
> > 
> > 
> > 100%. However since the conan recipe is not anywhere at this time (e.g., CCI), we cannot reference it yet. Once a recipe is published we can create an action for it.
> 
> that's fair, but is there any value to having other conan related github workflows running as part of every PR? for example, auditing the conan managed dependencies for vulnerabilities?

I think it will be a non-insignificant amount of work to identify the right tooling and checks to do that we should run as part of our github actions workflows. Today, we wouldn't even have a conan recipe to check against, nor do we have a mechanism in our CMake / general build process that is a manifest of current third-party libraries (incorporating any changes that someone may make within a given PR).

I think this requires more thought and I think should be handled similar to the comment in the corresponding RFC about having a follow-up issue created.

jeffdiclemente (approved):
I'm assuming that these changes have been tested in some way to ensure the "Conan build path" works. If that's the case and the entire cppms project builds successfully using a Conan recipe, this PR is good to go.

achristoforides:
I tested with all build configurations (default, no threading, static, static + no threading) on maca64 and win64. All configurations built and packaged successfully. Additionally, the test_package build/run was successful as well.

PR #1260, opened 2026-05-10
Hoist ToLower(key) out of loop in LDAPExpr::Evaluate

### LINKED ISSUE #1259: Performance of LDAPExpr::Evaluate could be improved
Profiler flagged the ToLower and comparison in this case:

```
else if (pPtr->GetType() == AnyMap::UNORDERED_MAP)
{
    auto value_iter = find_attr_value_in_map<any_map::unordered_any_map>(
        pPtr,
        d->m_attrName,
        [matchCase](AnyMap const* p, std::string const& key)
        {
            auto value_iter = p->findUO_TypeChecked(key);
            if (!matchCase && value_iter == p->endUO_TypeChecked())
            {
                for (auto value_iter = p->beginUO_TypeChecked(); value_iter != p->endUO_TypeChecked();
                        ++value_iter)
                {
                    if (std::string lower = LDAPExpr::ToLower(key); LDAPExpr::ToLower(value_iter->first) == lower)  // <-- bottleneck
                    {
                        return value_iter;
                    }
                }
                return p->endUO_TypeChecked();
            }

            return value_iter;
        },
        [](AnyMap const* p) { return p->endUO_TypeChecked(); });

    if (value_iter)
    {
        return Compare(value_iter.value()->second, d->m_operator, d->m_attrValue);
    }

    return false;
}
```

Looks like this could be hoisted out of the loop:
std::string lower = LDAPExpr::ToLower(key);

bweed-mathworks:
## Summary
- Fixes #1259
- Hoists `LDAPExpr::ToLower(key)` out of the case-insensitive search loops in `LDAPExpr::Evaluate`, eliminating redundant string transformations on every iteration
- Applied to both the `UNORDERED_MAP` and `ORDERED_MAP` code paths

## Test plan
- [x] Project builds cleanly (Release, MSVC 19.36)
- [x] All 27 existing tests pass

### DIFF
--- a/framework/src/util/LDAPExpr.cpp
+++ b/framework/src/util/LDAPExpr.cpp
@@ -552,11 +552,11 @@ namespace cppmicroservices
                         auto value_iter = p->findUO_TypeChecked(key);
                         if (!matchCase && value_iter == p->endUO_TypeChecked())
                         {
+                            std::string lower = LDAPExpr::ToLower(key);
                             for (auto value_iter = p->beginUO_TypeChecked(); value_iter != p->endUO_TypeChecked();
                                  ++value_iter)
                             {
-                                if (std::string lower = LDAPExpr::ToLower(key);
-                                    LDAPExpr::ToLower(value_iter->first) == lower)
+                                if (LDAPExpr::ToLower(value_iter->first) == lower)
                                 {
                                     return value_iter;
                                 }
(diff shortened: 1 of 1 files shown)


jeffdiclemente (approved):

bweed-mathworks:
I don't have permissions to merge from this account, if you were waiting for me to merge it.

Also, how often are the maps empty? This fix that moves the work out of the loop, would pay that cost when the map is completely empty. If that's a problem, we could add a if (p->empty()) check before the first find() ?

jeffdiclemente:
> I don't have permissions to merge from this account, if you were waiting for me to merge it.
> 
> Also, how often are the maps empty? This fix that moves the work out of the loop, would pay that cost when the map is completely empty. If that's a problem, we could add a if (p->empty()) check before the first find() ?

I think the maps are rarely empty. We will monitor the performance and make an adjustment if needed. Thanks.

PR #1263, opened 2026-05-12
I1260 improve ldap performance

mwmcarney:
Made several changes to the LDAPFilter implementation to improve its performance. Specifically:

- added a filter cache in ServiceRegistry
- Pre-lowercased attribute names for the most common case of case insensitive lookups
- Replaced std::function arguments with template parameters to avoid avoid type erasure
- Changed recursive wildcard matching to an iterative algorithm

For the typical ServiceRegistry hot path, expect **~25% faster** evaluation per service checked. The pathological-case fixes (wildcard exponential blowup, large case-insensitive map scans) serve as safety nets — they prevent rare but catastrophic slowdowns rather than improving the common case.


framework/src/service/ServiceRegistry.cpp
```diff
@@ -271,13 +289,13 @@
...
             {
-                ldap = LDAPExpr(filter);
+                ldap = &GetCachedLDAPExpr(filter);
             }
         }
 
         for (; s != send; ++s)
         {
-            if (filter.empty() || ldap.Evaluate(PropertiesHandle((s->d->coreInfo->properties), true), false))
+            if (filter.empty() || ldap->Evaluate(PropertiesHandle((s->d->coreInfo->properties), true), false))
```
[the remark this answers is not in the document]
mwmcarney: Fixed.

tcormackMW (approved):

framework/src/service/ServiceRegistry.cpp
```diff
@@ -208,6 +208,24 @@ namespace cppmicroservices
...
     }
 
+    LDAPExpr const&
+    ServiceRegistry::GetCachedLDAPExpr(std::string const& filter) const
+    {
+        auto it = filterCache.find(filter);
+        if (it != filterCache.end())
+        {
+            return it->second;
+        }
```
achristoforides: Can `it` be `const`? If so, apply.

framework/src/service/ServiceRegistry.h
```diff
@@ -184,6 +186,14 @@ namespace cppmicroservices
                           std::string const& filter,
                           BundlePrivate* bundle,
                           std::vector<ServiceReferenceBase>& serviceRefs) const;
+
+        // Returns a reference valid only until the next cache mutation.
+        // Caller must hold the registry lock (MultiThreaded::Lock).
+        LDAPExpr const& GetCachedLDAPExpr(std::string const& filter) const;
```
achristoforides: Use doxygen style comments

achristoforides (changes requested):
LGTM, if the iterator can be const then make it so.

Otherwise, the feedback is cosmetic but I think important still

jeffdiclemente (approved):

framework/src/service/ServiceRegistry.cpp
mwmcarney: Done.

framework/src/service/ServiceRegistry.h
mwmcarney: Done.

achristoforides (approved):

mwmcarney:
Improved the performance of LDAPFilter.

PR #1264, opened 2026-05-22
1203 race condition when a configuration is removed whilst a ds component with a configuration policyrequired dependency on that config is instantiated

### LINKED ISSUE #1203: Race condition when a Configuration is removed whilst a DS component with a `"configuration-policy":"required"` dependency on that config is instantiated
There is a [window](https://github.com/CppMicroServices/CppMicroServices/blob/28f319493c5341a443667797189de1b725af9374/compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp#L285-L319) during `ComponentConfigurationImpl::ConfigChangedState` where the merged properties have been updated to reflect the removal of a Configuration, but the component has not yet been `Deactivate`d (and hence unregistered). During this window, a concurrent call to `GetService` will be in the process of instantiating the component instance and will call `GetProperties` (e.g. from [here](https://github.com/CppMicroServices/CppMicroServices/blob/28f319493c5341a443667797189de1b725af9374/compendium/ServiceComponent/include/cppmicroservices/servicecomponent/detail/ComponentInstanceImpl.hpp#L429)), and this call will get the properties without the required configuration.

This is quite similar to the issue Jeff fixed in #801 for required references, but in this case it is required Configurations which are missing.

I wonder if we could atomically retrieve the properties and whether required configurations are satisfied (e.g. by adding a new `GetPropertiesAndSatisfiedStatus` method or something like that), and use that when we are creating a new component instance to throw an exception instead of calling the constructor with an `AnyMap` which does not contain all of the properties it is expecting.

I think a test to reproduce this might be possible - if we concurrently call `GetService` whilst removing a required `Configuration`, and in the constructor of the service capture whether the `AnyMap` contained a property that we expected to see from that `Configuration` or not (it is possible it will see that property if the removal of the config hasn't been processed by the time `DoCreate` calls `GetProperties`), some of the time the call to `GetService` would return a valid pointer to an object which was not fully satisfied when it was constructed.

JamesKnee:


compendium/ConfigurationAdmin/test/gtest/TestConfigAdminConfigRace.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include <thread>
+
+namespace
+{
+    auto const SERVICE_TIMEOUT = std::chrono::seconds(5);
+    int const NUM_ITERATIONS = 100;
+    int const NUM_GETSERVICE_ATTEMPTS = 50;
+
+    std::string
+    InstallAndStartBundle(cppmicroservices::BundleContext& ctx, std::string const& bundleName)
```
tcormackMW: This function already exists in testUtils. I would try to use that version as opposed to duplicating the code

compendium/ConfigurationAdmin/test/gtest/TestConfigAdminConfigRace.cpp
```diff
@@ -0,0 +1,166 @@
...
+        {
+            b.Start();
+        }
+#endif
+        return path;
+    }
+} // namespace
+
+/**
+ * Tests that a race condition between removing a required Configuration and
```
tcormackMW: May be worth decoupling this comment from the exact case. I think it is good to provide these details potentially, but I would argue tests should verify functionality rather than a specific line of code they are testing...

i.e. verify that getService concurrent with a removed configuration does xxxxxx.

compendium/ConfigurationAdmin/test/gtest/TestConfigAdminConfigRace.cpp
```diff
@@ -0,0 +1,166 @@
...
+ * call DoCreate -> GetProperties and get properties without the required
+ * configuration.
+ */
+TEST_F(tGenericDSAndCASuite, testGetServiceDuringConfigRemovalRace)
+{
+    std::string const configPid { "sample::ServiceComponentCA29" };
+
+    InstallAndStartBundle(context, "TestBundleDSCA29");
+
+    int const iterations = NUM_ITERATIONS;
```
tcormackMW: file level globals for NUM_ITERATIONS rather than test specific counts?

compendium/ConfigurationAdmin/test/gtest/TestConfigAdminConfigRace.cpp
```diff
@@ -0,0 +1,166 @@
...
+                                                               return true;
+                                                           }
+                                                       }
+                                                   }
+                                                   std::this_thread::yield();
+                                               }
+                                               return false;
+                                           });
+
+        go.set_value();
```
tcormackMW: use a barrier instead. we already define one of these in compendium/DeclarativeServices/test/gtest/ConcurrencyTestUtil.hpp

See if you can reference that one directly or have the implementation in both CA and DS. Good to use built objects rather than promises and futures directly imo.

compendium/DeclarativeServices/src/manager/ConfigurationManager.cpp
```diff
@@ -147,6 +147,22 @@ namespace cppmicroservices
...
 
+        bool
+        ConfigurationManager::WouldDeletionUnsatisfy(std::string const& pid) const noexcept
+        {
+            std::lock_guard<std::mutex> lock(propertiesMutex);
+            if (metadata->configurationPolicy != CONFIG_POLICY_REQUIRE)
+            {
+                return false;
+            }
+            auto it = configProperties.find(pid);
```
tcormackMW, line 160: some small comments in this function to explain what you're checking:

1. if not required, it wouldn't unsatisfy
2. if we don't currently track this config, would not unsatisfy
3. if we have multiple configs that satisfy, it would not unsatisfy

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
```diff
@@ -314,10 +336,8 @@ namespace cppmicroservices
                     }
                     break;
                 case cppmicroservices::service::cm::ConfigurationEventType::CM_DELETED:
-                    if (configWasSatisfied && !configNowSatisfied)
-                    {
-                        Deactivate();
-                    }
+                    // Non-unsatisfying deletion (e.g., optional policy or redundant config).
```
tcormackMW: I personally would remove this switch entirely and deal with both branches (unsatisfying and non unsatisfying) in the above if. 

```
i.e. 
if (deletion) {
    if (unsatisfying) {
        do x
     } else {
         do y
     }
}
````
this keeps like logic together. A bit more readable in my eyes

File: compendium/test_bundles/TestBundleDSCA29/src/ServiceImpl.hpp [hunk 7]
```diff
@@ -0,0 +1,29 @@
+#ifndef _SERVICE_IMPL_HPP_
```
jeffdiclemente, line 1: minor: use of C++ reserved `_` prefix and include guard collision risk; 
`#ifndef _SERVICE_IMPL_HPP_` This guard is generic and could collide with other test bundles. Should be `SERVICE_IMPL_CA29_HPP_` The `#endif` comment already says `_SERVICE_IMPL_CA29_HPP_` which is inconsistent with the actual guard.

compendium/test_bundles/TestBundleDSCA29/src/ServiceImpl.cpp
```diff
@@ -0,0 +1,26 @@
+#include "ServiceImpl.hpp"
+
+namespace sample
+{
+
+    std::atomic<int> ServiceComponentCA29::constructionCount { 0 };
+    std::atomic<int> ServiceComponentCA29::missingPropertyCount { 0 };
```
jeffdiclemente: these are incremented, but never otherwise used as far as I can see.  should be removed or used somewhere else in this change,.

compendium/ConfigurationAdmin/test/gtest/TestConfigAdminConfigRace.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include <cppmicroservices/BundleContext.h>
+#include <cppmicroservices/Framework.h>
+#include <cppmicroservices/FrameworkFactory.h>
+#include <cppmicroservices/cm/ConfigurationAdmin.hpp>
+
+#include "TestInterfaces/Interfaces.hpp"
+
+#include "TestFixtures.hpp"
+#include "../TestUtils.hpp"
+#include "../../../DeclarativeServices/test/gtest/ConcurrencyTestUtil.hpp"
```
jeffdiclemente, line 35: minor: this creates a tight coupling between the DS folder structure and this test. this test header should probably be moved to a more generic location that both DS tests and CA tests can reference.  this can be done after and is not blocking this PR.

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
```diff
@@ -305,22 +328,9 @@ namespace cppmicroservices
...
-                    break;
-                case cppmicroservices::service::cm::ConfigurationEventType::CM_DELETED:
-                    if (configWasSatisfied && !configNowSatisfied)
-                    {
-                        Deactivate();
-                    }
-                    break;
-                default:
-                    break;
+                Register();
```
jeffdiclemente, line 336: minor: The logic used to do nothing (i.e. no `Register`/`Deactivate`) if the `ConfigurationEventType` was unknown. It seems highly unlikely that the `ConfigurationEventType` enum will be modified, so maintaining this behavior is only useful if a new `ConfigurationEventType` enum value will be introduced.

jeffdiclemente (changes requested):

compendium/test_bundles/TestBundleDSCA29/src/ServiceImpl.hpp
JamesKnee: Would this be something that we want to change across all of the test bundles? All bundles use the `#ifndef _SERVICE_IMPL_HPP_`. The comment at the end was a typo I made, btu I can update it accordingly.

jeffdiclemente: yes, and I don't recommend doing it in this PR 😃

JamesKnee: Sounds good, I don't mind making another PR later today/early next week to clean that up across all the bundles :)

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
```diff
@@ -278,6 +278,29 @@ namespace cppmicroservices
...
+            if (notification.event == cppmicroservices::service::cm::ConfigurationEventType::CM_DELETED)
+            {
+                if (configManager->WouldDeletionUnsatisfy(notification.pid))
+                {
+                    // Deactivate first (closes the latch, drains in-flight GetService calls),
+                    // then update properties as bookkeeping.
+                    Deactivate();
+                }
+
+                bool configWasSatisfied = false;
```
tcormackMW, line 293: do we need to declare these bools? Can we just pass in 'false' directly? These are vars passed by reference used to control flow after the return of `updateMergedProperties` but given that we don't use these in this branch of code, I'm not sure it is worth declaring them.

One could argue that it is important to define them for code clarity, but given that we don't use them at all in this branch, does it really add clarity or does it just make me think "what the heck are these vars for"

tcormackMW (commented):
Aside from my one comment, which perhaps should be ignored -- those lines just gave me code smell - it looks good, pending jeff's comments, should be good to go

compendium/DeclarativeServices/src/manager/ComponentConfigurationImpl.cpp
JamesKnee: Since the function is looking for a bool reference, so unfortunately we can't just pass 'false' in directly. The two options that come to mind for me are:

1. Rename the variables to signify they are unused by adding something like an `unused_` prefix to the variable names
2. Overload the `UpdateMergedProperties` that does use the out-params so that we can have a cleaner function call

tcormackMW: mm that's fair. ok no worries, it was certainly a bit of a nit pick. If we want to be complete, we can just throw a comment above them that says 'these aren't used' but it's not a big deal either way

JamesKnee: Sounds good I will add that comment now

jeffdiclemente: there is another option, annotate the bools with `[[maybe_unused]]`  https://cppreference.com/cpp/language/attributes/maybe_unused. Doing this also removes the need for a comment (its "self-documenting").

jeffdiclemente (approved):

PR #1265, opened 2026-06-02
Any map refactor

mwmcarney:
# Plan: Refactor AnyMap from Manual Union to std::variant

**Status: COMPLETE** (2026-06-02)

## Context

The `any_map`/`AnyMap` class in CppMicroServices uses a hand-rolled discriminated union with heap-allocated map pointers, manual lifecycle management, and heap-allocated iterators. This results in ~1100 lines of boilerplate, a heap allocation per map construction and per iterator creation, and pointer indirection on every access. We're replacing this with `std::variant`-based inline storage per the decisions in `docs/adr/0001-anymap-variant-refactor.md`.

## Critical Files

| File                                          | Role                                         |
|-----------------------------------------------|----------------------------------------------|
| `framework/include/cppmicroservices/AnyMap.h` | Header — full rewrite of class layout        |
| `framework/src/util/AnyMap.cpp`               | Implementation — ~1304→~450 lines            |
| `framework/src/util/Properties.cpp`           | ~20 TypeChecked call sites → `std::get<T>()` |
| `framework/src/util/LDAPExpr.cpp`             | ~15 TypeChecked call sites → `std::get<T>()` |
| `framework/test/gtest/AnyMapTest.cpp`         | 2 death tests to update + unused var fixes   |
| `framework/src/bundle/BundleManifest.cpp`     | 1 iterator type mismatch fix (discovered during build) |

### Step 1: Rewrite AnyMap.h as a single collapsed class with variant storage

Replace the entire `any_map` base class + `AnyMap` subclass hierarchy with a single `AnyMap` class. The new layout:

```cpp
class US_Framework_EXPORT AnyMap {
public:
    // Existing type aliases (unchanged)
    using ordered_any_map = std::map<std::string, Any>;
    using unordered_any_map = std::unordered_map<std::string, Any>;
    using unordered_any_cimap = std::unordered_map<std::string, Any, detail::any_map_cihash, detail::any_map_ciequal>;
    using map_variant = std::variant<ordered_any_map, unordered_any_map, unordered_any_cimap>;
    enum map_type : uint8_t { ORDERED_MAP, UNORDERED_MAP, UNORDERED_MAP_CASEINSENSITIVE_KEYS };

    // Variant-based iterator classes
    class US_Framework_EXPORT const_iter { std::variant<std::monostate, ociter, uociter, uocciiter> it_; ... };
    class US_Framework_EXPORT iter { std::variant<std::monostate, oiter, uoiter, uociiter> it_; ... };

    // Constructors (delegate to variant)
    AnyMap(std::initializer_list<value_type> l = {});  // defaults to CI map
    AnyMap(map_type type, std::initializer_list<value_type> l = {});
    AnyMap(ordered_any_map const& m);
    // ... move/copy/from-each-map-type constructors

    // Defaulted special members (variant handles lifecycle)
    AnyMap(AnyMap const&) = default;
    AnyMap& operator=(AnyMap const&) = default;
    AnyMap(AnyMap&&) noexcept = default;
    AnyMap& operator=(AnyMap&&) noexcept = default;
    ~AnyMap() = default;

    // Container interface (switch on map_.index())
    // AtCompoundKey, GetType (maps variant.index() → enum)

    // Public storage for direct access by Properties/LDAPExpr
    map_variant map_;
};

using any_map = AnyMap;  // source-compatibility alias
```

### Step 2: Rewrite AnyMap.cpp

- **Iterators**: Replace 460 lines of heap-alloc union code with ~100 lines of index-based switch dispatch.
- **Constructors**: Each becomes a one-liner emplacing the right variant alternative. No `new`/`delete`.
- **Copy/move/destructor**: Removed (defaulted).
- **Method dispatch** (`at`, `find`, `size`, `empty`, etc.): Each uses `switch (map_.index())` with `std::get<N>()`. Methods returning iterators construct the variant via `std::in_place_index<N+1>` (offset by 1 for `std::monostate` at index 0).
- **TypeChecked functions**: Deleted entirely.
- **`o_m()`/`uo_m()`/`uoci_m()` private accessors**: Deleted.
- **AtCompoundKey and serialization**: Logic unchanged, just operates on `map_` via `std::visit` or `std::get`.
- **`operator==`**: `map_ == rhs.map_` (variant equality handles type mismatch).

### Step 3: Update Properties.cpp (~20 call sites)

Transform pattern:
```cpp
// Before:
for (auto itr = props.beginOM_TypeChecked(); itr != props.endOM_TypeChecked(); ++itr)

// After:
auto const& om = std::get<AnyMap::ordered_any_map>(props.map_);
for (auto itr = om.begin(); itr != om.end(); ++itr)
```

### Step 4: Update LDAPExpr.cpp (~15 call sites)

Transform pattern:
```cpp
// Before:
[](AnyMap const* p, std::string const& key) { return p->findUOCI_TypeChecked(key); }

// After:
[](AnyMap const* p, std::string const& key) {
    return std::get<AnyMap::unordered_any_cimap>(p->map_).find(key);
}
```

### Step 5: Update AnyMapTest.cpp (2 death tests + 3 unused variable fixes)

Lines 210 and 233: The current `ASSERT_DEATH({ o.size(); }, ".*")` expects a crash when calling `size()` on a moved-from object. With `std::variant`, a moved-from map is valid-but-empty (well-defined state). Change to:
```cpp
ASSERT_EQ(o.size(), 0u);  // moved-from variant holds valid empty map
```

Also removed `testing::FLAGS_gtest_death_test_style = "threadsafe"` from both tests, and added `(void)` casts for 3 variables that exist solely to test iterator constructors (suppresses `-Werror,-Wunused-variable`).

### Step 6: Fix BundleManifest.cpp (discovered during build)

Line 235 used `AnyMap::const_iterator` to iterate over `m_PropertiesDeprecated`, which is a `std::map<std::string, Any>` (not an `AnyMap`). Previously this compiled because the old `AnyMap::const_iterator` had implicit conversion from raw STL iterators. Fixed by replacing with `auto`.

### Step 7: Introduce `ci_unordered_map` wrapper class with distinct iterator types

Replaced the bare type alias:
```cpp
using unordered_any_cimap = std::unordered_map<std::string, Any, detail::any_map_cihash, detail::any_map_ciequal>;
```

With a wrapper class `ci_unordered_map` that:
1. Contains the `std::unordered_map<std::string, Any, any_map_cihash, any_map_ciequal>` as a private member
2. Provides its own `iterator` and `const_iterator` wrapper classes (distinct types from the raw STL iterators)
3. Forwards the full container interface (begin/end/find/insert/emplace/erase/at/operator[]/size/empty/count/clear)
4. Exposes standard type aliases (`hasher`, `key_equal`, `value_type`, etc.)
5. Supports initializer-list construction and defaulted copy/move

The type alias is preserved for API compatibility:
```cpp
using unordered_any_cimap = ci_unordered_map;
```

**Benefits:**
- The `map_variant` now holds three genuinely distinct types — no ambiguity when constructing the variant
- The *iterator* types are now distinct: `ci_unordered_map::iterator` / `ci_unordered_map::const_iterator` are wrapper classes, not aliases to `__hash_map_iterator`
- This eliminates the need for `std::in_place_index<N>` when constructing the iterator variant — type-based deduction works
- All iterator-returning methods (`begin`, `end`, `find`, `insert`, `emplace`) simplified from switch statements to one-line `std::visit` lambdas
- The iter→const_iter converting constructor now uses `std::visit` with type-based dispatch instead of index-based switch
- Hash/equality functors (`any_map_cihash`, `any_map_ciequal`) are encapsulated as implementation details
- `std::get<AnyMap::unordered_any_cimap>(map_)` works cleanly since `ci_unordered_map` is a unique type in the variant

**No changes required** in `Properties.cpp`, `LDAPExpr.cpp`, `AnyMapTest.cpp`, or `BundleManifestTest.cpp` — the `unordered_any_cimap` type alias and all existing call patterns continue to work unchanged.

### Apple libc++ iterator type aliasing (resolved)

On Apple's libc++, `std::unordered_map<K, V>::iterator` and `std::unordered_map<K, V, CustomHash, CustomEq>::iterator` resolve to the same underlying `__hash_map_iterator` type. This initially prevented:

1. Separate constructors for `const_iter(uociter)` and `const_iter(uocciiter)` — same signature.
2. `std::visit` on the iter variant — visitor can't distinguish same-type alternatives.

**Initial workaround** (Steps 1-6): `std::in_place_index<N>` for explicit index-based variant construction, and index-based switch for iter→const_iter conversion.

**Final solution** (Step 7): `ci_unordered_map` provides its own `iterator`/`const_iterator` wrapper classes, making all three iterator types in the variant genuinely distinct. This eliminated the `std::in_place_index` workaround entirely and enabled clean `std::visit`-based dispatch for all iterator-returning methods.

### ABI-breaking change required full rebuild

All downstream test binaries and shared libraries linked against the old AnyMap symbols. A full `cmake --build .` was required after the refactor — incremental builds of just `usFrameworkTests` would link but other binaries would crash with dyld symbol-not-found errors until rebuilt.

## Verification

```bash
# Build everything (full rebuild required — ABI break)
cmake --build build/maca64

# Run AnyMap + LDAP tests specifically
bin/usFrameworkTests --gtest_filter="*AnyMap*:*LDAP*"

# Run full framework test suite (398 tests)
bin/usFrameworkTests

# Run benchmarks
bin/usFrameworkBenchTests --benchmark_filter="AnyMap"

# Full ctest suite (27 test programs)
ctest
```

## Actual Outcome

- **Header**: ~528 lines → ~367 lines (includes ~100 lines for `ci_unordered_map` wrapper with iterator classes)
- **Implementation**: ~1304 lines → ~609 lines (includes serialization unchanged at ~100 lines)
- **Net reduction**: ~855 lines removed
- **Test results**: 398/398 framework tests pass, 27/27 ctest programs pass
- **Benchmarks**: Run successfully (debug build — no meaningful perf comparison, but no regressions observed)
- **Performance**: No heap allocations for maps or iterators, no pointer indirection
- **API compatibility**: Source-compatible via `using any_map = AnyMap` alias
- **ABI**: Breaking (full rebuild performed)

### DIFF
--- a/framework/include/cppmicroservices/AnyMap.h
+++ b/framework/include/cppmicroservices/AnyMap.h
@@ -118,7 +118,7 @@ namespace cppmicroservices
         ~ci_unordered_map() = default;
 
         iterator begin() noexcept { return {map_.begin()}; }
-        const_iterator begin() const noexcept { return const_iterator(map_.begin()); }
+        const_iterator begin() const noexcept { return {map_.begin()}; }
         const_iterator cbegin() const noexcept { return const_iterator(map_.cbegin()); }
         iterator end() noexcept { return iterator(map_.end()); }
         const_iterator end() const noexcept { return const_iterator(map_.end()); }
--- a/framework/include/cppmicroservices/AnyMap.h
+++ b/framework/include/cppmicroservices/AnyMap.h
@@ -119,7 +119,7 @@ namespace cppmicroservices
 
         iterator begin() noexcept { return {map_.begin()}; }
         const_iterator begin() const noexcept { return {map_.begin()}; }
-        const_iterator cbegin() const noexcept { return const_iterator(map_.cbegin()); }
+        const_iterator cbegin() const noexcept { return {map_.cbegin()}; }
         iterator end() noexcept { return iterator(map_.end()); }
         const_iterator end() const noexcept { return const_iterator(map_.end()); }
         const_iterator cend() const noexcept { return const_iterator(map_.cend()); }
--- a/framework/include/cppmicroservices/AnyMap.h
+++ b/framework/include/cppmicroservices/AnyMap.h
@@ -120,7 +120,7 @@ namespace cppmicroservices
         iterator begin() noexcept { return {map_.begin()}; }
         const_iterator begin() const noexcept { return {map_.begin()}; }
         const_iterator cbegin() const noexcept { return {map_.cbegin()}; }
-        iterator end() noexcept { return iterator(map_.end()); }
+        iterator end() noexcept { return {map_.end()}; }
         const_iterator end() const noexcept { return const_iterator(map_.end()); }
         const_iterator cend() const noexcept { return const_iterator(map_.cend()); }
 
--- a/framework/include/cppmicroservices/AnyMap.h
+++ b/framework/include/cppmicroservices/AnyMap.h
@@ -48,6 +48,113 @@ namespace cppmicroservices
 
     } // namespace detail
 
+    class US_Framework_EXPORT ci_unordered_map
+    {
+        using map_type = std::unordered_map<std::string, Any, detail::any_map_cihash, detail::any_map_ciequal>;
+        map_type map_;
+
+      public:
+        using key_type = map_type::key_type;
+        using mapped_type = map_type::mapped_type;
+        using value_type = map_type::value_type;
+        using size_type = map_type::size_type;
+        using difference_type = map_type::difference_type;
+        using hasher = map_type::hasher;
+        using key_equal = map_type::key_equal;
+        using reference = map_type::reference;
+        using const_reference = map_type::const_reference;
+
+        class const_iterator
+        {
+            map_type::const_iterator it_;
+
+          public:
+            using value_type = ci_unordered_map::value_type;
+            using reference = ci_unordered_map::const_reference;
+            using pointer = value_type const*;
+            using difference_type = ci_unordered_map::difference_type;
+            using iterator_category = std::forward_iterator_tag;
+
+            const_iterator() = default;
+            const_iterator(map_type::const_iterator it) : it_(it) {}
+
+            reference operator*() const { return *it_; }
+            pointer operator->() const { return &(*it_); }
+            const_iterator& operator++() { ++it_; return *this; }
+            const_iterator operator++(int) { auto tmp = *this; ++it_; return tmp; }
+            bool operator==(const_iterator const& o) const { return it_ == o.it_; }
+            bool operator!=(const_iterator const& o) const { return it_ != o.it_; }
+        };
+
+        class iterator
+        {
+            map_type::iterator it_;
+
+          public:
+            using value_type = ci_unordered_map::value_type;
+            using reference = ci_unordered_map::reference;
+            using pointer = value_type*;
+            using difference_type = ci_unordered_map::difference_type;
+            using iterator_category = std::forward_iterator_tag;
+
+            iterator() = default;
+            iterator(map_type::iterator it) : it_(it) {}
+            operator const_iterator() const { return const_iterator(map_type::const_iterator(it_)); }
+
+            reference operator*() const { return *it_; }
+            pointer operator->() const { return &(*it_); }
+            iterator& operator++() { ++it_; return *this; }
+            iterator operator++(int) { auto tmp = *this; ++it_; return tmp; }
+            bool operator==(iterator const& o) const { return it_ == o.it_; }
+            bool operator!=(iterator const& o) const { return it_ != o.it_; }
+        };
+
+        ci_unordered_map() = default;
+        ci_unordered_map(std::initializer_list<value_type> il) : map_(il) {}
+        ci_unordered_map(ci_unordered_map const&) = default;
+        ci_unordered_map(ci_unordered_map&&) noexcept = default;
+        ci_unordered_map& operator=(ci_unordered_map const&) = default;
+        ci_unordered_map& operator=(ci_unordered_map&&) noexcept = default;
+        ~ci_unordered_map() = default;
+
+        iterator begin() noexcept { return iterator(map_.begin()); }
+        const_iterator begin() const noexcept { return const_iterator(map_.begin()); }
+        const_iterator cbegin() const noexcept { return const_iterator(map_.cbegin()); }
+        iterator end() noexcept { return iterator(map_.end()); }
+        const_iterator end() const noexcept { return const_iterator(map_.end()); }
+        const_iterator cend() const noexcept { return const_iterator(map_.cend()); }
+
+        bool empty() const noexcept { return map_.empty(); }
+        size_type size() const noexcept { return map_.size(); }
+        size_type count(key_type const& key) const { return map_.count(key); }
+        void clear() noexcept { map_.clear(); }
+
+        mapped_type& at(key_type const& key) { return map_.at(key); }
+        mapped_type const& at(key_type const& key) const { return map_.at(key); }
+        mapped_type& operator[](key_type const& key) { return map_[key]; }
+        mapped_type& operator[](key_type&& key) { return map_[std::move(key)]; }
+
+        std::pair<iterator, bool> insert(value_type const& value)
+        {
+            auto p = map_.insert(value);
+            return { iterator(p.first), p.second };
+        }
+
+        template <class... Args>
+        std::pair<iterator, bool> emplace(Args&&... args)
+        {
+            auto p = map_.emplace(std::forward<Args>(args)...);
+            return { iterator(p.first), p.second };
+        }
+
+        iterator find(key_type const& key) { return iterator(map_.find(key)); }
+        const_iterator find(key_type const& key) const { return const_iterator(map_.find(key)); }
+        size_type erase(key_type const& key) { return map_.erase(key); }
+
+        bool operator==(ci_unordered_map const& rhs) const { return map_ == rhs.map_; }
+        bool operator!=(ci_unordered_map const& rhs) const { return !(*this == rhs); }
+    };
+
     /**
      * \ingroup MicroServicesUtils
      *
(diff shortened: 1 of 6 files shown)


jeffdiclemente (commented):
can you submit the design document (`0001-anymap-variant-refactor.md`) to https://github.com/CppMicroServices/rfcs ?

framework/include/cppmicroservices/AnyMap.h
```diff
@@ -282,232 +313,40 @@ namespace cppmicroservices
...
-         * \endcode
-         *
-         * @param key The key hierachy to query.
-         * @param defaultValue is the value to be returned if the key is not found
-         * @return A copy of the key's value.
-         */
         mapped_type AtCompoundKey(key_type const& key, mapped_type defaultValue) const noexcept;
+
+        // ----- Public variant storage -----
+        map_variant map_;
```
jeffdiclemente: this breaks encapsulation. I'd rather have `friend` declarations or a type safe accessor method than expose an internal data member.

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -199,16 +202,14 @@ TEST(AnyMapTest, AnyMap)
...
     AnyMap o = {
         AnyMap::ORDERED_MAP,
         {{ "do", 1 }, { "re", 2 }}
     };
 
     AnyMap o_anymap_move_ctor(std::move(o));
     ASSERT_EQ(any_cast<int>(o_anymap_move_ctor.at("do")), 1);
-    ASSERT_DEATH({ o.size(); }, ".*") << "This call should result in a crash because "
-                                         "the object has been moved from";
+    ASSERT_EQ(o.size(), 0u) << "Moved-from AnyMap should be valid but empty";
```
jeffdiclemente: the C++ standard guarantees a `std::map` or `std::unordered_map` is valid, but not that its size == 0. Consider relaxing the check to asserting that `o.size()` doesn't throw or the moved from map is still valid or enforce a size of 0 in the move coinstructor/assignment operator.

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -230,8 +230,7 @@ TEST(AnyMapTest, MoveAssignment)
     AnyMap uo_anymap_move_assign(AnyMap::UNORDERED_MAP);
     uo_anymap_move_assign = std::move(o);
     ASSERT_EQ(any_cast<int>(uo_anymap_move_assign.at("re")), 2);
-    ASSERT_DEATH(o.size(), ".*") << "This call should result in a crash because "
-                                    "the object has been moved from";
+    ASSERT_EQ(o.size(), 0u) << "Moved-from AnyMap should be valid but empty";
```
jeffdiclemente: the C++ standard guarantees a `std::map` or `std::unordered_map` is valid, but not that its size == 0. Consider relaxing the check to asserting that `o.size()` doesn't throw or the moved from map is still valid or enforce a size of 0 in the move coinstructor/assignment operator.

framework/src/util/AnyMap.cpp
```diff
@@ -177,1002 +177,325 @@ namespace cppmicroservices
...
-    any_map::operator=(any_map&& m) noexcept
+    AnyMap::map_type
+    AnyMap::GetType() const
     {
-        destroy();
-        type = m.type;
-        move_from(std::move(m));
-        return *this;
+        static constexpr map_type types[] = { ORDERED_MAP, UNORDERED_MAP, UNORDERED_MAP_CASEINSENSITIVE_KEYS };
+        return types[map_.index()];
```
jeffdiclemente: minor: If the map variant types are ever re-ordered, this breaks. Consider some manner of check, like a `static_assert` on the ordering, to alert developers changing this code.

jeffdiclemente (changes requested):

mwmcarney:
> can you submit the design document (`0001-anymap-variant-refactor.md`) to https://github.com/CppMicroServices/rfcs ?

Will do.

framework/include/cppmicroservices/AnyMap.h
mwmcarney: Done.

framework/test/gtest/AnyMapTest.cpp
mwmcarney: Done.

framework/test/gtest/AnyMapTest.cpp
mwmcarney: Done.

framework/src/util/AnyMap.cpp
mwmcarney: Done. The refactor changes the code to do a lookup based on C++ type returning the correct enumeration accordingly. No static_assert is necessary.

tcormackMW (approved):

tcormackMW:
LGTM. Let's definitely wait on Jeff to make sure he agrees.

mwmcarney:
> > can you submit the design document (`0001-anymap-variant-refactor.md`) to https://github.com/CppMicroServices/rfcs ?
> 
> Will do.

Pull request has been created in rfcs with that document: https://github.com/CppMicroServices/rfcs/pull/29

framework/test/gtest/AnyMapTest.cpp
```diff
@@ -59,14 +59,17 @@ TEST(AnyMapTest, IteratorTest)
...
 
     AnyMap uoci {
         {{ "do", 1 }, { "re", 2 }}
     };
     AnyMap::const_iter uoccciiter(uoci.begin());
     AnyMap::const_iter uoccciiter1(uoci.cbegin());
+    (void)uoccciiter1;
 
     AnyMap::const_iter ociter_temp(AnyMap(AnyMap::ORDERED_MAP).cbegin());
+    (void)ociter_temp;
```
jeffdiclemente: minor: use the `[[maybe_unused]]` attribute instead (modern c++[17]).

jeffdiclemente (approved):

achristoforides (approved):
LGTM :)

framework/test/gtest/AnyMapTest.cpp
mwmcarney: Done.

PR #1271, opened 2026-06-29
updateIfDifferent(emptyMap) on first update SHOULD cause notification

tcormackMW:
According to OSGI: [When no update has occurred since this object was created, getProperties returns null](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.cm.html#service.cm-updating.configuration:~:text=When%20no%20update%20has%20occurred%20since%20this%20object%20was%20created%2C%20getProperties%20returns%20null)

This means that a series of calls like this:

```
configAdmin->GetConfiguration("UpdateIfDifferent")->UpdateIfDifferent().second.get();
```

SHOULD update the configuration and trigger a notification because `emptyAnymap != null`

### DIFF
--- a/compendium/ConfigurationAdmin/src/ConfigurationImpl.cpp
+++ b/compendium/ConfigurationAdmin/src/ConfigurationImpl.cpp
@@ -224,7 +224,10 @@ namespace cppmicroservices
             {
                 throw std::runtime_error(REMOVED_EXCEPTION_MESSAGE);
             }
-            if (properties == newProperties)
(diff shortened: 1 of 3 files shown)


jeffdiclemente (approved):
looks like there's an unrelated issue on macOS that is probably a result of the runner's compiler being upgraded; https://github.com/CppMicroServices/CppMicroServices/actions/runs/28874877691/job/85647143210?pr=1271. This can be fixed separately if needed.

PR #1274, opened 2026-07-17
update rapidjson to latest from github repo

tcormackMW:
This pulls in the latest code from rapidjson. This is still technically under v1.1, but only because they have not released new tars with updated code from the last 10 years (we are currently on a version that is from ~2019)


jeffdiclemente (commented):
many Windows build configs are failing that need an investigation.

jeffdiclemente (approved):

PR #1278, opened 2026-07-23
localize locking more to remove deadlock potential

### LINKED ISSUE #1192: Out of order mutex locking
Spawned from #1112 

https://github.com/CppMicroServices/CppMicroServices/pull/1112#issuecomment-3145584286

The bug here is described below:

There are two services, **serviceA** and **serviceB**. **serviceA** depends on **serviceB**. The two threads in the above comment:

**THREAD 1**
**serviceA** is is being `Activate`d asynchronously because its reference to `serviceB` has become satisfied. In doing its `Activate`ion, it calls `InitializeServicesCache` which initializes the member variable `boundServicesCache` which stores the _referenced_ services that are bound to this service. For each referenced interface, `boundServicesCache`, which is a `Guarded` variable, is locked (**Lock Mutex 2**) and `GetService` is called on the _referenced_ `ServiceObject`. 

In the call to `GetService`, eventually `Activate` is invoked on the `CCActiveState` of **serviceB**, the _referenced_ service. Within the call to `serviceB::CCActiveState::Activate`, we lock the `oneAtAtime` mutex  (**Lock Mutex 1**).

**THREAD 2**
**serviceB** is modified. The `CCActiveState::Modified` method is invoked (**Lock Mutex 1**). As a result, all `ServiceListeners` are alerted to the changed service. **serviceA** is one of those listeners. As a result of the changed property of **serviceB**, **serviceA** becomes unsatisfied and through `CCActiveState::Deactivate` the `SingletonComponentConfiguration::DestroyComponentInstances` is invoked. This method clears the `ComponentContextImpl` associated with the _referencing_ singleton (**serviceA**), invoking the destructor for that object which invokes the destructor of the `Guarded` `boundServicesCache` object of `serviceA`  (**Lock Mutex 2**).

tcormackMW:
### Thread 1: serviceA rebinding, fetches serviceB

```
ComponentContextImpl::AddToBoundServicesCache(serviceA)
  <- LOCK boundServicesCache (serviceA)
  └─ ServiceObjects::GetService
     └─ SingletonComponentConfigurationImpl::GetService(serviceB)
        └─ CCActiveState::Activate(serviceB)
           <- LOCK oneAtATimeMutex (serviceB)
```

### Thread 2: serviceB is Modified, cascades into serviceA destruction

```
CCActiveState::Modified(serviceB)
  <- LOCK oneAtATimeMutex (serviceB)
  └─ SetRegistrationProperties
     └─ ServiceListeners::ServiceChanged
        └─ serviceA becomes unsatisfied
           └─ Deactivate(serviceA)
              └─ DestroyComponentInstances
                 └─ ComponentContextImpl::Invalidate(serviceA)
                    <- LOCK boundServicesCache (serviceA)
```

### Fix

In `AddToBoundServicesCache` (and `InitializeServicesCache`), call `GetService` before locking `boundServicesCache`. Lock only for the insert.
@Burgch To answer your points from the issue linked, #1192, 

1. "Do we need to be holding a lock on the boundServicesCache whilst we're calling GetService..." -- no I think you are right. We do not technically need to. The problem is it is a `guarded` data member which means we are forced to lock it at some point for access, but it does NOT need to be around the `getService` call. 
2. "should be queued to run after the Activate..." I am not sure if this is a question or a statement, but I am guessing it is a statement because that is what I believe after reading back through code. The `Deactivate` will be blocked till the `Activate` transition task is finished.

### DIFF
--- a/compendium/DeclarativeServices/src/ComponentContextImpl.cpp
+++ b/compendium/DeclarativeServices/src/ComponentContextImpl.cpp
@@ -76,15 +76,16 @@ namespace cppmicroservices::scrimpl
                               {
                                   ServiceReferenceU sRefU(sRef);
                                   auto bc = GetBundleContext();
-                                  auto boundServicesCacheHandle = boundServicesCache.lock();
-                                  auto& serviceMap = (*boundServicesCacheHandle)[refName];
-                                  // get us the service instance
+                                  // Acquire the service outside the boundServicesCache lock to
+                                  // avoid lock inversion with oneAtATimeMutex (see issue #1192).
                                   cppmicroservices::ServiceObjects<void> sObjs = bc.GetServiceObjects(sRefU);
                                   auto interfaceMap = sObjs.GetService();
                                   if (interfaceMap && !interfaceMap->empty())
                                   {
                                       foundAtLeastOneValidBoundService = true;
-                                      serviceMap.emplace_back(std::make_pair(sRef, interfaceMap));
+                                      auto boundServicesCacheHandle = boundServicesCache.lock();
+                                      (*boundServicesCacheHandle)[refName].emplace_back(
+                                          std::make_pair(sRef, interfaceMap));
                                   }
                               }
                           });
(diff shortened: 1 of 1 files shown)


jeffdiclemente (approved):

Burgch (commented):
Thanks @tcormackMW !

PR #1281, opened 2026-07-24
Bundle lifecycle

howellx:


framework/src/bundle/BundlePrivate.cpp
```diff
@@ -66,659 +73,112 @@ namespace cppmicroservices
...
-        auto frameworkBlock = coreCtx->GetFrameworkStateAndBlock();
-        if (frameworkBlock->frameworkHasStopped)
-        {
-            throw std::runtime_error("Bundle " + symbolicName + " (location=" + location
-                                     + ") belongs to a stopped framework");
-        }
-        if (state == Bundle::STATE_UNINSTALLED)
+        // Make sure that bundleContext is invalid
+        std::shared_ptr<BundleContextPrivate> ctx;
+        if ((ctx = bundleContext.Exchange(ctx)))
```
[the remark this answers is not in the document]
tcormackMW: Is this a real warning? This certainly seems scary to me

framework/src/bundle/states/BPStartingState.h
```diff
@@ -0,0 +1,24 @@
...
+        public:
+        using BundlePrivateState::BundlePrivateState;
+        virtual ~BPStartingState() = default;
+
+        void Start(BundlePrivate& mgr, uint32_t options) override;
+        void Stop(BundlePrivate& mgr, uint32_t options) override;
+        void Uninstall(BundlePrivate& mgr) override;
+
+        uint32_t GetState() override {
+            return 0x00000008;
```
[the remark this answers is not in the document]
tcormackMW: +1

framework/src/bundle/states/BPStoppingState.h
```diff
@@ -0,0 +1,27 @@
...
+        virtual ~BPStoppingState() = default;
+
+        void Start(BundlePrivate& mgr, uint32_t options) override;
+
+        void Stop(BundlePrivate& mgr, uint32_t options) override;
+
+        void Uninstall(BundlePrivate& mgr) override;
+
+        uint32_t GetState() override {
+            return 0x00000010;
```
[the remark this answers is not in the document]
tcormackMW: +1

framework/src/bundle/states/BPActiveState.h
```diff
@@ -0,0 +1,25 @@
+#ifndef BPActiveState_h
+#define BPActiveState_h
+
+#include "BundlePrivateState.h"
+
+namespace cppmicroservices
+{
+    class BPActiveState final : public BundlePrivateState
```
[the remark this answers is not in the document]
howellx: defined the constructors and operators in a later commit

framework/src/bundle/states/BPInstalledState.h
```diff
@@ -0,0 +1,25 @@
+#ifndef BPInstalledState_h
+#define BPInstalledState_h
+
+#include "BundlePrivateState.h"
+
+namespace cppmicroservices
+{
+    class BPInstalledState final : public BundlePrivateState
```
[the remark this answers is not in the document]
howellx: defined the constructors and operators in a later commit

framework/src/bundle/states/BundlePrivateState.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <cstdint>
+#include <future>
+#include <memory>
+
+namespace cppmicroservices
+{
+    class BundlePrivate;
+
+    class BundlePrivateState : public std::enable_shared_from_this<BundlePrivateState>
```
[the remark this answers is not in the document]
howellx: defined the constructors and operators in a later commit

framework/src/bundle/states/BPActiveState.h
```diff
@@ -0,0 +1,25 @@
...
+    { 
+        public:
+        using BundlePrivateState::BundlePrivateState;
+        virtual ~BPActiveState() = default;
```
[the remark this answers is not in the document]
howellx: applied suggestion in a later commit

framework/src/bundle/states/BPResolvedState.h
```diff
@@ -0,0 +1,25 @@
+#ifndef BPResolvedState_h
+#define BPResolvedState_h
+
+#include "BundlePrivateState.h"
+
+namespace cppmicroservices
+{
+    class BPResolvedState final : public BundlePrivateState
```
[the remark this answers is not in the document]
howellx: defined the constructors and operators in a later commit

framework/src/bundle/states/BPStartingState.h
```diff
@@ -0,0 +1,24 @@
+#ifndef BPStartingState_h
+#define BPStartingState_h
+
+#include "BundlePrivateState.h"
+
+namespace cppmicroservices
+{
+    class BPStartingState final : public BundlePrivateState
```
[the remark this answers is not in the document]
howellx: defined the constructors and operators in a later commit

framework/src/bundle/states/BPStoppingState.h
```diff
@@ -0,0 +1,27 @@
+#ifndef BPStoppingState_h
+#define BPStoppingState_h
+
+#include "BundlePrivateState.h"
+
+namespace cppmicroservices
+{
+    class BPStoppingState final : public BundlePrivateState
```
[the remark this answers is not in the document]
howellx: defined the constructors and operators in a later commit

framework/src/bundle/states/BPStartingState.h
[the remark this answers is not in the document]
howellx: replaced with a named constant

framework/src/bundle/states/BundlePrivateState.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+    BundlePrivateState::BundlePrivateState(std::shared_future<void> blockUntil)
+    : ready(std::move(blockUntil))
+    {
+    }
+
+    void BundlePrivateState::SetAutostart(BundlePrivate& mgr, uint32_t options){
+        if ((options & Bundle::START_TRANSIENT) == 0)
+        {
+            mgr.SetAutostartSetting(options);
```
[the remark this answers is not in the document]
howellx: this pattern was kept from the old bundle lifecycle implementation

framework/src/bundle/BundlePrivate.cpp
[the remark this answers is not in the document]
howellx: This code is from the old bundle lifecycle implementation in the function `setStateInstalled`. There are other files in the codebase that depend on this function, so it was not refactored out. The new bundle lifecycle implementation does not rely on this function.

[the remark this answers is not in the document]
tcormackMW: Ah I see, this was github doing the wrong diff. Ok... weird function. Alright we can resolve this

framework/src/bundle/BundlePrivate.h
```diff
@@ -181,7 +157,7 @@ namespace cppmicroservices
          */
         // GCC 4.6 atomics do not support custom trivially copyable types
         // like enums yet, so we use the underlying primitive type here.
-        std::atomic<uint32_t> state;
+        // std::atomic<uint32_t> state;
```
tcormackMW: this can probs go, along with its comment

framework/src/util/FrameworkPrivate.cpp
```diff
@@ -62,7 +69,10 @@ namespace cppmicroservices
     void
     FrameworkPrivate::DoInit()
     {
-        state = Bundle::STATE_STARTING;
+        auto startingState = std::make_shared<BPStartingState>();
+        auto currState = GetStateObj();
+        CompareAndSetState(&currState, startingState);
```
tcormackMW: Can this fail? I guess based on the original code we assume not

framework/src/bundle/states/BPInstalledState.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+        auto currState = shared_from_this(); 
+        std::promise<void> transitionAction; 
+        auto fut = transitionAction.get_future();
+        auto uninstalledState = std::make_shared<BPUninstalledState>(std::move(fut));
+
+        while(currState->GetState() == Bundle::STATE_INSTALLED){
+            if (mgr.CompareAndSetState(&currState, uninstalledState))
+            {
+                currState->WaitForTransitionTask();
```
tcormackMW: I find it helpful here to clarify what `currState` is. At this point, `currState` is still the `BPInstalledState`. The confusing bit here is that `currState` is not actually the state of the bundle for this entire function.

Essentially by this line we are no longer in `BPInstalled` so it being `currState` is weird (to me). I know this is the way DS does it... but maybe we do it different. 

I may only be advocating for a var name change though. something as simple as naming `currState` -> `installedState` instead or something like that.

framework/src/bundle/states/BPInstalledState.cpp
```diff
@@ -0,0 +1,92 @@
...
+                    { BundleEvent::BUNDLE_RESOLVED, MakeBundle(mgr.shared_from_this()) });
+                    
+                transitionAction.set_value();
+                resolvedState->Start(mgr, options);
+                break;
+            }
+        }
+    }
+
+    void BPInstalledState::Stop(BundlePrivate& mgr, uint32_t options){
```
tcormackMW, line 39: I am not sure about the setAutostart here. I am looking at `BundlePrivate.cpp` (original) and in stop for STATE_INSTALLED, I see it returning without `setAutostartSetting` unless `            if ((options & Bundle::STOP_TRANSIENT) == 0)`

Can you clarify why this was set to default `SetAutostart` here?

framework/src/bundle/states/BundlePrivateState.cpp
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+
+    void BundlePrivateState::CheckFrameworkHasStopped(BundlePrivate& mgr){
```
tcormackMW: This needs to return the block. If you look at the original code, those blocks are held for the entire state change. This ensures that the framework does not stop while we are starting a bundle

framework/src/util/FrameworkPrivate.cpp
howellx: The old code assumes not but I'm guessing in practice that it can. Since refactoring `FrameworkPrivate` is likely its own undertaking, I just made the bare minimum changes to get it to work with the new state object.

framework/src/bundle/states/BPInstalledState.cpp
howellx: I had made a quick custom `BundlePrivateState::SetAutostart` that does the `if ((options & Bundle::STOP_TRANSIENT) == 0)` check. Maybe I could change the function name from something that isn't `SetAutostart` to avoid this confusion?

framework/src/bundle/states/BPResolvedState.cpp
```diff
@@ -0,0 +1,62 @@
...
+        return;
+    };
+
+    void BPResolvedState::Uninstall(BundlePrivate& mgr){
+        auto currState = shared_from_this(); 
+        auto installedState = std::make_shared<BPInstalledState>();
+
+        while(currState->GetState() == Bundle::STATE_RESOLVED){
+            if (mgr.CompareAndSetState(&currState, installedState))
+            {
```
tcormackMW: even though it is a noop, i would probably like to see a waitForTransitionTask here as well

framework/src/bundle/states/BPInstalledState.cpp
howellx: My understanding is that `currState` is meant to be the current state of the bundle, not the state of the call site. For instance, if a different thread had interrupted this function and set the state to something else, `currState` would capture that. Maybe something such as `currentBundleState` would be more explicit?

tcormackMW: The CAS has succeeded which means the current state of the bundle is `uninstalledState`. `currState` is the old `installedState`. 

If the CAS fails, the variable `currState` assumes that value. So I guess in some cases it is the actual current state, but in the success path, at this point `currState` is the old state

framework/src/bundle/states/BPInstalledState.cpp
tcormackMW: Ah I see that makes sense. That is fine.

framework/src/bundle/states/BPInstalledState.cpp
howellx: Ah good point, maybe something such as `observedState` or `previousState` is a better name. 

In general, the new design is a bit strange in that it first sets the new state, then does the work required to get to the new state under a promise/future. The old design would do the work required to get to the new state before setting the state variable, which makes more intuitive sense imo. This new behavior was based on the way DS handles concurrency as you mentioned, and I think it is safe theoretically as the work is protected by a promise. If there is another design that flips the order around and still maintains safe concurrency mechanisms, it may be worth investigating.

framework/src/bundle/states/BPInstalledState.cpp
```diff
@@ -0,0 +1,123 @@
...
+            if (mgr.CompareAndSetState(&currState, resolvedState)){
+                currState->WaitForTransitionTask();      
+                
+                auto frameworkBlock = CheckAndBlockFramework(mgr);
+                SetAutostart(mgr, options);
+
+                mgr.coreCtx->listeners.BundleChanged(
+                    { BundleEvent::BUNDLE_RESOLVED, MakeBundle(mgr.shared_from_this()) });
+                
+                frameworkBlock.reset();
```
tcormackMW: I understand why we reset these. It makes sense. This, however, is my reasoning essentially for the while loop rather than recursion. 

If we were in a loop, the RAII would just do this and then the next transition would happen from the top and there wouldn't be these questions of if/when to release this....

That being said, for this iteration I understand the purpose and reasoning. Just marking this for future conversations

framework/src/bundle/states/BPStartingState.cpp
```diff
@@ -0,0 +1,268 @@
...
+                                        thisBundle,
+                                        "The bundle validation function threw an exception",
+                                        std::current_exception()));
+                        frameworkBlock.reset();
+                        transitionAction.set_value();
+                        activeState->Stop(mgr, options);
+                        throw (SecurityException { util::GetLastExceptionStr(), thisBundle });
+                    }
+
+                    if(!valid){
```
tcormackMW: why have the `valid` thing? why not just return from line 77. if we hit line 77 we are guaranteed to go directly to line 93 -- ergo we should just run this up there, no?

framework/src/bundle/states/BundlePrivateState.cpp
```diff
@@ -0,0 +1,76 @@
...
+            case Bundle::STATE_STOPPING:
+                return "STATE_STOPPING";
+            case Bundle::STATE_ACTIVE:
+                return "STATE_ACTIVE";
+            default:
+                return "UNKNOWN_STATE(" + std::to_string(state) + ")";
+        }
+    }
+
+    void BundlePrivateState::LogDroppedTransition(BundlePrivate& mgr,
```
tcormackMW: probably worth clarifying in the message what a dropped message is--

`we dropped it because you lost a race -- look into why your bundle is transitioning between states so much... etc`

framework/src/bundle/states/BPStoppingState.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+    void BPStoppingState::Stop(BundlePrivate& mgr, uint32_t options){
+        
+        bool successfulTransition = false;
+        auto currState = shared_from_this(); 
+        std::promise<void> transitionAction; 
+        auto fut = transitionAction.get_future();
+        auto resolvedState = std::make_shared<BPResolvedState>(std::move(fut)); 
+
+        while(currState->GetState() == Bundle::STATE_STOPPING){
```
tcormackMW: I am not sure this is correct. 

As I see it, `Stop0` (the thing that calls `stop1` which is the thing that calls `stop2` if `op==deactivating` ... goodness we have good naming) is only invoked in case STATE_ACTIVE or STATE_STARTING. Else it returns:

```
    BundlePrivate::Stop(uint32_t options)
    {
        std::exception_ptr savedException;

        {
            auto l = this->Lock();
            US_UNUSED(l);

            if (state == Bundle::STATE_UNINSTALLED)
            {
                throw std::logic_error("Bundle " + symbolicName + " (location=" + location + ") is uninstalled");
            }

            if ((options & Bundle::STOP_TRANSIENT) == 0)
            {
                SetAutostartSetting(-1);
            }
            switch (static_cast<Bundle::State>(state.load()))
            {
                case Bundle::STATE_INSTALLED:
                case Bundle::STATE_RESOLVED:
                case Bundle::STATE_STOPPING: //FINDME I THINK WE FALL THROUGH TO THE RETURN IN STATE_UNINSTALLED
                case Bundle::STATE_UNINSTALLED:
                    return;

                case Bundle::STATE_ACTIVE:
                case Bundle::STATE_STARTING: // Lazy start...
                    savedException = Stop0();
                    break;
            }
        }

        if (savedException)
        {
            std::rethrow_exception(savedException);
        }
    }
```

framework/src/bundle/states/BPStoppingState.cpp
```diff
@@ -0,0 +1,103 @@
...
+    void BPStoppingState::Uninstall(BundlePrivate& mgr){
+        bool successfulTransition = false;
+        auto currState = shared_from_this(); 
+        std::promise<void> transitionAction; 
+        auto fut = transitionAction.get_future();
+        auto resolvedState = std::make_shared<BPResolvedState>(std::move(fut)); 
+
+        while(currState->GetState() == Bundle::STATE_STOPPING){
+            if (mgr.CompareAndSetState(&currState, resolvedState)){
+                currState->WaitForTransitionTask();
```
tcormackMW: I don't think this is correct. 

The uninstall mathod:

```
                case Bundle::STATE_STOPPING:
                {
                    std::exception_ptr exception;
                    try
                    {
                        exception = (state & (Bundle::STATE_ACTIVE | Bundle::STATE_STARTING)) != 0 ? Stop0() : nullptr;
                    }
                    catch (...)
                    {
                        // Force to install
                        SetStateInstalled(false);
                        exception = std::current_exception();
                    }
                    operation = BundlePrivate::OP_UNINSTALLING;

```
in our case, we would fall through all of this because state & ... would be 0, so exception would be nullptr -- this makes sense, we are already stopping, we don't need to do that work again.

Now we fall through all the wya to state_installed:
```
                case Bundle::STATE_INSTALLED:
                {
                    coreCtx->bundleRegistry.Remove(location, id);
                    if (operation != BundlePrivate::OP_UNINSTALLING)
                    {
                        try
                        {
                            operation = BundlePrivate::OP_UNINSTALLING;
                        }
                        catch (...)
                        {
                            // Make sure that bundleContext is invalid
                            std::shared_ptr<BundleContextPrivate> ctx;
                            if ((ctx = bundleContext.Exchange(ctx)))
                            {
                                ctx->Invalidate();
                            }
                            operation = BundlePrivate::OP_UNINSTALLING;
                            coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
                                                                                 MakeBundle(shared_from_this()),
                                                                                 std::string(),
                                                                                 std::current_exception()));
                        }
                    }
```

I.e., and I think your state diagram confirms this, we should really just be moving from state_stopping to state_installed and then running uninstall on state_installed.

framework/src/bundle/states/BPStoppingState.cpp
```diff
@@ -0,0 +1,103 @@
...
+                }
+                mgr.coreCtx->listeners.BundleChanged({ BundleEvent::BUNDLE_STOPPED, MakeBundle(mgr.shared_from_this()) }); //listener stopp-ED event
+                transitionAction.set_value();
+                successfulTransition = true;
+                resolvedState->Uninstall(mgr);
+                break;
+            }
+        }
+
+        if(!successfulTransition){
```
tcormackMW: general comment. would be nice if we could somehow integrate this into an raii object or something... i just don't love the repetative if statement at the end of each state transition. if we could just create an object at the top and the destructor ran this if it just cleans up these functions to be more focussed on the state transition work

framework/src/bundle/states/BPStartingState.cpp
howellx: The only issue is that if we throw an exception at line 77, it will get caught by the `catch` block on line 80 and execute code we don't want to execute. That being said, I will rewrite the conditional statement that sets `valid` to be more direct as it is kind of roundabout currently.

framework/src/bundle/states/BPStoppingState.cpp
howellx: Done!

framework/src/bundle/states/BPStartingState.cpp
tcormackMW: I don't understand. If we throw at the if statement on line 74

```
                    bool valid = true;
                    try
                    {
                        if (mgr.coreCtx->validationFunc && (mgr.lib.GetFilePath() != util::GetExecutablePath())
                            && !mgr.coreCtx->validationFunc(thisBundle))
                        {
                            valid = false;
                        }
                    }
                    catch (...)
                    {
                        mgr.coreCtx->listeners.SendFrameworkEvent(
                            FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_WARNING,
                                        thisBundle,
                                        "The bundle validation function threw an exception",
                                        std::current_exception()));
                        frameworkBlock.reset();
                        transitionAction.set_value();
                        activeState->Stop(mgr, options);
                        throw (SecurityException { util::GetLastExceptionStr(), thisBundle });
                    }

                    if(!valid){
                        frameworkBlock.reset();
                        transitionAction.set_value();
                        activeState->Stop(mgr, options);
                        throw (SecurityException {
                            "Bundle " + mgr.symbolicName + " (location=" + mgr.location + ") failed bundle validation.",
                            thisBundle });
                    }
```

we get caught below, and then we rethrow a security exception. we would never hit the valid check.

howellx: Yeah, that's the intended behavior for the case where the validation function throws. However, if the validation function executes successfully but returns that we just failed validation (which is what the first if statement checks for), then we want to throw a different exception. If we throw this second exception directly in the try block, it'll be caught by the catch block and also run the code for a thrown validation function, which we don't want.

tcormackMW: ohhhhhhhhhhhhhhhhhhhhhhhhh. yup.

framework/src/bundle/states/BundlePrivateState.cpp
howellx: Updated log message to be more explicit, feel free to provide feedback on the content/wording

framework/src/bundle/states/BPStoppingState.cpp
howellx: You make a good point. In the original code, if `Stop()` is called on a bundle that's already in `STATE_STOPPING`, then it does nothing like you said. However, if `Stop()` is called on a bundle that is `STATE_ACTIVE` or `STATE_STARTING`, it will move the bundle to `STATE_STOPPING` and then do some stopping work depending on which of the two states it came from. It would check which state we came from using a member variable `wasStarted`.  In the current implementation, I had some of the work occur under `BPActiveState::Stop()` and `BPStartingState::Stop()` to get them to `STATE_STOPPING` and the stopping work they both need to do would run under `BPStoppingState::Stop()`. Thinking about it now though, it feels a bit weird to split the work in half. I think I felt weird having `BPStoppingState::Stop()` be a no-op, like in theory I felt like stopping work should occur under the `STATE_STOPPING > STATE_RESOLVED` transition.

framework/src/bundle/states/BPStoppingState.cpp
howellx: Done!

framework/src/bundle/states/BPInstalledState.cpp
howellx: Discussed with Toby offline, this is mainly just a naming thing. I renamed `currState` to `observedState` as I feel like it better reflects the nature of that variable.

framework/src/bundle/states/BPStoppingState.cpp
howellx: Discussed with Toby in a call. In the original implementation, there is another code path during a Bundle stop that sets the state as STOPPING and then continues to do work:

```
    std::exception_ptr
    BundlePrivate::Stop0()
    {
        wasStarted = state == Bundle::STATE_ACTIVE;
        state = Bundle::STATE_STOPPING;
        operation = OP_DEACTIVATING;

        std::exception_ptr savedException = Stop1(); //does a bunch of work before setting to RESOLVED later

        if (state != Bundle::STATE_UNINSTALLED)
        {
            state = Bundle::STATE_RESOLVED;
            coreCtx->listeners.BundleChanged({ BundleEvent::BUNDLE_STOPPED, MakeBundle(shared_from_this()) });
            operation = OP_IDLE;
        }
        return savedException;
```

The work I currently have in `Stopping->Stop` essentially does the `Stop1()` work from above. The case you mentioned with the instant return in the old `Stop()` is only if `Stop()` is explicitly/externally called on a bundle in Stopping. This is not possible due to locks, _except_ in the case where bundle activator code calls stop on its own bundle. 

The new implementation accounts for this by splitting the Stop work in half, around the bundle activator call. All the Stop work prior to the bundle activator call lives in `Active->Stop`, and all the work after lives in `Stopping->Stop`. The new implementation allows bundle activator code to take the bundle to a different state, but then when we return from the bundle activator call and we try to do the rest of the Stop work through a recursive `Stopping->Stop` call, it will fail the while loop or CAS check. 

This is functionally the same as the original implementation. Say `Stop` is called on a `Stopping` bundle in a bundle activator call. In the original implementation, the bundle activator call will no-op and then the original `Stop` call will finish the stopping work. In the new implementation, the bundle activator `Stop` call will finish the stopping work, and the original call will no-op after that returns and the subsequent `Stopping->Stop` fails CAS/the while loop condition. 

The only difference in behavior is that as the bundle goes through the transitions from the bundle activator `Stop` call, it can be interrupted by another thread. imo this tradeoff is okay and happens throughout the new implementation in general.

framework/src/bundle/states/BPStoppingState.cpp
howellx: Correction, after further investigation, the original implementation for `Uninstall` does have code path where it goes from `Stopping` to `Resolved` to `Installed` to finally `Uninstalled`. This occurs when `Uninstall` is called on either an `Active` or `Starting` bundle. `Uninstall` will only skip the `Resolved` state if `Uninstall` is directly called on a stopping bundle. Since `Stopping` is an intermediate state that only lives during transition work, this is not possible. There is the case where bundle activator code calls uninstall in its `Stop` function but even is this situation, the original code just hangs due to the lock it uses. 

I think we should still go to `Resolved` even if the `Uninstall` is called in the bundle activator code, since every other `Uninstall` path goes to `Resolved` and there isn't any reason this particular `Uninstall` path shouldn't. The only difference in behavior is we don't hang in the edge-case described above, and we preserve the behavior for all the more standard cases.

framework/src/bundle/states/BPStartingState.cpp
```diff
@@ -0,0 +1,118 @@
...
+                    mgr.coreCtx->listeners.BundleChanged(
+                        BundleEvent(BundleEvent::BUNDLE_STARTED, MakeBundle(mgr.shared_from_this())));
+                    
+                }
+                catch (cppmicroservices::SecurityException const& ex)
+                {
+                    frameworkBlock.reset();
+                    completeTransition.Complete();
+                    activeState->Stop(mgr, options);
+                    throw ex;
```
jeffdiclemente: minor: I recommend using `throw` to prevent an unnecessary copy and to prevent any possibility of future slicing (should the `SecurityException` class ever be marked as not `final`).

framework/src/bundle/BundlePrivate.cpp
```diff
@@ -66,659 +73,112 @@ namespace cppmicroservices
...
-            {
-                resolveFailException = std::current_exception();
-                coreCtx->listeners.SendFrameworkEvent(FrameworkEvent(FrameworkEvent::Type::FRAMEWORK_ERROR,
-                                                                     MakeBundle(this->shared_from_this()),
-                                                                     std::string(),
-                                                                     std::current_exception()));
-            }
-        }
-        return static_cast<Bundle::State>(state.load());
+        return std::atomic_compare_exchange_strong(&state, expectedState, desiredState);
```
jeffdiclemente: minor: please make a comment that we should move to using `std::atomic<std::shared_ptr<T>>`, instead of the `std::atomic` free functions, once this library officially supports C++20.

framework/src/bundle/states/BPActiveState.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "cppmicroservices/Bundle.h"
+#include "cppmicroservices/BundleEvent.h"
+#include "cppmicroservices/util/Error.h"
+#include "cppmicroservices/FrameworkEvent.h"
+
+namespace cppmicroservices
+{
+    uint32_t BPActiveState::GetState(){
+        return Bundle::STATE_ACTIVE;
+    };
```
jeffdiclemente: minor: unneeded `;`

framework/src/bundle/states/BPActiveState.cpp
```diff
@@ -0,0 +1,103 @@
...
+    
+    void BPActiveState::Start(BundlePrivate& mgr, uint32_t options){
+        auto frameworkBlock = CheckAndBlockFramework(mgr);
+        US_UNUSED(frameworkBlock);
+        US_UNUSED(options);
+        return;
+    };
```
jeffdiclemente: minor: unneeded `;`

framework/src/bundle/states/BPResolvedState.cpp
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                    catch (...)
+                    {
+                        mgr.coreCtx->logger->Log(logservice::SeverityLevel::LOG_INFO,
+                                            "Failed to start Bundle " + mgr.symbolicName + " (location=" + mgr.location + ")",
+                                            std::current_exception());
+                        frameworkBlock.reset();
+                        completeTransition.Complete();
+                        startingState->Stop(mgr, options);
+                        throw (std::runtime_error("Bundle " + mgr.symbolicName + " (location= " + mgr.location
```
jeffdiclemente: minor: the three functions are called in both `catch` blocks; consider rolling these operations into a RAII type object or consolidating into a single function.

jeffdiclemente (approved):
overall this looks good and I'm glad to see this major refactoring of the bundle lifecycle logic that should improve concurrency and further reduce race conditions. Thanks!

framework/src/bundle/states/BPStartingState.cpp
howellx: Done

framework/src/bundle/states/BPResolvedState.cpp
howellx: `frameworkBlock.reset();` and `completeTransition.Complete();` were already wrapped in RAII objects, removed the redundant calls I had like these

framework/src/bundle/states/BPInstalledState.cpp
howellx: Discussed with Toby, we ended up ditching the recursive calls for `Stop()` and `Start()` so this `frameworkBlock` pattern no longer is relevant.

